PR #3, opened 2018-12-22
Add a fixed-seed API to the simulator

josephlr:
This allows the simulator to behave deterministicly for the purposes of testing.


simulator/simulator.go
```diff
@@ -73,6 +79,23 @@ func Get() (*Simulator, error) {
...
 }
 
+// GetWithFixedSeed behaves like Get() expect that all of the internal
+// heirarchy seeds are derived from the input seed.
+func GetWithFixedSeed(seed int) (*Simulator, error) {
+	s, err := Get()
+	if err != nil {
+		return nil, err
+	}
+	for i := 0; i < 64; i++ {
```
twitchy-jsonp: Is this intended to be deterministically puesdorandom? Could we just use `math/rand` if so?

josephlr: Fixed to use math.rand

simulator/simulator_test.go
```diff
@@ -84,3 +84,36 @@ func TestGetRandom(t *testing.T) {
...
+func TestFixedSeedExpectedModulus(t *testing.T) {
+	s, err := GetWithFixedSeedInsecure(0)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer s.Close()
+
+	modulus := getEKModulus(t, s)
+	if modulus.Cmp(zeroSeedModulus()) != 0 {
+		t.Fatalf("Got %v expected %v", modulus, zeroSeedModulus())
```
twitchy-jsonp: I think asserts are meant to have error messages that stand alone (though not 100% on this one).

Thoughts on: `getEKModulus() = %v, want %v` ?

josephlr: Good idea, changed

PR #8, opened 2019-01-18
Utility for flushing active TPM handles.

samdamana:
Reads all active handles through a call to the TPM `rw`:

```golang
... := tpm2.GetCapability(rw, tpm2.CapabilityHandles, ^uint32(0), 0x80000000)
```

Then flushes those handles with `tpm2.FlushContext(...)`.

I am not 100% certain on the magic const `0x80000000` used in GetCapability but it appears to work correctly. I found it here: https://github.com/google/go-tpm/blob/master/tpm2/tpm2_test.go#L135


awly:
Move the files to a separate directory under `github.com/google/go-tpm-tools/cmd` and make it package main. This is mostly useful as a standalone binary.

awly:
`0x80000000` is `property` parameter of the `TPM2_GetCapability` call.
For handles it defines which type of handle to return. The type of the handle (using `TPM_HT_...` constants) is placed in most significant octet of the property.
In this case it's `TPM_HT_TRANSIENT`.

The code should make it more obvious instead of using an opaque constant.

tpm2tools/flush_active_handles.go
```diff
@@ -0,0 +1,40 @@
...
+        if err != nil {
+            return err
+        }
+    }
+    return nil
+}
+
+// activeHandles returns a slice of tpmutil.Handle objects of all handles
+// active within the TPM at ioReadWriter rw.
+func activeHandles(rw io.ReadWriter) ([]tpmutil.Handle, error) {
```
josephlr: This is the main function which is useful at the library level. It would be nice to have a way to get a slice of all the handles with some specific property. However, this function isn't correct as written.

tpm2tools/flush_active_handles.go
```diff
@@ -0,0 +1,40 @@
...
+    vals, _, err := tpm2.GetCapability(rw, tpm2.CapabilityHandles, ^uint32(0), 0x80000000)
```
josephlr: Here we discard the `moredata` return parameter. This is not correct if I'm reading the spec correctly. We should just check that `moredata` is false (as this should [_never_ return true](https://github.com/google/go-tpm-tools/blob/f14bb1799b8e2e113c8e15f1df230254d3249904/simulator/Object.c#L874) for tpm2.CapabilityHandles)

tpm2tools/flush_active_handles_test.go
```diff
@@ -0,0 +1,100 @@
...
+        }
+        if i < maxHandles {
+            i++
+            loadRandomExternalKey(t, simulator)
+        } else {
+            break
+        }
+        
+    }
+}
```
josephlr: Always have a newline at the end of a file

tpm2tools/flush_active_handles_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestActiveHandles(t *testing.T) {
+    simulator, err := simulator.Get()
+    if err != nil {
+        t.Fatal(err)
+    }
+    defer simulator.Close()
+
+    i := 0
+    for {
```
josephlr: This would be much easier to read as a normal for loop.
```go
// Check that we have no active handles
for i = 1 , i <= maxHandles, i++ {
  // Load another handle
  // Check the num handles
}
```

tpm2tools/flush_active_handles_test.go
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    // Ensure there are no active handles after all that.
+    h, err := activeHandles(simulator)
+    if err != nil {
+        t.Fatal(err)
+    }
+    if len(h) != 0 {
+        t.Fatal("Bah this should be empty!")
```
josephlr: "Handles still active after FlushAllTransientHandles()"

tpm2tools/flush_active_handles_test.go
```diff
@@ -0,0 +1,100 @@
...
+    
+
+func TestFlushActiveHandles(t *testing.T) {
+    simulator, err := simulator.Get()
+    if err != nil {
+        t.Fatal(err)
+    }
+    defer simulator.Close()
+
+    // Loads then flushes 1, 2, ...maxHandles handles.
```
josephlr: We only really need one run of FlushAllTransientHandles here

tpm2tools/flush_active_handles_test.go
```diff
@@ -0,0 +1,100 @@
...
+    // How many keys/handles can the simulator contain at once.
+    maxHandles = 3
+)
+
+func loadRandomExternalKey(t *testing.T, simulator *simulator.Simulator) {
+    pk, err := rsa.GenerateKey(rand.Reader, 2048)
+    if err != nil {
+        t.Fatal(err)
+    }
+    public := tpm2.Public{
```
josephlr: If possible copy then modify one of the templates already in this package, rather then rewriting all of this again

tpm2tools/flush_active_handles_test.go
```diff
@@ -0,0 +1,100 @@
...
+    "crypto/rsa"
+    "testing"
+
+    "github.com/google/go-tpm/tpm2"
+
+    "github.com/google/go-tpm-tools/simulator"
+)
+
+const (
+    // How many keys/handles can the simulator contain at once.
```
josephlr: This makes it sound like our simulator can only possibly hold 3 values. Make it clear that this is just how many we are trying for our tests.

tpm2tools/flush_active_handles_test.go
```diff
@@ -0,0 +1,100 @@
...
+    maxHandles = 3
+)
+
+func loadRandomExternalKey(t *testing.T, simulator *simulator.Simulator) {
```
josephlr: As @awly said, clearing all transient keys is best put in as a standalone binary. The function to list all handles of a specific type; however, should be a library function.

tpm2tools/flush_active_handles.go
josephlr: Here we are passing ^uint32(0) as the count parameter. We should instead pass MAX_CAP_HANDLES (a constant that might need to be defined in this library. See section 5.5 of [part 4 of the spec](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-4-Supporting-Routines-01.38-code.pdf)

tpm2tools/flush_active_handles.go
```diff
@@ -0,0 +1,40 @@
...
+    for _, handle := range handles {
+        err = tpm2.FlushContext(rw, handle)
+        if err != nil {
+            return err
+        }
+    }
+    return nil
+}
+
+// activeHandles returns a slice of tpmutil.Handle objects of all handles
```
josephlr: In general, I think this method should be public and be parameterized over the _type_ of handles we want to list. This method assumes that we want to list the transient handles (which is correct if we are trying to clear the handles). However, I would like it better if some sort of `HandleType` was defined and had constants corresponding to values of TPM_HT. [See table 2:27 in part 4 of the spec](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-4-Supporting-Routines-01.38-code.pdf)

The name could be something like ListHandles (or similar)

tpm2tools/flush_active_handles.go
```diff
@@ -0,0 +1,40 @@
+package tpm2tools
+
+
+import (
+    "io"
+    "github.com/google/go-tpm/tpm2"
+    "github.com/google/go-tpm/tpmutil"
+)
+
+// FlushActiveHandles calls FlushContext() on all active handles within the
```
josephlr: Here and elsewhere the correct terminology is "transient handles" instead of "active handles", as there are actually multiple types of handles we might want to clear:
  - Transient Handles (`TPM_HT_TRANSIENT`)
  - Loaded Session Handles (`TPM_HT_LOADED_SESSION`)
  - Saved Session Handles (`TPM_HT_SAVED_SESSION`)

tpm2tools/flush_active_handles.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+    "io"
+    "github.com/google/go-tpm/tpm2"
+    "github.com/google/go-tpm/tpmutil"
+)
+
+// FlushActiveHandles calls FlushContext() on all active handles within the 
+// TPM at io.ReadWriter rw. Returns nil if successful or TPM has no active
+// handles.
+func FlushActiveHandles(rw io.ReadWriter) error {
```
josephlr: As @awly said, flushing all transient handles is best put in as a standalone binary. The function to list all handles of a specific type; however, should be a library function.

tpm2tools/flush_active_handles.go
josephlr: Specifically, we would want to define (at least) the following constants:

```go
const (
	CapBufferLenMax uint32 = 1024
	CapDataMax      uint32 = CapBufferLenMax - 4 - 4
	CapHandlesMax   uint32 = CapDataMax / 4
)
```

@awly, should these be in tpm2/constants.go or this library?

awly: Since these directly map to constants from the spec, they should go in tpm2/constants.go.

samdamana: Where does the `CapBufferLenMax` of `1024` come from? I am adding this to tpm2/constants.go but following the documentation it isn't clear what this number is. Does it vary from TPM to TPM?

josephlr: I was looking at the note on `MAX_CAP_BUFFER` in Section 10.10 of [Part 2 of the spec](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf).

However, while a value of 1024 seems to be typical, the spec never actually says it must be that value. The code in part 4 of the spec is merely "illustrative". However, looking at the other code for `GetCapability` it doesn't seem like TPM2 has support for a `MAX_CAP_BUFFER` bigger than the [TPM's max response size](https://github.com/google/go-tpm/blob/79047b3352714e31d8b648e8f2d783c4a88db580/tpmutil/run.go#L31).

So I think it's best we just pass [MaxUint32](https://golang.org/pkg/math/#pkg-constants) for the number of handles, and just make sure `moredata` is false. Sorry, I thought MAX_CAP_BUFFER was a fixed constant.

tpm2tools/flush_active_handles.go
samdamana: Implemented in standalone binary.

samdamana:
Ready for another look.

cmd/flush_handles_test.go
```diff
@@ -0,0 +1,79 @@
...
+		{false, true, false, false, []tpm2.HandleType{tpm2.HandleTypeLoadedSession}},
+		{false, false, true, false, []tpm2.HandleType{tpm2.HandleTypeSavedSession}},
+		{false, true, true, false, []tpm2.HandleType{tpm2.HandleTypeLoadedSession, tpm2.HandleTypeSavedSession}},
+		{true, true, false, false, []tpm2.HandleType{tpm2.HandleTypeTransient, tpm2.HandleTypeLoadedSession}},
+		{true, false, true, false, []tpm2.HandleType{tpm2.HandleTypeTransient, tpm2.HandleTypeSavedSession}},
+		{true, true, true, false, []tpm2.HandleType{tpm2.HandleTypeTransient, tpm2.HandleTypeLoadedSession, tpm2.HandleTypeSavedSession}},
+		{false, false, false, true, []tpm2.HandleType{tpm2.HandleTypeTransient, tpm2.HandleTypeLoadedSession, tpm2.HandleTypeSavedSession}},
+	} {
+		// Overwrite flag pointer values.
+		flushTransient = &test.flushTransientFlag
```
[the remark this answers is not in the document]
samdamana: The documentation on the flag package isn't clear about usage in test contexts. It implies the [flag.Parse() method](https://golang.org/pkg/flag/#Parse) must be used before accessing flags. I didn't want to call Parse() because I am not sure what the argvs would be at runtime during tests. 

I like your solution of using flag.Set since it is more readable. It just isn't very clear in the documentation that you can do this.

cmd/flush_handles_test.go
```diff
@@ -0,0 +1,79 @@
...
+		flushLoadedSession = &test.flushLoadedSessionFlag
+		flushSavedSession = &test.flushSavedSessionFlag
+		flushAllTypes = &test.flushAllTypesFlag
+
+		got := handleTypesFromFlags()
+		if len(got) != len(test.want) {
```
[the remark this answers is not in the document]
samdamana: Done.

cmd/flush_handles_test.go
```diff
@@ -0,0 +1,79 @@
...
+		t.Fatal(err)
+	}
+	defer simulator.Close()
+
+	// Loads then flushes 1, 2, ...maxHandles transient handles.
+	for i := 0; i <= maxHandles; i++ {
+		for j := 0; j < i; j++ {
+			tpm2toolstest.LoadRandomExternalKey(t, simulator)
+		}
+		err = flushHandlesOfType(simulator, tpm2.HandleTypeTransient)
```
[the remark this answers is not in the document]
samdamana: Done.

cmd/flush_handles_test.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+
+	// Ensure there are no active handles after all that.
+	h, err := tpm2tools.Handles(simulator, tpm2.HandleTypeTransient)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(h) != 0 {
+		t.Fatal("Simulator should be empty of transient handles.")
```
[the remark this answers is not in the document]
samdamana: Done.

tpm2tools/handles.go
```diff
@@ -0,0 +1,30 @@
...
+	vals, moreData, err := tpm2.GetCapability(rw, tpm2.CapabilityHandles, math.MaxUint32, property)
+	if err != nil {
+		return nil, err
+	}
+	if moreData {
+		return nil, fmt.Errorf("tpm2.GetCapability() should never return moreData==true for tpm2.CapabilityHandles")
+	}
+	handles := make([]tpmutil.Handle, len(vals))
+	for i, v := range vals {
+		handles[i] = v.(tpmutil.Handle)
```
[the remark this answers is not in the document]
samdamana: Done

tpm2tools/handles_test.go
```diff
@@ -0,0 +1,35 @@
+package tpm2tools
+
+import (
+	"testing"
+
+	"github.com/google/go-tpm-tools/tpm2tools/tpm2toolstest"
+	"github.com/google/go-tpm/tpm2"
+
```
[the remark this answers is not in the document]
samdamana: Done.

tpm2tools/tpm2toolstest/load_random_external_key.go
```diff
@@ -0,0 +1,41 @@
...
+	"crypto/rsa"
+	"testing"
+
+	"github.com/google/go-tpm-tools/simulator"
+	"github.com/google/go-tpm/tpm2"
+)
+
+// LoadRandomExternalKey loads a randomly generated external key into the
+// TPM simulator. If any errors occur, calls t.Fatal() on the passed testing.T.
+func LoadRandomExternalKey(t *testing.T, simulator *simulator.Simulator) {
```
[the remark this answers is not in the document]
samdamana: Done.

tpm2tools/tpm2toolstest/load_random_external_key.go
```diff
@@ -0,0 +1,41 @@
...
+			KeyBits:  2048,
+			Exponent: uint32(pk.PublicKey.E),
+			Modulus:  pk.PublicKey.N,
+		},
+	}
+	private := tpm2.Private{
+		Type:      tpm2.AlgRSA,
+		Sensitive: pk.Primes[0].Bytes(),
+	}
+	_, _, err = tpm2.LoadExternal(simulator, public, private, tpm2.HandleNull)
```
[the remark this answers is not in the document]
samdamana: Done.

cmd/flush_handles/flush_handles.go
```diff
@@ -0,0 +1,82 @@
...
+	flushTransient     = flag.Bool("flush-transient", true, "Flush all transient handles.")
+	flushLoadedSession = flag.Bool("flush-loaded-session", false, "Flush all loaded session handles.")
+	flushSavedSession  = flag.Bool("flush-saved-session", false, "Flush all saved session handles.")
+	flushAllTypes      = flag.Bool("flush-all-types", false, "Flush handles of all handle types. If enabled, settings for other handle types are ignored.")
+)
+
+func main() {
+	flag.Parse()
+	handleTypes := handleTypesFromFlags()
+	err := openTPMAndFlushHandlesOfTypes(handleTypes)
```
awly: Merge with error handling

cmd/flush_handles/flush_handles.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	if *flushSavedSession {
+		types = append(types, tpm2.HandleTypeSavedSession)
+	}
+	return types
+}
+
+// openTPMAndFlushHandlesOfTypes opens the TPM defined at the flag tpm-path
+// and calls flushHandlesOfType on every type within handleTypes.
+func openTPMAndFlushHandlesOfTypes(handleTypes []tpm2.HandleType) error {
```
awly: optional: these func names look too verbose to me. I'd go with
```
func flushAll(types []tpm2.HandleType) error
func flush(rw io.ReadWriter, ht tpm2.HandleType) error
```

cmd/flush_handles_test.go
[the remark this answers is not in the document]
awly: TL;DR: avoid global state (including flags) whenever possible.

Generally speaking, tests shouldn't know about flags at all (unless you test flag parsing).
Flags are `func main`-level concern and all other code should get configuration passed explicitly through func arguments or struct fields.
Looking outside of their context into the global vars (which get set through flags at runtime) makes testing harder: you can't parallelize tests that read flags due to memory safety, you must remember to clean up modified values to make tests order-independent.

However, this program is not production-critical so we don't need to be so strict about code structure or testing.

cmd/flush_handles/flush_handles_test.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+
+	// Ensure there are no active handles after all that.
+	h, err := tpm2tools.Handles(simulator, tpm2.HandleTypeTransient)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(h) != 0 {
+		t.Fatalf("Simulator should be empty of transient handles; got: %d", len(h))
```
awly: Put "got" before "want" in message

tpm2tools/handles.go
```diff
@@ -0,0 +1,34 @@
...
+		return nil, err
+	}
+	if moreData {
+		return nil, fmt.Errorf("tpm2.GetCapability() should never return moreData==true for tpm2.CapabilityHandles")
+	}
+	handles := make([]tpmutil.Handle, len(vals))
+	for i, v := range vals {
+		handle, ok := v.(tpmutil.Handle)
+		if !ok {
+			return nil, fmt.Errorf("unable to assert type tpmutil.Handle of value %v", v)
```
awly: `%#v` to include type name

cmd/flush_handles/flush_handles.go
samdamana: Done.

cmd/flush_handles/flush_handles.go
samdamana: Done.

cmd/flush_handles/flush_handles_test.go
samdamana: Fixed I think?

tpm2tools/handles.go
samdamana: Done.

tpm2tools/flush_active_handles.go
samdamana: Done.

tpm2tools/flush_active_handles.go
samdamana: Done.

tpm2tools/flush_active_handles.go
samdamana: Done.

samdamana: Done.

tpm2tools/flush_active_handles_test.go
samdamana: Done.

tpm2tools/flush_active_handles_test.go
samdamana: Done.

tpm2tools/flush_active_handles_test.go
samdamana: I've moved this into its own testing helper under /tpm2tools/tpm2toolstest/

tpm2tools/flush_active_handles_test.go
samdamana: It only loads in transient handles. I kind of like looping load then flush. It helped me identify bugs in the past where I returned incorrect numbers of handles.

tpm2tools/flush_active_handles_test.go
samdamana: Done.

tpm2tools/flush_active_handles_test.go
samdamana: Fixed.

awly:
LGTM, leaving this up to @josephlr

tpm2tools/tpm2toolstest/load_random_external_key.go
```diff
@@ -0,0 +1,41 @@
...
+func LoadRandomExternalKey(t *testing.T, simulator *simulator.Simulator) {
+	t.Helper()
+	pk, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		t.Fatal(err)
+	}
+	public := tpm2.Public{
+		Type:       tpm2.AlgRSA,
+		NameAlg:    tpm2.AlgSHA1,
+		Attributes: tpm2.FlagSign | tpm2.FlagSensitiveDataOrigin | tpm2.FlagUserWithAuth,
```
mikedanese: SensitiveDataOrigin doesn't make sense in this context. I'm disappointed if this is not an error.

samdamana: Its not haha. But removed.

tpm2tools/handles_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/google/go-tpm/tpm2"
+)
+
+const (
+	// Maximum number of handles to keys tests can create within a simulator.
+	maxHandles = 3
+)
+
+func TestHandles(t *testing.T) {
+	simulator, err := simulator.Get()
```
josephlr: Use `internal.GetTPM()` see [this example](https://github.com/google/go-tpm-tools/blob/551cd9014fbdf21ed8bbe3e1a5cb7ec1564c3c2d/tpm2tools/keys_test.go#L30-L32)

cmd/flush_handles/flush_handles_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+		got := handleTypesFromFlags()
+		if !reflect.DeepEqual(got, test.want) {
+			t.Fatalf("got %v; want %v;", got, test.want)
+		}
+	}
+}
+
+func TestFlush(t *testing.T) {
+	simulator, err := simulator.Get()
```
josephlr: Use `internal.GetTPM()` see [this example](https://github.com/google/go-tpm-tools/blob/551cd9014fbdf21ed8bbe3e1a5cb7ec1564c3c2d/tpm2tools/keys_test.go#L30-L32)

tpm2tools/tpm2toolstest/load_random_external_key.go
```diff
@@ -0,0 +1,41 @@
+package tpm2toolstest
```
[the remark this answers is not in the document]
samdamana: Oh neat. The internal package didn't exist when I started on this PR. :P Done!

tpm2tools/tpm2toolstest/load_random_external_key.go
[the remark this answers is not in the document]
samdamana: Interesting. Okay done.

cmd/flush_handles/flush_handles_test.go
samdamana: Done.

tpm2tools/handles_test.go
samdamana: Done.

PR #9, opened 2019-01-19
Introduce SRK Template

stevenrutherford:
Adds in a reasonable SRK template based on the TCG's guidance.


tpm2tools/keys.go
```diff
@@ -28,6 +28,11 @@ func EndorsementKeyRSA(rw io.ReadWriter) (*Key, error) {
 	return NewKey(rw, tpm2.HandleEndorsement, DefaultEKTemplateRSA())
 }
 
+// StorageRootKeyRSA generates and loads a key from DefaultSRKTemplateRSA.
```
josephlr: Typo DefaultSRKTemplateRSA should be SRKTemplateRSA

PR #10, opened 2019-01-22
Add library for (re|un|)sealing with TPM2.0

stevenrutherford:
Adds in a library for sealing, unsealing, and resealing with the TPM2.0.

Also adds tests for typical usage of these tools.

### DIFF
--- a/tpm2tools/keys.go
+++ b/tpm2tools/keys.go
@@ -3,6 +3,7 @@ package tpm2tools
 import (
 	"crypto"
 	"crypto/rsa"
+	"crypto/sha256"
 	"fmt"
 	"io"
 
@@ -132,4 +133,174 @@ func (k *Key) PublicKey() crypto.PublicKey {
 // do as most TPMs can only have a small number of key simultaneously loaded.
 func (k *Key) Close() {
 	tpm2.FlushContext(k.rw, k.handle)
+}
+
+// SealedBytes stores the result of a TPM2_Seal. The private portion (priv) has
+// already been encrypted and is no longer sensitive.
+type SealedBytes struct {
+	priv []byte
+	pub  []byte
+}
+
+// Seal seals the sensitive byte buffer to the provided PCRs under the owner
+// hierarchy using the SHA256 versions of the provided PCRs.
+// The Key k is used as the parent key.
+func (k *Key) Seal(pcrs []int, sensitive []byte) (SealedBytes, error) {
+	auth, err := getPCRSessionAuth(k.rw, pcrs)
+	if err != nil {
+		return SealedBytes{}, fmt.Errorf("could not get pcr session auth: %v", err)
+	}
+
+	return sealHelper(k.rw, k.Handle(), auth, sensitive)
+}
+
+func sealHelper(rw io.ReadWriter, parentHandle tpmutil.Handle, auth []byte, sensitive []byte) (SealedBytes, error) {
+	priv, pub, err := tpm2.Seal(rw, parentHandle, "", "", auth, sensitive)
+	if err != nil {
+		return SealedBytes{}, fmt.Errorf("failed to seal data: %v", err)
+	}
+	return SealedBytes{priv, pub}, nil
+}
+
+// Unseal takes a private/public pair of buffers and attempts to reverse the
+// sealing process under the owner hierarchy using the SHA256 versions of the
+// provided PCRs.
+// The Key k is used as the parent key.
+func (k *Key) Unseal(pcrs []int, in SealedBytes) ([]byte, error) {
+	session, err := createPCRSession(k.rw, pcrs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unseal: %v", err)
+	}
+	defer tpm2.FlushContext(k.rw, session)
+
+	sealed, _, err := tpm2.Load(
+		k.rw,
+		k.Handle(),
+		/*parentPassword=*/ "",
+		in.pub,
+		in.priv)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load sealed object: %v", err)
+	}
+	defer tpm2.FlushContext(k.rw, sealed)
+
+	return tpm2.UnsealWithSession(k.rw, session, sealed, "")
+}
+
+// Reseal unwraps a secret and rewraps it under the auth value that would be
+// produced by the PCR state in pcrs. Similar to seal and unseal, this acts on
+// the SHA256 PCRs and uses the owner hierarchy.
+// The Key k is used as the parent key.
+func (k *Key) Reseal(pcrs map[int][]byte, in SealedBytes) (SealedBytes, error) {
+	pcrNums := make([]int, 0, len(pcrs))
+	for key := range pcrs {
+		pcrNums = append(pcrNums, key)
+	}
+
+	sensitive, err := k.Unseal(pcrNums, in)
+	if err != nil {
+		return SealedBytes{}, fmt.Errorf("failed to unseal: %v", err)
+	}
+
+	auth, err := computePCRSessionAuth(pcrs)
+	if err != nil {
+		return SealedBytes{}, fmt.Errorf("failed to compute pcr session auth: %v", err)
+	}
+
+	return sealHelper(k.rw, k.Handle(), auth, sensitive)
+}
+
+type tpmsPCRSelection struct {
+	Hash tpm2.Algorithm
+	Size byte
+	PCRs tpmutil.RawBytes
+}
+
+type sessionSummary struct {
+	OldDigest      tpmutil.RawBytes
+	CmdIDPolicyPCR uint32
+	NumPcrSels     uint32
+	Sel            tpmsPCRSelection
+	PcrDigest      tpmutil.RawBytes
+}
+
+func computePCRSessionAuth(pcrs map[int][]byte) ([]byte, error) {
+	var pcrBits [3]byte
+	for pcr := range pcrs {
+		byteNum := pcr / 8
+		bytePos := byte(1 << byte(pcr%8))
+		pcrBits[byteNum] |= bytePos
+	}
+	pcrDigest := digestPCRList(pcrs)
+
+	summary := sessionSummary{
+		OldDigest:      make([]byte, sha256.Size),
+		CmdIDPolicyPCR: uint32(tpm2.CmdPolicyPCR),
+		NumPcrSels:     1,
+		Sel: tpmsPCRSelection{
+			Hash: tpm2.AlgSHA256,
+			Size: 3,
+			PCRs: pcrBits[:],
+		},
+		PcrDigest: pcrDigest,
+	}
+	b, err := tpmutil.Pack(summary)
+	if err != nil {
+		return nil, fmt.Errorf("failed to pack for hashing: %v ", err)
+	}
+
+	digest := sha256.Sum256(b)
+	return digest[:], nil
+}
+
+func digestPCRList(pcrs map[int][]byte) []byte {
+	hash := crypto.SHA256.New()
+	for i := 0; i < 24; i++ {
+		if pcrValue, exists := pcrs[i]; exists {
+			hash.Write(pcrValue)
+		}
+	}
+	return hash.Sum(nil)
+}
+
+func getPCRSessionAuth(rw io.ReadWriter, pcrs []int) ([]byte, error) {
+	handle, err := createPCRSession(rw, pcrs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get digest: %v", err)
+	}
+	defer tpm2.FlushContext(rw, handle)
+
+	digest, err := tpm2.PolicyGetDigest(rw, handle)
+	if err != nil {
+		return nil, fmt.Errorf("could not get digest from session: %v", err)
+	}
+
+	return digest, nil
+}
+
+func createPCRSession(rw io.ReadWriter, pcrs []int) (tpmutil.Handle, error) {
+	nonceIn := make([]byte, 16)
+	/* This session assumes the bus is trusted.  */
+	handle, _, err := tpm2.StartAuthSession(
+		rw,
+		tpm2.HandleNull,
+		tpm2.HandleNull,
+		nonceIn,
+		/*secret=*/ nil,
+		tpm2.SessionPolicy,
+		tpm2.AlgNull,
+		tpm2.AlgSHA256)
+	if err != nil {
+		return tpm2.HandleNull, fmt.Errorf("failed to start auth session: %v", err)
+	}
+
+	sel := tpm2.PCRSelection{
+		Hash: tpm2.AlgSHA256,
+		PCRs: pcrs,
+	}
+	if err = tpm2.PolicyPCR(rw, handle, nil, sel); err != nil {
+		return tpm2.HandleNull, fmt.Errorf("auth step PolicyPCR failed: %v", err)
+	}
+
+	return handle, nil
 }
--- a/tpm2tools/seal_test.go
+++ b/tpm2tools/seal_test.go
@@ -1,0 +1,247 @@
+package tpm2tools
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"testing"
+
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
+
+	"github.com/google/go-tpm-tools/internal"
+)
+
+func TestSeal(t *testing.T) {
+	rwc := internal.GetTPM(t)
+	defer rwc.Close()
+
+	key, err := StorageRootKeyRSA(rwc)
+	if err != nil {
+		t.Fatalf("can't create srk from template: %v", err)
+	}
+	defer key.Close()
+
+	secret := []byte("test")
+	pcrList := []int{7, 23}
+	pcrToExtend := tpmutil.Handle(23)
+
+	sealed, err := key.Seal(pcrList, secret)
+	if err != nil {
+		t.Fatalf("failed to seal: %v", err)
+	}
+
+	unseal, err := key.Unseal(pcrList, sealed)
+	if err != nil {
+		t.Fatalf("failed to unseal: %v", err)
+	}
+	if !bytes.Equal(secret, unseal) {
+		t.Fatalf("unsealed (%v) not equal to secret (%v)", unseal, secret)
+	}
+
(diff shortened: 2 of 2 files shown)


tpm2tools/keys_test.go
```diff
@@ -70,3 +72,184 @@ func BenchmarkStorageRootKeyRSA(b *testing.B) {
...
+
+	secret := []byte("test")
+
+	extension := bytes.Repeat([]byte{0xAA}, sha256.Size)
+	err = tpm2.PCRExtend(rwc, tpmutil.Handle(7), tpm2.AlgSHA256, extension, "")
+	if err != nil {
+		t.Fatalf("failed to extend pcr: %v", err)
+	}
+
+	priv, pub, err := key.Seal(rwc, []int{7}, secret)
```
josephlr: Instead of having tpmutil.Handle(7) scattered about, we should define a specific test PCR at the top of this file. It should ideally be one that's unlikely to be used by another application so that testing on real hardware isn't too disruptive. I'd recommend tpmutil.Handle(15).

tpm2tools/keys_test.go
```diff
@@ -70,3 +72,184 @@ func BenchmarkStorageRootKeyRSA(b *testing.B) {
 		key.Close()
 	}
 }
+
```
josephlr: I think having the sealing tests in a seperate seal_test.go file would make more sense

tpm2tools/keys.go
```diff
@@ -133,3 +135,163 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+
+type sessionSummary struct {
+	OldDigest      tpmutil.RawBytes
+	CmdIDPolicyPCR uint32
+	NumPcrSels     uint32
+	Sel            tpmsPCRSelection
+	PcrDigest      tpmutil.RawBytes
+}
+
+func computePCRSessionAuth(rw io.ReadWriter, pcrs map[int][]byte) ([]byte, error) {
```
josephlr: Why is rw io.ReadWriter passed in here? isn't the session auth deterministicly computed from the PCR values alone?

tpm2tools/keys_test.go
```diff
@@ -70,3 +72,184 @@ func BenchmarkStorageRootKeyRSA(b *testing.B) {
...
+	defer rwc.Close()
+
+	key, err := StorageRootKeyRSA(rwc)
+	if err != nil {
+		t.Fatalf("can't create srk from template: %v", err)
+	}
+
+	secret := []byte("test")
+
+	if err != nil {
```
josephlr: What is this checking?

tpm2tools/keys_test.go
```diff
@@ -70,3 +72,184 @@ func BenchmarkStorageRootKeyRSA(b *testing.B) {
...
```
josephlr: What is this checking?

tpm2tools/keys_test.go
```diff
@@ -70,3 +72,184 @@ func BenchmarkStorageRootKeyRSA(b *testing.B) {
...
+
+	unseal, err := key.Unseal(rwc, []int{7}, priv, pub)
+	if err != nil {
+		t.Fatalf("failed to unseal: %v", err)
+	}
+	if !bytes.Equal(secret, unseal) {
+		t.Fatalf("unsealed (%v) not equal to secret (%v)", unseal, secret)
+	}
+
+	// extend pcr8 as a hack to get the intended value of pcr7 post extension.
```
josephlr: This commend is confusing, why are two PCRs ever used if we're only ever sealing/unsealing on a single PCR?

It might be more useful to just do the extension purely in software. Perhaps as an operation on some PCRValue struct.

tpm2tools/keys.go
```diff
@@ -133,3 +135,163 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+
+	if err != nil {
+		return tpm2.HandleNull, fmt.Errorf("failed to start auth session: %v", err)
+	}
+
+	sel := tpm2.PCRSelection{
+		Hash: tpm2.AlgSHA256,
+		PCRs: pcrs,
+	}
+	err = tpm2.PolicyPCR(rw, handle, nil, sel)
```
josephlr: Here and elsewhere, combine like:

if err := tpm2.PolicyPCR(rw, handle, nil, sel); err != nil {
  return tpm2.HandleNull, fmt.Errorf("auth step PolicyPCR failed: %v", err)
}

tpm2tools/keys.go
stevenrutherford: Done.

tpm2tools/keys_test.go
stevenrutherford: Done.

tpm2tools/keys_test.go
stevenrutherford: I could write the rest of predictive sealing, but I considered it out of scope for this change.

tpm2tools/keys.go
```diff
@@ -133,3 +135,163 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+	for pcr := range pcrs {
+		byteNum := pcr / 8
+		bytePos := byte(1 << byte(pcr%8))
+		pcrBits[byteNum] |= bytePos
+	}
+	pcrDigest := digestPCRList(pcrs)
+
+	summary := sessionSummary{
+		OldDigest:      make([]byte, sha256.Size),
+		CmdIDPolicyPCR: uint32(0x0000017f),
```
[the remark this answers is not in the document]
stevenrutherford: Yes, but it is private.

tpm2tools/keys.go
```diff
@@ -133,3 +135,163 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+
+func createPCRSession(rw io.ReadWriter, pcrs []int) (tpmutil.Handle, error) {
+	nonceIn := make([]byte, 16)
+	/* This session assumes the bus is trusted.  */
+	handle, _, err := tpm2.StartAuthSession(
+		rw,
+		tpm2.HandleNull,
+		tpm2.HandleNull,
+		nonceIn,
+		nil,
```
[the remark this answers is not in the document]
stevenrutherford: Done.

tpm2tools/keys.go
stevenrutherford: Done.

tpm2tools/keys.go
```diff
@@ -133,3 +135,163 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+
+	hash := crypto.SHA256.New()
+	hash.Write(b)
+	return hash.Sum(nil), nil
+}
+
+func digestPCRList(pcrs map[int][]byte) []byte {
+	hash := crypto.SHA256.New()
+	for i := 0; i < 24; i++ {
+		_, exists := pcrs[i]
```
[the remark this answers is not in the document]
stevenrutherford: Done.

tpm2tools/keys.go
```diff
@@ -133,3 +135,163 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+			PCRs: pcrBits,
+		},
+		PcrDigest: pcrDigest,
+	}
+	b, err := tpmutil.Pack(summary)
+	if err != nil {
+		return nil, fmt.Errorf("failed to pack for hashing: %v ", err)
+	}
+
+	hash := crypto.SHA256.New()
```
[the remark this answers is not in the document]
stevenrutherford: Done.

tpm2tools/keys.go
```diff
@@ -133,3 +135,163 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+type sessionSummary struct {
+	OldDigest      tpmutil.RawBytes
+	CmdIDPolicyPCR uint32
+	NumPcrSels     uint32
+	Sel            tpmsPCRSelection
+	PcrDigest      tpmutil.RawBytes
+}
+
+func computePCRSessionAuth(rw io.ReadWriter, pcrs map[int][]byte) ([]byte, error) {
+	pcrBits := make([]byte, 3)
```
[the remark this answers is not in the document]
stevenrutherford: Done. tpmutil.RawBytes is just another name for a slice of bytes.

tpm2tools/keys_test.go
stevenrutherford: Done.

tpm2tools/keys_test.go
stevenrutherford: Done.

tpm2tools/keys_test.go
stevenrutherford: Done.

tpm2tools/keys_test.go
stevenrutherford: Split into a helper for clarity.

tpm2tools/keys.go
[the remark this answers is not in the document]
josephlr: See https://github.com/google/go-tpm/pull/67

tpm2tools/keys.go
```diff
@@ -133,3 +135,160 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+	handle, _, err := tpm2.StartAuthSession(
+		rw,
+		tpm2.HandleNull,
+		tpm2.HandleNull,
+		nonceIn,
+		/*secret=*/ nil,
+		tpm2.SessionPolicy,
+		tpm2.AlgNull,
+		tpm2.AlgSHA256)
+
```
josephlr: Remove newline

tpm2tools/keys.go
```diff
@@ -133,3 +135,160 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+	if err != nil {
+		return tpm2.HandleNull, fmt.Errorf("failed to start auth session: %v", err)
+	}
+
+	sel := tpm2.PCRSelection{
+		Hash: tpm2.AlgSHA256,
+		PCRs: pcrs,
+	}
+
```
josephlr: Remove newline

tpm2tools/keys.go
```diff
@@ -133,3 +135,160 @@ func (k *Key) PublicKey() crypto.PublicKey {
 func (k *Key) Close() {
 	tpm2.FlushContext(k.rw, k.handle)
 }
+
+// Seal seals the sensitive byte buffer to the provided PCRs under the owner
+// hierarchy using the SHA256 versions of the provided PCRs.
+// The Key k is used as the parent key.
+func (k *Key) Seal(rw io.ReadWriter, pcrs []int, sensitive []byte) ([]byte, []byte, error) {
```
josephlr: It might be better to have a struct that contains both parts of the "Wrapped Key", namely it it the public template and the encrypted secret. I think it's needed to make it clear to users that `priv` is not a secret.

tpm2tools/keys.go
```diff
@@ -133,3 +135,160 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+		byteNum := pcr / 8
+		bytePos := byte(1 << byte(pcr%8))
+		pcrBits[byteNum] |= bytePos
+	}
+	pcrDigest := digestPCRList(pcrs)
+
+	summary := sessionSummary{
+		OldDigest:      make([]byte, sha256.Size),
+		CmdIDPolicyPCR: uint32(0x0000017f),
+		NumPcrSels:     uint32(1),
```
josephlr: Here and elsewhere, we don't need the type annotations. Your option to leave them in if you think it's clearer, but as the field names have a type, Go can do the right deductions.

tpm2tools/keys.go
```diff
@@ -133,3 +135,160 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+
+	digest := sha256.Sum256(b)
+	return digest[:], nil
+}
+
+func digestPCRList(pcrs map[int][]byte) []byte {
+	hash := crypto.SHA256.New()
+	for i := 0; i < 24; i++ {
+		if pcrValue, exists := pcrs[i]; exists {
+			binary.Write(hash, binary.BigEndian, pcrValue)
```
josephlr: hash.Write(pcrValue)

It's a byte array so endianness does not matter

stevenrutherford: Done.

tpm2tools/keys.go
stevenrutherford: Done.

tpm2tools/keys.go
stevenrutherford: Done.

tpm2tools/keys.go
stevenrutherford: Done.

tpm2tools/keys.go
stevenrutherford: Done.

tpm2tools/keys.go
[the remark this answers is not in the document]
stevenrutherford: Will fix this after resolving that issue.

tpm2tools/keys.go
```diff
@@ -133,3 +134,173 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+	for pcr := range pcrs {
+		byteNum := pcr / 8
+		bytePos := byte(1 << byte(pcr%8))
+		pcrBits[byteNum] |= bytePos
+	}
+	pcrDigest := digestPCRList(pcrs)
+
+	summary := sessionSummary{
+		OldDigest:      make([]byte, sha256.Size),
+		CmdIDPolicyPCR: 0x0000017f,
```
josephlr: https://github.com/google/go-tpm/pull/67 has been merged, so this can be changed.

tpm2tools/seal_test.go
```diff
@@ -0,0 +1,238 @@
...
+	rwc := internal.GetTPM(t)
+	defer rwc.Close()
+
+	key, err := StorageRootKeyRSA(rwc)
+	if err != nil {
+		t.Fatalf("can't create srk from template: %v", err)
+	}
+
+	secret := []byte("test")
+	pcrList := []int{1, 2, 3, 4, 5, 6, 7}
```
josephlr: Two things:
a) This should be a var defined above as it's used multiple times
b) Why are we extending 7 different PCRs here? Why not just extend a single PCR between our 2 unseal attempts?

tpm2tools/seal_test.go
```diff
@@ -0,0 +1,238 @@
...
+	pcrVal, err := tpm2.ReadPCR(rwc, pcrNum, tpm2.AlgSHA256)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read pcr %v", err)
+	}
+	return pcrVal, nil
+}
+
+func TestReseal(t *testing.T) {
+
+	extensions := [][]byte{
```
josephlr: Why 3 extensions?

tpm2tools/seal_test.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		t.Fatalf("unseal failed: %v", err)
+	}
+}
+
+func computePCRValue(t *testing.T, extensions [][]byte) ([]byte, error) {
+	rwc := internal.GetTPM(t)
+	defer rwc.Close()
+
+	pcrNum := 0
```
josephlr: I'm confused here, why does this use TPM commands at all? Can't we just do this extension just using go's sha256 hash?

tpm2tools/keys.go
```diff
@@ -133,3 +134,173 @@ func (k *Key) PublicKey() crypto.PublicKey {
...
+		return nil, fmt.Errorf("failed to load sealed object: %v", err)
+	}
+	defer tpm2.FlushContext(rw, sealed)
+
+	return tpm2.UnsealWithSession(rw, session, sealed, "")
+}
+
+// Reseal unwraps a secret and rewraps it under the auth value that would be
+// produced by the PCR state in pcrs. Similar to seal and unseal, this acts on
+// the SHA256 PCRs and uses the ownerhierarchy.
```
josephlr: owner hierarchy

tpm2tools/seal_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/google/go-tpm/tpmutil"
+
+	"github.com/google/go-tpm-tools/internal"
+)
+
+func TestSeal(t *testing.T) {
+	rwc := internal.GetTPM(t)
+	defer rwc.Close()
+
+	key, err := StorageRootKeyRSA(rwc)
```
josephlr: We do these two operations a bunch (GetTPM + StorageRootKeyRSA), might be good as a helper function.

We also need to make sure we are always calling key.Close() to prevent handle depletion. Something like:
```go
func getSRK(tb testing.TB) (io.ReadWriteCloser, *Key) {
        tb.Helper()
        // ...
}
func TestSeal(t *testing.T) {
	rwc, srk := getSRK(t)
	defer rwc.Close()
        defer srk.Close()
        // ...
}
```

josephlr: The tests are currently failing on real hardware, I think the handle leak might be why.

stevenrutherford: Fixed the handle leak.

stevenrutherford: Something feels off about combining those in to one function.
I'd still like to catch the handle leaks in some way other than hardware testing. Once the activeHandles method goes in from Sam's patch, we can query active handles on simulator.Close() and fail the test.

tpm2tools/keys.go
stevenrutherford: Done.

tpm2tools/seal_test.go
stevenrutherford: a) Moving this to be a file constant makes the tests seem more implicit, which I'd like to avoid. 
b) Using more than one seems like a negligibly more thorough test. Easy to change if you want it to be one.

tpm2tools/seal_test.go
stevenrutherford: I was trying to avoid implementing hash extension in this patch, as I mentioned in the previous thread, to keep this from patch from ballooning too much.

This is the third ambiguous comment here. If you want a change please state as much.

tpm2tools/keys.go
stevenrutherford: Done.

tpm2tools/seal_test.go
stevenrutherford: changed.

tpm2tools/seal_test.go
stevenrutherford: done.

tpm2tools/seal_test.go
stevenrutherford: Follow up PR will have handle leak checking.

PR #11, opened 2019-01-24
Add testing helpers and benchmarks

josephlr:
When testing go-tpm-tools on real hardware, I noticed that some operations take much more time to complete than others. There is also a great difference in the amount of time needed between real hardware and virtualized hardware on Google cloud. This PR should make getting performance numbers much easier.

First, a new package called `internal` is added to contain testing helper code (so that we can use it with multiple packages in go-tpm-tools, but external users can't inadvertently bring it in). Then the testing code is cleaned up to use the internal package and benchmarks.

This allows very easy benchmark testing. For example (on Linux):
```bash
# Compile the benchmarking code
-> go test ./tpm2tools -c

# Run against the simulator
-> ../tpm2tools.test -test.bench=. -test.run=^$
goos: linux
goarch: amd64
pkg: github.com/google/go-tpm-tools/tpm2tools
BenchmarkEndorsementKeyRSA-8   	     200	  14541982 ns/op
BenchmarkStorageRootKeyRSA-8   	      50	  29673783 ns/op
PASS

# Run against real hardware
-> sudo ./tpm2tools.test -test.bench=. -test.run=^$ -tpm-path=/dev/tpm0
goos: linux
goarch: amd64
pkg: github.com/google/go-tpm-tools/tpm2tools
BenchmarkEndorsementKeyRSA-8   	       1	3318950780 ns/op
BenchmarkStorageRootKeyRSA-8   	       1	7106487932 ns/op
PASS
```

### DIFF
--- a/internal/test_linux.go
+++ b/internal/test_linux.go
@@ -1,0 +1,32 @@
+package internal
+
+import (
+	"flag"
+	"io"
+	"testing"
+
+	"github.com/google/go-tpm/tpm2"
+
+	"github.com/google/go-tpm-tools/simulator"
+)
+
+var tpmPath = flag.String("tpm-path", "", "Path to Linux TPM character device (i.e. /dev/tpm0 or /dev/tpmrm0). Empty value (default) will run tests against the simulator.")
+
+// GetTPM is a cross-platform testing helper function that retrives the
+// appropriate TPM device from the flags passed into "go test".
+func GetTPM(tb testing.TB) io.ReadWriteCloser {
+	tb.Helper()
+	if *tpmPath != "" {
+		rwc, err := tpm2.OpenTPM(*tpmPath)
+		if err != nil {
+			tb.Fatalf("Opening TPM at %s failed: %v", *tpmPath, err)
+		}
+		return rwc
+	}
+
+	simulator, err := simulator.Get()
+	if err != nil {
+		tb.Fatalf("Initializing simulator failed: %v", err)
+	}
+	return simulator
+}
--- a/internal/test_linux.go
+++ b/internal/test_linux.go
@@ -10,7 +10,7 @@ import (
 	"github.com/google/go-tpm-tools/simulator"
 )
 
-var tpmPath = flag.String("tpm-path", "", "Path to Linux TPM character device (i.e. /dev/tpm0 or /dev/tpmrm0). Empty value (default) will run tests against the simulator.")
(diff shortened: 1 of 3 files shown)


internal/test_linux.go
```diff
@@ -0,0 +1,32 @@
...
+	"flag"
+	"io"
+	"testing"
+
+	"github.com/google/go-tpm/tpm2"
+
+	"github.com/google/go-tpm-tools/simulator"
+)
+
+var tpmPath = flag.String("tpm-path", "", "Path to Linux TPM character device (i.e. /dev/tpm0 or /dev/tpmrm0). Empty value (default) will run tests against the simulator.")
```
stevenrutherford: Nit: underscores for flag names are standard for most google-ish binaries.

internal/test_linux.go
```diff
@@ -0,0 +1,32 @@
...
+		rwc, err := tpm2.OpenTPM(*tpmPath)
+		if err != nil {
+			tb.Fatalf("Opening TPM at %s failed: %v", *tpmPath, err)
+		}
+		return rwc
+	}
+
+	simulator, err := simulator.Get()
+	if err != nil {
+		tb.Fatalf("Initializing simulator failed: %v", err)
```
stevenrutherford: "simulator initialization failed" or "initialization of simulator failed".

also, avoid caps at the start of errors (unless it's a proper noun).

stevenrutherford:
Looks good. A few nits.

internal/test_linux.go
josephlr: Fixed here and elsewhere. However per [go style guidelines](https://chromium.googlesource.com/chromiumos/platform/tast/+/HEAD/docs/writing_tests.md#formatting) errors never start with caps, but log statements (like those in `t.Fatal` or `b.Error`) always start with caps.

internal/test_linux.go
josephlr: Done, here and elsewhere

tpm2tools/keys_test.go
```diff
@@ -42,3 +42,33 @@ func TestCreateSigningKeysInHierarchies(t *testing.T) {
...
+	rwc := internal.GetTPM(b)
+	defer rwc.Close()
+	for n := 0; n < b.N; n++ {
+		b.StartTimer()
+		key, err := EndorsementKeyRSA(rwc)
+		b.StartTimer()
+		if err != nil {
+			b.Fatal(err)
+		}
+		key.Close()
```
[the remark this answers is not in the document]
josephlr: Done, here and elsewhere

PR #12, opened 2019-01-30
Add a benchmark for null hierarchy key creations

josephlr:
This will let us compare latency for RSA key creation across the three different hierarchies

### DIFF
--- a/tpm2tools/keys_test.go
+++ b/tpm2tools/keys_test.go
@@ -70,3 +70,19 @@ func BenchmarkStorageRootKeyRSA(b *testing.B) {
 		key.Close()
 	}
 }
+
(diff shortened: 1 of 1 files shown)


tpm2tools/keys_test.go
```diff
@@ -70,3 +70,19 @@ func BenchmarkStorageRootKeyRSA(b *testing.B) {
 		key.Close()
 	}
 }
+
+func BenchmarkNullSigningKeyRSA(b *testing.B) {
+	b.StopTimer()
+	rwc := internal.GetTPM(b)
+	defer rwc.Close()
+
```
stevenrutherford: Nit: delete this line to match other benchmarks.

PR #13, opened 2019-01-30
Fix handle leak in TestHandles

josephlr:
This test (when run on real hardware) would not properly free handles.


josephlr:
Obsoleted by https://github.com/google/go-tpm-tools/pull/18

PR #14, opened 2019-01-30
Make `internal` flags consistent with go-tpm

josephlr:
The flags in github.com/google/go-tpm all use dashes, as the flags in
the `internal` package are supposed to match, they should use dashes
as well.

For example:
  - in [go-tpm-tools](https://github.com/google/go-tpm-tools/blob/f38bf7b6e5c01f2820c5ae2017ef13c639e1546e/cmd/flush_handles/flush_handles.go#L13-L19) 
  - in [go-tpm](https://github.com/google/go-tpm/blob/612987aec54bad657671bbabd215093f123d8248/examples/tpm2-ekcert/main.go#L19-L23)

Related: https://github.com/google/go-tpm/pull/69

### DIFF
--- a/internal/test_linux.go
+++ b/internal/test_linux.go
@@ -10,7 +10,7 @@ import (
 	"github.com/google/go-tpm-tools/simulator"
 )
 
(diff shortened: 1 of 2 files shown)


josephlr:
Closing in favor of https://github.com/google/go-tpm-tools/pull/18

PR #15, opened 2019-01-31
Introduce CheckedClose

stevenrutherford:
Tests can now enforce error checking on tpm close.
Tests now fail if they leak TPM Handles.


josephlr:
Closing in favor of #19

PR #17, opened 2019-03-30
Add additional documentation for the command line tools

josephlr:
The `README` now discusses the two tools that live in `cmd`. It also make the `--help` output for these commands more informative and readable. Specifically:

```bash
> get_key --help
get_key - Write PEM formatted TPM public keys to stdout

  -hierarchy string
    	Hierarchy of the requested key. Valid options are endorsement,
    	platform, owner, and null. (default "endorsement")
  -template-index uint
    	NVRAM index of the key template. If this flag is not given,
    	the default RSA EK/SRK template is used.
  -tpm-path string
    	Path to a TPM character device or socket. (default "/dev/tpm0")
```

```bash
> flush_handles --help
flush_handles - Flush active TPM handles from the device

  -flush-all-types
    	Flush handles of all handle types. If enabled,
    	settings for other handle types are ignored.
  -flush-loaded-session
    	Flush all loaded session handles.
  -flush-saved-session
    	Flush all saved session handles.
  -flush-transient
    	Flush all transient handles. (default true)
  -tpm-path string
    	Path to a TPM character device or socket. (default "/dev/tpm0")
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ use with [Go-TPM](https://github.com/google/go-tpm):
     with this tool allows users to address this problem with out rebooting
     their machine.
   - `get_key`: a command line tool for querying TPM2 public keys. This tool can
-    use either the default key template or a template in NVRAM to retrive PEM
+    use either the default key template or a template in NVRAM to retrieve PEM
     formatted public keys specific to a user's TPM.
 
 ## Legal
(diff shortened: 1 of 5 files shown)


josephlr:
Obsoleted by https://github.com/google/go-tpm-tools/pull/18

PR #18, opened 2019-05-21
Move existing command to single "gotpm" binary

josephlr:
This change moves all of our existing command line tools to a single `cobra` CLI tool. This makes it easier to debug issues, see command usage, and add commands.

The cli tools has the following structure

- `gitub.com/google/gotpm/cmd`
  - Library containing all of the actual implementations of all the commands
  - Contains io/flag helper methods to make writing functions easier
  - Very small external API
    - `RootCmd`: the entry point
    - `ExternalTPM`:  allows an external simulator to be used

- `gitub.com/google/gotpm`
  - Actual binary package wrapping the `cmd` package
  - Only 3 lines of Go
- `gitub.com/google/gotpm/tests`
  - Tests against the external interface of the `gotpm` tool
  - Tests pass in arguments to the CLI tool, and make sure the commands succeed
  - Uses the simulator

This change also:
  - Fixes a few tests to use `Error` instead of `Fatal`
  - Uses `_other.go` files instead of `_linux.go` files to allow for building/testing on non Linux/Windows platforms
  - Fixes flags to use consistent style
  - Adds a `.gitignore`


internal/test_other.go
```diff
@@ -10,7 +12,7 @@ import (
 	"github.com/google/go-tpm-tools/simulator"
 )
 
-var tpmPath = flag.String("tpm_path", "", "Path to Linux TPM character device (i.e. /dev/tpm0 or /dev/tpmrm0). Empty value (default) will run tests against the simulator.")
+var tpmPath = flag.String("tpm-path", "", "Path to Linux TPM character device (i.e. /dev/tpm0 or /dev/tpmrm0). Empty value (default) will run tests against the simulator.")
```
twitchy-jsonp: Are these packages only part of the build for tests? Otherwise I would be concerned with the flag potentially conflicting with a real flag by the users' binary.

gotpm/cmd/flush.go
```diff
@@ -0,0 +1,74 @@
...
+			handles, err := tpm2tools.Handles(rwc, handleType)
+			if err != nil {
+				return fmt.Errorf("getting handles: %v", err)
+			}
+			for _, handle := range handles {
+				if err = tpm2.FlushContext(rwc, handle); err != nil {
+					return fmt.Errorf("flushing handle 0x%x: %v", handle, err)
+				}
+
+				fmt.Fprintf(debugOutput(), "Handle 0x%x flushed\n", handle)
```
twitchy-jsonp: The `debugOutput()` thing is a really neat pattern xD

internal/test_other.go
josephlr: This flag is only used for testing. I added some comments as to why this is the case, and how the conflict is avoided.

josephlr:
> This package is only meant for TPM2.0 tools, or is it more a case that we only need TPM2 tools for the timebeing?

I didn't set out to make the tool TPM2 specific (hence the `gotpm` name instead of `gotpm2`). I just don't know enough about TPM 1.2 to add support, but at least some of the operations we have planned should work with 1.2

josephlr:
Also, I didn't include tests for `gotpm pubkey`, but I was able to test it on [Google Cloud with Shielded VMs](https://cloud.google.com/security/shielded-cloud/retrieving-endorsement-key).

- Encryption Key outputs matched:
  - `sudo gotpm pubkey endorsement`
  - `encryptionKey` member of `gcloud compute instances get-shielded-instance-identity`
- Signing Key outputs matched:
  - `sudo gotpm pubkey endorsement --index=0x1c10001`
  - `signingKey` member of `gcloud compute instances get-shielded-instance-identity`

PR #23, opened 2019-07-19
Make SealedBytes a proto structure

josephlr:
This allows makes it easier to read/write the structure to disk.

We commit the generated `*.pb.go` files to be compatible with `go get`. 

This also changes `Unseal()` so that it no longer takes a `pcrs` argument, opting to instead put that data in the proto.

Data sealed with the previous version can still be unsealed by this version. You just have to manually set the fields in `SealedBytes`


josephlr:
@twitchy-jsonp and @stevenrutherford I've changed this PR some to make sure we have forward compatibility with introducing support for ECC SRKs or non sha256 hash functions.

PTAL, as some of the code has changed.

proto/tpm.proto
```diff
@@ -0,0 +1,44 @@
+syntax = "proto3";
+
+package proto;
+
+enum ObjectType {
```
stevenrutherford: As discussed oob, only list supported types.

josephlr: Done

PR #24, opened 2019-07-19
Add Seal and Unseal Commands

josephlr:
Depends on #31

Tests pass and manual invocations also succeed of real and virtual hardware.

### DIFF
--- a/gotpm/cmd/seal.go
+++ b/gotpm/cmd/seal.go
@@ -1,0 +1,123 @@
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	pb "github.com/golang/protobuf/proto"
+	"github.com/spf13/cobra"
+
+	"github.com/google/go-tpm-tools/proto"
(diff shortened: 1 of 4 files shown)


twitchy-jsonp:
7s seems consistent with my experience for generating an RSA key on a real TPM. Can you instead load the existing SRK from NVRAM and use that for testing?

gotpm/cmd/seal.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil
+	},
+}
+
+var unsealCmd = &cobra.Command{
+	Use:   "unseal",
+	Short: "Unseal some data previously sealed to the TPM",
+	Long: `Decrypt the input data using the TPM
+
+The opposite of "gotpm seal" ... TODO(joerichey): finish`,
```
stevenrutherford: What's up with this latent todo?

josephlr: Fixed

PR #27, opened 2019-07-25
Add SRK EK templates and gotpm pubkey support for ECC

lulumel0n:
Add SRK EK templates and seal test for ECC

### DIFF
--- a/tpm2tools/seal_test.go
+++ b/tpm2tools/seal_test.go
@@ -3,8 +3,8 @@ package tpm2tools
 import (
 	"bytes"
 	"crypto/sha256"
-	"testing"
 	"io"
+	"testing"
 
 	"github.com/google/go-tpm/tpm2"
 	"github.com/google/go-tpm/tpmutil"
@@ -20,14 +20,12 @@ func TestSeal(t *testing.T) {
 		name   string
 		getSRK func(io.ReadWriter) (*Key, error)
 	}{
-			{"RSA", StorageRootKeyRSA},
-			{"ECC", StorageRootKeyECC},
+		{"RSA", StorageRootKeyRSA},
+		{"ECC", StorageRootKeyECC},
 	}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-
 			srk, err := test.getSRK(rwc)
-
 			if err != nil {
 				t.Fatalf("can't create %s srk from template: %v", test.name, err)
 			}
@@ -36,12 +34,12 @@ func TestSeal(t *testing.T) {
 			secret := []byte("test")
 			pcrList := []int{7, 23}
 			pcrToExtend := tpmutil.Handle(23)
-		
+
 			sealed, err := srk.Seal(pcrList, secret)
 			if err != nil {
 				t.Fatalf("failed to seal: %v", err)
 			}
-		
+
 			unseal, err := srk.Unseal(sealed)
 			if err != nil {
 				t.Fatalf("failed to unseal: %v", err)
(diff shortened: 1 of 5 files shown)


tpm2tools/template.go
```diff
@@ -47,6 +47,20 @@ func defaultRSADecrypt() *tpm2.RSAParams {
 	}
 }
 
+func defaultECCDecrypt() *tpm2.ECCParams {
```
josephlr: Calling them `default*Decrypt` was a bit confusing on my part `defaultRSAParams` and `defaultECCParams` are better names, so I'd change them.

tpm2tools/template.go
```diff
@@ -47,6 +47,20 @@ func defaultRSADecrypt() *tpm2.RSAParams {
...
+	return &tpm2.ECCParams{
+		Symmetric:  defaultSymScheme(),
+		CurveID: tpm2.CurveNISTP256,
+		Point: tpm2.ECPoint {
+			XRaw: make([]byte, 32),
+			YRaw: make([]byte, 32),
+		},
+		KDF: &tpm2.KDFScheme{
```
josephlr: Per the [documentation for `ECCParams`](https://godoc.org/github.com/google/go-tpm/tpm2#ECCParams), the `KDF` param is allowed to be `nil`, so you can just not set it here (`nil` is the default value).

If the `KDFScheme` is `nil`. [it still gets encoded as `AlgNull`](https://github.com/google/go-tpm/blob/ae6dd98980d43b6b1a71c9737490896ca6e6e9ed/tpm2/structures.go#L537)

tpm2tools/seal_test.go
```diff
@@ -15,39 +15,60 @@ func TestSeal(t *testing.T) {
...
-	defer key.Close()
-
-	secret := []byte("test")
-	pcrList := []int{7, 23}
-	pcrToExtend := tpmutil.Handle(23)
-
-	sealed, err := key.Seal(pcrList, secret)
-	if err != nil {
-		t.Fatalf("failed to seal: %v", err)
+	type SRKey struct {
```
josephlr: So these table driven subtests are a little messy, one way to improve them would be use annonymous structs, and just pass the function that creates the key to the test. Specifically

```go
tests := []struct {
	name   string
	getSRK func(io.ReadWriter) (*Key, error)
}{
        {"RSA", StorageRootKeyRSA},
        {"ECC", StorageRootKeyECC},
}

for _, test := range tests {
        t.Run(test.name, func(t *testing.T) {
                srk, err := test.getSRK(rwc)
                // Check errors and use keys normally
        }
}
```

tpm2tools/template.go
lulumel0n: Done

tpm2tools/template.go
lulumel0n: Done

tpm2tools/seal_test.go
lulumel0n: Done

tpm2tools/seal_test.go
```diff
@@ -15,39 +16,52 @@ func TestSeal(t *testing.T) {
...
+	tests := []struct {
+		name   string
+		getSRK func(io.ReadWriter) (*Key, error)
+	}{
+			{"RSA", StorageRootKeyRSA},
+			{"ECC", StorageRootKeyECC},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+
```
josephlr: Delete blank line

tpm2tools/seal_test.go
```diff
@@ -15,39 +16,52 @@ func TestSeal(t *testing.T) {
...
+			srk, err := test.getSRK(rwc)
+
```
josephlr: Delete blank line

tpm2tools/seal_test.go
```diff
@@ -15,39 +16,52 @@ func TestSeal(t *testing.T) {
...
-
-	// unseal should not succeed.
-	_, err = key.Unseal(sealed)
-	if err == nil {
-		t.Fatalf("unseal should have caused an error: %v", err)
+	tests := []struct {
+		name   string
+		getSRK func(io.ReadWriter) (*Key, error)
+	}{
+			{"RSA", StorageRootKeyRSA},
```
josephlr: Is this indentation right, can you run `go fmt` on the whole thing.

tpm2tools/seal_test.go
```diff
@@ -15,39 +16,52 @@ func TestSeal(t *testing.T) {
...
+			}
+		
+			extension := bytes.Repeat([]byte{0xAA}, sha256.Size)
+			err = tpm2.PCRExtend(rwc, pcrToExtend, tpm2.AlgSHA256, extension, "")
+			if err != nil {
+				t.Fatalf("failed to extend pcr: %v", err)
+			}
+		
+			// unseal should not succeed.
+			_, err = srk.Unseal(sealed)
```
josephlr: Move this into the if statement (I realize that this is how it was before, but we want to be consistent).

PR #30, opened 2019-07-29
Add ECC algo for seal/unseal cmds

lulumel0n:


tpm2tools/keys.go
```diff
@@ -49,6 +49,18 @@ func EndorsementKeyFromNvIndex(rw io.ReadWriter, idx uint32) (*Key, error) {
 	return KeyFromNvIndex(rw, tpm2.HandleEndorsement, idx)
 }
 
+// Load SRK based on the given tpm2.Algorithm.
+func LoadSRK(rwc io.ReadWriter, algo tpm2.Algorithm) (*Key, error) {
```
josephlr: I think it would be more reasonable to have `getSRK` and `getEK` helper functions in `gotpm/cmd/utils.go`. We can move these to the public API of `tpm2tools` later if that becomes necessary.

gotpm/cmd/utils.go
```diff
@@ -85,3 +94,17 @@ func dataInput() io.Reader {
...
 }
+
+// Get the algorithm for public key.
+func getAlgo(keyAlgo string) (tpm2.Algorithm, error) {
+	switch keyAlgo {
+	case "rsa":
+		return tpm2.AlgRSA, nil
+	case "ecc":
+		return tpm2.AlgECC, nil
+	case "":
```
josephlr: I made a mistake adding this `case`. Just handle `""` in the default.

gotpm/cmd/pubkey.go
```diff
@@ -68,27 +68,11 @@ NVDATA instead (and --algo is ignored).`,
...
-	case "rsa":
-		return tpm2.AlgRSA, nil
-	case "ecc":
-		return tpm2.AlgECC, nil
-	case "":
-		panic("--algo flag not properly setup")
-	default:
-		return tpm2.AlgNull, fmt.Errorf("invalid argument %q for \"--algo\" flag", keyAlgo)
-	}
+	addPublicKeyAlgoFlag(pubkeyCmd)
```
josephlr: Good to use the `addPublicKeyAlgoFlag` helper functions here.

We should also use the `getEK` and `getAIK` helper methods in `getKey` below.

gotpm/cmd/utils.go
lulumel0n: removed case ""

gotpm/cmd/pubkey.go
lulumel0n: I also added getEKwithAlgo and getSRKwithAlgo, for use cases like we need to provide the algorithm directly instead of referring from the Flag value. For example, when Unsealing, we need to get the algo from the sealed file instead of the Flag value.

gotpm/cmd/utils.go
```diff
@@ -96,15 +97,53 @@ func dataInput() io.Reader {
...
 }
+
+func getSRKwithAlgo(rwc io.ReadWriter, algo tpm2.Algorithm) (*tpm2tools.Key, error) {
+	switch algo {
+	case tpm2.AlgRSA:
+		return tpm2tools.StorageRootKeyRSA(rwc)
+	case tpm2.AlgECC:
+		return tpm2tools.StorageRootKeyECC(rwc)
+	default:
+		return nil, fmt.Errorf("Cannot create SRK for the given algorithm.")
```
josephlr: Add the invalid algorithm in the error message. i.e. a `%v` in the `Errorf`

gotpm/cmd/utils.go
```diff
@@ -96,15 +97,53 @@ func dataInput() io.Reader {
...
+}
+
+func getEKwithAlgo(rwc io.ReadWriter, algo tpm2.Algorithm) (*tpm2tools.Key, error) {
+	switch algo {
+	case tpm2.AlgRSA:
+		return tpm2tools.EndorsementKeyRSA(rwc)
+	case tpm2.AlgECC:
+		return tpm2tools.EndorsementKeyECC(rwc)
+	default:
+		return nil, fmt.Errorf("Cannot create EK for the given algorithm.")
```
josephlr: Same here

gotpm/cmd/pubkey.go
```diff
@@ -82,27 +82,13 @@ func getKey(rw io.ReadWriter, hierarchy tpmutil.Handle, algo tpm2.Algorithm) (*t
...
-			return nil, fmt.Errorf("There is no default ECC key for this hierarchy")
-		}
+	switch hierarchy {
+	case tpm2.HandleEndorsement:
+		return getEK(rw)
+	case tpm2.HandleOwner:
+		return getSRK(rw)
 	default:
-		panic("Unreachable")
+		return nil, fmt.Errorf("There is no default key for this hierarchy")
```
josephlr: Put the `hierarchy` in this error message.

tpm2tools/keys.go
lulumel0n: Done

gotpm/cmd/utils.go
lulumel0n: Okay, but here the algorithm will be printed as an integer which won't add that much information.

josephlr: That’s ok, we can get %v to work properly in a future pr

lulumel0n: I think %x hex probably makes more sense here, as in tpm2 those constants are represented by hex

josephlr: Good catch! That’s correct

gotpm/cmd/pubkey.go
lulumel0n: done

josephlr:
So this got closed due to #24 getting merged. @lulumel0n, can you reopen a PR with all of this rebased onto master? Sorry for the mixup.

PR #32, opened 2019-07-30
Add ECC algo for seal/unseal cmds

lulumel0n:
Add "algo" cmd flag for the "seal" cmd
"unseal" cmd will infer the algorithm from the sealed proto
Refactor getKey() in pubkey.go


gotpm/tests/seal_test.go
```diff
@@ -38,10 +38,14 @@ func TestSealPlain(t *testing.T) {
 
 	tests := []struct {
 		name   string
+		algo   string
 		pcrArg string
 	}{
-		{"Plain", ""},
-		{"PCRs", "7,8"},
+		{"Default", "", ""},
```
josephlr: I like this structure!

I don’t think we need 5 subtests here though.

RSA, RSAWithPCRs, and ECCWithPCRs is enough

lulumel0n: done

PR #33, opened 2019-08-01
Add random CMD for gotpm

lulumel0n:
Add a command to generate random bytes from the TPM.

Nit: also fixed 2 capitalizations in the description of the flag.


josephlr:
So this is a good think we might want to add. However, I think that the first step should be to add functionality to `tpm2tools` that implements an [`io.Reader`](https://golang.org/pkg/io/#Reader) interface around `TPM2_GetRandom`.

This would be functionally similar to [`crypto/rand.Reader`](https://golang.org/pkg/crypto/rand/#pkg-variables)

Once we have this, building on top of the RNG will be much easier.

josephlr:
Closing per offline discussion

PR #36, opened 2019-09-04
Add server and client Import workflows

BenDHillier:
Add server.CreateImportRequest for a server side Import workflow.
Add tpm2tools.Import for a client side Import workflow.
Add import testing in server/import_test.go. Only add the one test as there's no way to test the server side and client side workflows separately (Need to run client side to verify server side worked correctly, and vice versa).

Also add a function for creating tpm2.Public objects from crypto.PublicKey's and a corresponding test.


tpm2tools/import.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, err
+	}
+	ek, err := getEK(*request.Algorithm, rw)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to retrieve EK: %v", err)
+	}
+
+	defer ek.Close()
+
+	session, _, err := tpm2.StartAuthSession(
```
BenDHillier: Do we want all of these values hard-coded or will different users have different auth setups?

server/import.go
```diff
@@ -0,0 +1,192 @@
...
+
+	packedSeed, err := tpmutil.Pack(encSeed)
+	if err != nil {
+		return nil, err
+	}
+	return packedSeed, nil
+}
+
+func createRandomSeed(ek tpm2.Public) []byte {
+	symBlockSize := ek.RSAParameters.Symmetric.KeyBits / 8
```
josephlr: Strictly speaking, this is the key size, not the block size, as for AES block size is always 128 bits even when using AES 256. I would just call this `keySize` or `seedSize`.

server/import.go
```diff
@@ -0,0 +1,192 @@
...
+	seed := make([]byte, symBlockSize)
+	io.ReadFull(rand.Reader, seed)
```
josephlr: Here (and in other calls to `rand.Reader`) we need to be checking errors. It's OK to just panic if the call fails, as not having RNG is a very rare event regardless.

server/import.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func createPublic(private tpm2.Private, hashConsts hashConstants) tpm2.Public {
+	publicHash := hashConsts.create()
+	publicHash.Write(private.SeedValue)
+	publicHash.Write(private.Sensitive)
+	return tpm2.Public{
+		Type:       tpm2.AlgKeyedHash,
+		NameAlg:    hashConsts.alg,
+		Attributes: tpm2.FlagUserWithAuth,
```
josephlr: Should we have any additional attributes set here?

server/import.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
+	"hash"
+	"io"
+
+	pb "github.com/google/go-tpm-tools/proto"
+)
+
+// hashConstants contains all the per-EK constant hash values in a convenient struct.
+type hashConstants struct {
```
josephlr: This seems unnecessary,

We should just pass `ek.NameAlg` (a `tpm2.Algorithm`) everywhere and use the `HashConstructor` and `Size` methods when necessary.

Right now we aren't consistent. Sometimes we pass a `hash.Hash` sometimes a `hashConstants`, sometimes a `tpm2.Algorithm`. If you want things to be more readable a helper functions like:

```go
func getHash(alg tpm2.Algorithm) hash.Hash {
	create, err := alg.HashConstructor()
	if err != nil {
		panic(err)
	}
	return create()
}
```

might be useful. Then you can just pass `alg` everywhere and use `alg`, `getHash(alg)`, and `getHash(alg).Size()`

server/import.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, err
+	}
+	return tpmutil.Pack(tpm2.IDObject{
+		IntegrityHMAC: macSum,
+		EncIdentity:   encryptedSecret,
+	})
+}
+
+func createHMAC(encryptedSecret, nameEncoded, seed []byte, hashAlg tpm2.Algorithm) ([]byte, error) {
+	macKey, err := tpm2.KDFa(hashAlg, seed, "INTEGRITY" /*contextU=*/, nil /*contextV=*/, nil, getHash(hashAlg).Size()*8)
```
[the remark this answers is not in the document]
BenDHillier: go fmt moves them over for some reason. I was thinking I might just get rid of them because I really don't like them being before them comma either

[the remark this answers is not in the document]
BenDHillier: putting the args on multiple lines resolved this.

tpm2tools/keys_test.go
```diff
@@ -117,6 +118,37 @@ func TestKeyCreation(t *testing.T) {
 	}
 }
 
+func TestCreatePublicAreaFromPublicKey(t *testing.T) {
```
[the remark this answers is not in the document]
BenDHillier: Would comparing the name be sufficient for checking for exact public area match, since it's a hash of the public area? For generating a key from the TPM I used EndorsmentKeyRSA() which returns a Key, but Key types don't give public access to their tpm2.Public so I can't easily do a comparison.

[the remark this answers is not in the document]
josephlr: Name comparison is fine!

PR #38, opened 2019-09-05
Revert #37

josephlr:
Cirrus isn't working for some reason


fkorotkov:
@josephlr you should make sure it's installed for this repository here: https://github.com/apps/cirrus-ci. Maybe admin previously only installed it for particular repositories. 🤷‍♂

josephlr:
> @josephlr you should make sure it's installed for this repository here: https://github.com/apps/cirrus-ci. Maybe admin previously only installed it for particular repositories.

Yup that's correct. I'm not sure why I need admin approval for Cirrus but not for Travis or Google Cloud Build, but it's not a big deal. One weird thing is that going to https://github.com/apps/cirrus-ci/installations/new and trying to configure the Google Org installation gives a 404 instead of a permission denied error. I guess it's just a GitHub bug.

fkorotkov:
Might be some GitHub issues. I also see some 500s on some pages.

As far as I know Travis still defaults to OAuth installation method rather then the new GitHub App which provides granual reposioty access and some fancy APIs like GitHub Checks API. 

There is an option for a GitHub App be installed for all repositories so maybe the admin had a recent security training and was extra cautious in case of Cirrus but not in case of the Google's own Cloud Build. 😅

PR #42, opened 2019-09-09
Delete apt cache from Docker image

josephlr:
See https://github.com/google/go-tpm/pull/126#discussion_r322009896


twitchy-jsonp:
If you want to remove cached debs, consider doing `apt-get clean` as well.

PR #46, opened 2019-09-19
Add PCR Policy to Import

BenDHillier:
Allows users to require the TPM to have specific PCR values in order to unseal the ImportBlob.


josephlr:
Can you rebase this PR and use the [`Pcrs`](https://github.com/google/go-tpm-tools/blob/33ab828ac16f5e8e8fea8e0ae7dc4a1c1089990a/proto/tpm.proto#L35-L38) proto to send/use lists of PCR values?

BenDHillier:
> Can you rebase this PR and use the [`Pcrs`](https://github.com/google/go-tpm-tools/blob/33ab828ac16f5e8e8fea8e0ae7dc4a1c1089990a/proto/tpm.proto#L35-L38) proto to send/use lists of PCR values?

Rebased. Were we wanting to let the user choose the hash alg now? Because otherwise, I don't see a reason to use the Pcrs proto.

server/import.go
```diff
@@ -70,19 +82,26 @@ func createPrivate(sensitive []byte, hashAlg tpm2.Algorithm) tpm2.Private {
...
+	public = tpm2.Public{
+		Type:    tpm2.AlgKeyedHash,
+		NameAlg: hashAlg,
 		KeyedHashParameters: &tpm2.KeyedHashParams{
 			Alg:    tpm2.AlgNull,
 			Unique: publicHash.Sum(nil),
 		},
 	}
+	if len(pcrs) > 0 {
+		public.AuthPolicy, err = tpm2tools.ComputePCRSessionAuth(pcrs)
```
josephlr: We should set AdminWithPolicy in this case.

josephlr:
> Were we wanting to let the user choose the hash alg now? Because otherwise, I don't see a reason to use the Pcrs proto.

I think for API consistancy we would want to. Using the `Pcrs` proto everywhere is makes sense will make the API more consistent.

An idea might be to move `ComputePCRSessionAuth` to `pcrs.go` and have it take the `Pcrs` proto. If we do this, then we will also want to have `Reseal` take the Proto object as well.

BenDHillier:
> An idea might be to move `ComputePCRSessionAuth` to `pcrs.go` and have it take the `Pcrs` proto. If we do this, then we will also want to have `Reseal` take the Proto object as well.

Ok, I'm thinking it'd make sense to put that in a separate PR before submitting this PR.

PR #47, opened 2019-09-19
Add gotpm read pcr.

stevenrutherford:
* Adds new command and subcommands for read and read pcr.
* Adds misc.go to store oddball wrappers for commands like tpm2.ReadPCRs.


josephlr:
Two quick things,

- Rebase this onto master (to get rid of the conflict)
- Can you make it so that if no `--pcrs` flag is provided, we just read them all?

tpm2tools/misc.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"io"
+
+	"github.com/google/go-tpm-tools/proto"
+	"github.com/google/go-tpm/tpm2"
+)
+
+// ReadPCRs fetchs the values of the specified PCRs for the specified hash.
+func ReadPCRs(rw io.ReadWriter, pcrs []int, hash tpm2.Algorithm) (*proto.Pcrs, error) {
```
josephlr: Does this need to return a proto? Right now we only use protos when we have to pass data to/from the disk or to/from the server.

Right now we normally just pass PCR values as a map from index to value. Would that work as the return value here?

tpm2tools/misc_test.go
```diff
@@ -0,0 +1,43 @@
+package tpm2tools
```
josephlr: These two tests should be [table-driven](https://github.com/golang/go/wiki/TableDrivenTests) to prevent duplication.

tpm2tools/misc_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"bytes"
+	"crypto/sha1"
+	"crypto/sha256"
+	"testing"
+
+	"github.com/google/go-tpm-tools/internal"
+	"github.com/google/go-tpm/tpm2"
+)
+
```
josephlr: We should also add a test that makes sure we get the right value after extending a PCR (this could just be a component of your table driven tests.

tpm2tools/misc.go
stevenrutherford: Obviously that would work, but the pattern established elsewhere is to return protos where possible. I find this somewhat natural for testability.

For example, seal returns the sealedbytes proto instead of an equivalent go structure. Note, this contains more than just the map, it's also tagged with the hash algorithm. Not that it is hard to return both pieces of data, it's that this seems pretty comparable.

stevenrutherford:
> * Can you make it so that if no `--pcrs` flag is provided, we just read them all?
(In the future it's probably easier to have changes be threaded conversations or whatever github calls them, so we can make sure everything is resolved)

I've done this (and I think it improves usability), but I'm a little weary, since this leads to inconsistent defaults for flags.

tpm2tools/misc_test.go
stevenrutherford: Done. Only extended PCR0, but this should be fine. 

I'm tempted to extend the tests to compare the marshaled protos to strings, but I'm not sure it's worth how gross the table driven tests would need to look.

tpm2tools/misc_test.go
stevenrutherford: Done.

tpm2tools/misc_test.go
josephlr: No need to compare the marshaled strings, but compute the out values (like with `sha256.Sum(whatever)`).

tpm2tools/misc_test.go
```diff
@@ -0,0 +1,62 @@
...
+	defer CheckedClose(t, rwc)
+	pcrCount, err := GetPCRCount(rwc)
+	if err != nil {
+		t.Fatalf("Failed to fetch pcr count: %v", err)
+	}
+
+	if pcrCount != 24 {
+		t.Fatalf("tpm simulator has unexpected number of PCRs: %v instead of 24", pcrCount)
+	}
+
```
josephlr: Delete this line

tpm2tools/misc_test.go
```diff
@@ -0,0 +1,62 @@
...
+func TestGetPCRCount(t *testing.T) {
+	rwc := internal.GetTPM(t)
+	defer CheckedClose(t, rwc)
+	pcrCount, err := GetPCRCount(rwc)
+	if err != nil {
+		t.Fatalf("Failed to fetch pcr count: %v", err)
+	}
+
+	if pcrCount != 24 {
+		t.Fatalf("tpm simulator has unexpected number of PCRs: %v instead of 24", pcrCount)
```
josephlr: Errorf

tpm2tools/misc_test.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+		}
+
+		proto, err := ReadPCRs(rwc, test.inPcrs, test.inAlg)
+		if err != nil {
+			t.Fatalf("failed to read pcrs %v", err)
+		}
+
+		if !bytes.Equal(proto.Pcrs[0], test.out) {
+			t.Fatalf("%v not equal to expected %v", proto.Pcrs[0], test.out)
```
josephlr: Errorf

tpm2tools/misc_test.go
```diff
@@ -0,0 +1,62 @@
...
+	{[]int{0}, tpm2.AlgSHA256, [][]byte{bytes.Repeat([]byte{0x00}, sha256.Size)}, []byte{245, 165, 253, 66, 209, 106, 32, 48, 39, 152, 239, 110, 211, 9, 151, 155, 67, 0, 61, 35, 32, 217, 240, 232, 234, 152, 49, 169, 39, 89, 251, 75}},
+}
+
+func TestReadPCRs(t *testing.T) {
+	rwc := internal.GetTPM(t)
+	defer CheckedClose(t, rwc)
+
+	for _, test := range tests {
+		for _, extension := range test.inExtensions {
+			err := tpm2.PCRExtend(rwc, tpmutil.Handle(0), test.inAlg, extension, "")
```
josephlr: Combine lines here

tpm2tools/misc_test.go
```diff
@@ -0,0 +1,88 @@
...
+	{tpm2.AlgSHA1, [][]byte{bytes.Repeat([]byte{0x02}, sha1.Size)}},
+	{tpm2.AlgSHA256, nil},
+	{tpm2.AlgSHA256, [][]byte{bytes.Repeat([]byte{0x00}, sha256.Size)}},
+	{tpm2.AlgSHA256, [][]byte{bytes.Repeat([]byte{0x01}, sha256.Size)}},
+	{tpm2.AlgSHA256, [][]byte{bytes.Repeat([]byte{0x02}, sha256.Size)}},
+}
+
+func pcrExtend(alg tpm2.Algorithm, old, new []byte) ([]byte, error) {
+	var h hash.Hash
+	switch alg {
```
josephlr: This switch isn't needed, we have [`Algorithm.HashConstructor`](https://godoc.org/github.com/google/go-tpm/tpm2#Algorithm.HashConstructor)

tpm2tools/misc_test.go
```diff
@@ -0,0 +1,88 @@
...
+	{tpm2.AlgSHA1, [][]byte{bytes.Repeat([]byte{0x00}, sha1.Size)}},
+	{tpm2.AlgSHA1, [][]byte{bytes.Repeat([]byte{0x01}, sha1.Size)}},
+	{tpm2.AlgSHA1, [][]byte{bytes.Repeat([]byte{0x02}, sha1.Size)}},
+	{tpm2.AlgSHA256, nil},
+	{tpm2.AlgSHA256, [][]byte{bytes.Repeat([]byte{0x00}, sha256.Size)}},
+	{tpm2.AlgSHA256, [][]byte{bytes.Repeat([]byte{0x01}, sha256.Size)}},
+	{tpm2.AlgSHA256, [][]byte{bytes.Repeat([]byte{0x02}, sha256.Size)}},
+}
+
+func pcrExtend(alg tpm2.Algorithm, old, new []byte) ([]byte, error) {
```
josephlr: As we only pass hardcoded algos here, just panic in `pcrExtend` if we (somehow) get a bad algo.

tpm2tools/misc_test.go
stevenrutherford: TBH, I might be confused, but I'd still need a switch in that case to convert the TPM alg type to the go alg type.

tpm2tools/misc_test.go
stevenrutherford: Agreed, but that seems unidiomatic, and not worth fixing.

gotpm/cmd/read.go
```diff
@@ -0,0 +1,85 @@
...
+	Long:  `Read from the TPM`,
+	Args:  cobra.NoArgs,
+}
+
+var pcrCmd = &cobra.Command{
+	Use:   "pcr",
+	Short: "Read PCRs from the TPM",
+	Long: `Read PCRs from the TPM
+
+Based on the --pcrs flag, this reads the contents of the TPM's PCRs.
```
josephlr: I would rephrase this to clarify that there is always _some_ PCR selection and always _some_ `hashAlgo` selection.

Something like:
```
Read PCRs from the TPM

Read the contents of each PCR specified with --pcrs. If --pcrs is not provided,
all pcrs in the bank are read.

This command only supports reading from a single PCR bank. The bank can be
changed using the --hash-algo flag.
```

gotpm/cmd/utils.go
```diff
@@ -55,6 +57,11 @@ func addPublicKeyAlgoFlag(cmd *cobra.Command) {
 		"Public key algorithm, \"rsa\" or \"ecc\"")
 }
 
+func addHashAlgoFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&hashAlgo, "hashAlgo", "sha256",
```
josephlr: `hash-algo` to be like `tpm-path`.

tpm2tools/misc.go
```diff
@@ -0,0 +1,46 @@
+package tpm2tools
```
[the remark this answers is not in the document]
stevenrutherford: Ack: will do this as a final step, but won't until we're done with the rest (to avoid marking the entire file as needing to be reviewed).

tpm2tools/misc_test.go
stevenrutherford: Done.

PR #50, opened 2019-09-26
Add support for importing using ECC EKs

BenDHillier:
This change adds support for importing using an ECC EK.

Supports NIST-P224, NIST-P256, NIST-P384, and NIST-P521.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,10 +2,8 @@ module github.com/google/go-tpm-tools
 
 go 1.12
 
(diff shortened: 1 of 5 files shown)


josephlr:
@BenDHillier I rebased this onto #51, PTAL

BenDHillier:
> @BenDHillier I rebased this onto #51, PTAL

Rebase looks good

PR #53, opened 2019-10-04
Add client Quote and server Verify

jkl73:
Quote: allow a signing key to Quote a list of pcrs and some extra data
Verify: will verify the quoted data and extra data


josephlr:
Now that https://github.com/google/go-tpm/pull/142 has been merged can you incorporate that into this PR?

server/verify.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("signature decoding failed: %v", err)
+	}
+	if rsaPubKey, ok := pubkey.(*rsa.PublicKey); ok {
+		if sig.Alg != tpm2.AlgRSASSA && sig.Alg != tpm2.AlgRSAPSS {
+			return nil, fmt.Errorf("given signature is not a RSA signature: %v", sig.Alg)
+		}
+
+		var attestationHash hash.Hash
+		if sig.RSA.HashAlg == tpm2.AlgSHA256 {
```
josephlr: Should we be hardcoding this hash algo?

server/verify.go
```diff
@@ -0,0 +1,84 @@
...
+// the PCR digest for quoted data, which can insure that the given PCR state is the
+// same as when signing this attestation. This function will return the "Extra Data"
+// appended in the attestation when quoting if passed all verification.
+func Verify(pubkey crypto.PublicKey, quote *proto.Quote, pcrs *proto.Pcrs) ([]byte, error) {
+	sig, err := tpm2.DecodeSignature(bytes.NewBuffer(quote.GetRawSig()))
+	if err != nil {
+		return nil, fmt.Errorf("signature decoding failed: %v", err)
+	}
+	if rsaPubKey, ok := pubkey.(*rsa.PublicKey); ok {
+		if sig.Alg != tpm2.AlgRSASSA && sig.Alg != tpm2.AlgRSAPSS {
```
josephlr: Don't we need different verification methods for SSA vs PSS?

See: https://golang.org/pkg/crypto/rsa/#VerifyPSS

lulumel0n: Yes we need, I over looked this part..

server/verify.go
lulumel0n: You mean we should support all RSA hash algorithms?
I only added sha256, because I think all tpm 2.0 default key templates have sha256 as hashalgo.

lulumel0n: Yes we need, I over looked this part..

server/verify.go
lulumel0n: Yes we need, I over looked this part..

server/verify.go
lulumel0n: You mean we should support all RSA hash algorithms?
I only added sha256, because I think all tpm 2.0 default key templates have sha256 as hashalgo.

josephlr: We should (at least initially) just support whatever hash algorithm we get with sig.RSA.HashAlg. This is what we do in other places as well

server/verify.go
lulumel0n: I only added support for AlgRSASSA for now, since that the scheme used by the default AIK template

server/verify.go
lulumel0n: Yes currently this only support SHA256 which is used by the default template , otherwise it will throw exception

josephlr:
Can you rebase this onto master and use the merged changes from https://github.com/google/go-tpm/pull/143

go.sum
```diff
@@ -15,9 +15,15 @@ github.com/google/go-tpm v0.2.1-0.20190910203116-33a9c3f38379 h1:spUrNOxKZPzgPd/
...
 github.com/google/go-tpm v0.2.1-0.20190926184827-9f71d680f4ab/go.mod h1:gTv8GNuqS7CI+tQWrpt5BMMaD5W3G+dZULQLhhAKT5c=
+github.com/google/go-tpm v0.2.1-0.20191010051517-9737ea7ab8c4 h1:vxxGeCiuu+A0BTehA8rbiTYQZ1u0tqMtjGyZMeQRhuY=
+github.com/google/go-tpm v0.2.1-0.20191010051517-9737ea7ab8c4/go.mod h1:gTv8GNuqS7CI+tQWrpt5BMMaD5W3G+dZULQLhhAKT5c=
+github.com/google/go-tpm v0.2.1-0.20191015210219-431489f43254 h1:ahwXCP5Ui0eYBO9nM71/BzWwVtnGjfZ4OsnDE0V2pMA=
+github.com/google/go-tpm v0.2.1-0.20191015210219-431489f43254/go.mod h1:gTv8GNuqS7CI+tQWrpt5BMMaD5W3G+dZULQLhhAKT5c=
 github.com/google/go-tpm-tools v0.0.0-20190906225433-1614c142f845/go.mod h1:AVfHadzbdzHo54inR2x1v640jdi1YSi3NauM2DUsxk0=
 github.com/hashicorp/hcl v1.0.0/go.mod h1:E5yfLk+7swimpb2L/Alb/PJmXilQ/rhwaUYs4T20WEQ=
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
+github.com/lulumel0n/go-tpm v0.1.2-0.20191004021704-b54438b68c90 h1:0K5NghSfoiHokcIwvZPEMCmMKRl8Ao3RBlBE4kkrJuY=
+github.com/lulumel0n/go-tpm v0.1.2-0.20191004021704-b54438b68c90/go.mod h1:gTv8GNuqS7CI+tQWrpt5BMMaD5W3G+dZULQLhhAKT5c=
```
josephlr: These lines should not be in the go.sum

server/verify.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+			hashConstructor := hash.New()
+			hashConstructor.Write(quote.GetQuote())
+			if err := rsa.VerifyPKCS1v15(rsaPubKey, hash, hashConstructor.Sum(nil), sig.RSA.Signature); err != nil {
+				return nil, fmt.Errorf("failed to verify the signature: %v", err)
+			}
+		default:
+			return nil, fmt.Errorf("given signature scheme is not supported: %x, only RSASSA-PKCS1-v1_5 is supported", sig.Alg)
+		}
+	} else {
```
josephlr: I would put the error branch earlier to make the code easier to read:
```go
rsaPubKey, ok := pubkey.(*rsa.PublicKey)
if !ok {
        // Sorry, only RSA works right now
}
if sig.Alg != tpm2.AlgRSASSA {
        // Sorry, only RSASSA works right now
}

//  Hashing and PCKS v1.5 verification

```

go.sum
lulumel0n: removed

tpm2tools/pcr.go
```diff
@@ -55,3 +58,21 @@ func ReadPCRs(rw io.ReadWriter, pcrs []int, hash tpm2.Algorithm) (*proto.Pcrs, e
 
 	return &pl, nil
 }
+
+// ComputePCRDigest will take in a PCR proto and compute the SHA256 digest based on the
+// given PCR proto.
+// Followed PCRComputeCurrentDigest() in the TPM spec.
+func ComputePCRDigest(pcrs *proto.Pcrs) ([]byte, error) {
+	var pcrDigestHash hash.Hash
+	pcrDigestHash = sha256.New()
```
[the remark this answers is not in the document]
lulumel0n: added a parameter that can pass in tpm2.Algorithm

server/verify.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	if err := validatePCRDigest(attestedQuoteInfo, pcrs); err != nil {
+		return nil, err
+	}
+	return attestationData.ExtraData, nil
+}
+
+func validatePCRDigest(quoteInfo *tpm2.QuoteInfo, pcrs *proto.Pcrs) error {
+	quotedPCRList := quoteInfo.PCRSelection.PCRs
+	sort.Ints(quotedPCRList)
```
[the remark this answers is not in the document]
lulumel0n: Forget to clean, it is not necessary here.

server/verify.go
```diff
@@ -0,0 +1,80 @@
...
+			hashConstructor := hash.New()
+			hashConstructor.Write(quote.GetQuote())
+			if err := rsa.VerifyPKCS1v15(rsaPubKey, hash, hashConstructor.Sum(nil), sig.RSA.Signature); err != nil {
+				return nil, fmt.Errorf("failed to verify the signature: %v", err)
+			}
+		default:
+			return nil, fmt.Errorf("given signature scheme is not supported: %x, only RSASSA-PKCS1-v1_5 is supported", sig.Alg)
+		}
+	} else {
+		return nil, fmt.Errorf("given public key is not supported yet, only RSA with RSASSA-PKCS1-v1_5 scheme is supported")
```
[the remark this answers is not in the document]
lulumel0n: doen

server/verify.go
lulumel0n: done

josephlr:
@lulumel0n can you rebase this onto the current master (which is now the [`v0.2.0` release](https://github.com/google/go-tpm-tools/releases/tag/v0.2.0))

lulumel0n:
@josephlr Done rebasing, PTAL

alexmwu:
Rebased on the tpm2tools -> client package name change

server/verify.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/google/go-tpm-tools/client"
+	tpmpb "github.com/google/go-tpm-tools/proto"
+	"github.com/google/go-tpm/tpm2"
+)
+
+// Verify checks an AK's signature on a Quote against the quoted data and extra data.
+// Then, it matches that quoted data (PCR digest) against a group of PCRs.
+//
+// Verify supports ECDSA and RSASSA signature verification.
+func Verify(pubKey crypto.PublicKey, quote *tpmpb.Quote, pcrs *tpmpb.Pcrs, extraData []byte) error {
```
alexmwu, line 20: Thoughts on VerifyQuote? I was thinking we may want a high level VerifyAttestation that encapsulates the PCRs, quotes, and event log.

josephlr: Ya this is good, and checks the right stuff.

josephlr:
@googlebot I consent.

josephlr (approved):
I added some stuff, but for the most part this looks good.

Things I added:

- https://github.com/google/go-tpm-tools/pull/53/commits/b6b869cf5fa2624784b5620a2fc05d1464a337e3: Check that we have a restricted signing key _before_ calling Quote
  - This gives better error messages
  - It also prevents inadvertent misuse with an unrestricted signing key (which would be insecure)
- https://github.com/google/go-tpm-tools/pull/53/commits/b629070ead2058349bf7c6abc4373e9eedb3a52b: Additional Quote tests
- https://github.com/google/go-tpm-tools/pull/53/commits/cf8eb4bb3b5fa028bac5154cc3ab96d180fdb9f7: Compare nonces in constant time
- https://github.com/google/go-tpm-tools/pull/53/commits/808eb78402fb8669e9ec2a2c1f4f4294757a123e: Use `HasSamePCRSelection` helper
  - This also makes sure that our hash bank matches between the desired selection and the quote

@alexmwu I didn't add tests cases for all of this stuff, so it might be useful to go back (later) and add any missing tests for things where I added checks.

alexmwu:
sg, we should at minimum add tests for the restricted signing key check and perhaps the constant time check

PR #56, opened 2019-10-24
Add TpmSigner 

BenDHillier:
TpmSigner only supports RSA PKCS#1 v1.5 signing. Subsequent PRs will add PSS and ECC support.

There will also be a PR for adding a (k *Key) GetSigner function, which will check that the Key is appropriate to be used as a Signer.

Also enables SHA512 in the simulator. The simulator told me to set MAX_CONTENT_SIZE to 1360 when I tried to run it after enabling SHA512.

### DIFF
--- a/tpm2tools/signer.go
+++ b/tpm2tools/signer.go
@@ -1,0 +1,77 @@
+package tpm2tools
+
+import (
+	"crypto"
+	"fmt"
+	"io"
+	"sync"
+
+	"github.com/google/go-tpm/tpm2"
+)
+
+// Global mutex to protect against concurrent TPM access.
+var signerMutex sync.Mutex
+
+type tpmSigner struct {
+	Key  *Key
+	Hash crypto.Hash
+}
+
+// Public returns the tpmSigners public key.
+func (signer *tpmSigner) Public() crypto.PublicKey {
+	return signer.Key.PublicKey()
+}
+
+// Sign uses the TPM key to  sign the digest.
+// The digest must be hashed from the same hash algorithm as the keys scheme.
+// The opts hash function must also match the keys scheme.
+// Concurrent use of Sign is thread safe, but it is not safe to access the TPM
+// from other sources while Sign is executing.
+func (signer *tpmSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
+	if opts.HashFunc() != signer.Hash {
+		return nil, fmt.Errorf("opts hash: %v does not match the keys signing hash: %v", opts.HashFunc(), signer.Hash)
+	}
+	if len(digest) != signer.Hash.Size() {
+		return nil, fmt.Errorf("digest length: %d does not match hash size: %d", digest, signer.Hash.Size())
+	}
+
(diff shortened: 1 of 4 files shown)


tpm2tools/signer.go
```diff
@@ -0,0 +1,81 @@
...
+
+// TpmSigner implements the crypto.Signer interface for TPM Keys.
+// Concurrent use of one or more TpmSigners is thread safe, but it is not safe
+// to read/write to the TPM from other sources while using a TpmSigner.
+type TpmSigner struct {
+	key *Key
+}
+
+// Public returns the TpmSigners public key.
+func (signer TpmSigner) Public() crypto.PublicKey {
```
josephlr: If we add another member to `TpmSigner`, Here and elsewhere, we should use pointer recivers for the methods, as that's more typical for `crypto.Signer` implementations.

BenDHillier: I don't really see a reason to use a pointer receiver, since TpmSigner is essentially already a pointer to a Key and crypto.Hash is just a uint so not much additional overhead.

tpm2tools/signer_test.go
```diff
@@ -0,0 +1,87 @@
...
+	return template
+}
+
+func TestSign(t *testing.T) {
+	rwc := internal.GetTPM(t)
+	defer CheckedClose(t, rwc)
+
+	tests := []struct {
+		name     string
+		hash     crypto.Hash
```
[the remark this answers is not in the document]
BenDHillier: Were you thinking the crypto.Hash member should be public? I was thinking it would be private, but I don't really like using private members in tests.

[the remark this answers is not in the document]
BenDHillier: We'd talked about this offline a bit, and we'd decided on the TpmSigner becoming private but having its fields public. I'm going to leave the hash field in the test struct though, as the TpmSigner hash field still isn't part of the public API, so shouldn't be used in unit tests.

tpm2tools/signer.go
```diff
@@ -0,0 +1,77 @@
...
+	Key  *Key
+	Hash crypto.Hash
+}
+
+// Public returns the tpmSigners public key.
+func (signer *tpmSigner) Public() crypto.PublicKey {
+	return signer.Key.PublicKey()
+}
+
+// Sign uses the TPM key to  sign the digest.
```
josephlr: Delete extra space.

tpm2tools/signer.go
```diff
@@ -0,0 +1,77 @@
...
+// The digest must be hashed from the same hash algorithm as the keys scheme.
+// The opts hash function must also match the keys scheme.
+// Concurrent use of Sign is thread safe, but it is not safe to access the TPM
+// from other sources while Sign is executing.
+func (signer *tpmSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
```
josephlr: For consistency with https://golang.org/pkg/crypto/rsa/#PrivateKey.Sign, we should make sure `opts` isn't a `*PSSOptions`.

PR #55, opened 2019-10-17
Refactor of Sealing/Unsealing, expose helper methods

BenDHillier:
Export ComputePCRSessionAuth and add ComputePCRDigest

ComputePCRSessionAuth was modified to take a proto.Pcrs and the hash algorithm used to create the digest.

PR #57, opened 2019-10-25
Support CertifyCreation for Seal/Unseal

lulumel0n:
* Changed Seal() Unseal() Reseal() APIs to accept SealingConfig and CertifyConfig
* Removed duplicated tests in ./gotpm

I saw some changes may have conflict/duplicate with #55 , I will update this PR when #55 is merged.
 
Known Issue:
* When try to seal/reseal with an empty pcr selection, seems like the computePCRSessionAuth() is not yielding the correct Auth digest, which will cause the unsealing to fail. Currently I am by passing this by calling getPCRSessionAuth() directly from the TPM()


lulumel0n:
@googlebot I fixed it

lulumel0n:
Closing. Create a new PR #60

PR #61, opened 2019-11-20
Change gotpm cmd input pcrs from []int to []uint

lulumel0n:
Current gotpm cmd allow user to input negative int for a PCR# without any exception/warning (negative value will be ignored).

This change should prevent any negative PCR# input.


josephlr:
@lulumel0n can you rebase this?

josephlr:
Closing in favor of #66

PR #62, opened 2019-11-22
Add ECC support to tpmSigner

BenDHillier:
As per the crypto.Signer spec, ecdsa signatures are returned in a DER-serialised, ASN.1 signature structure (and is done the same way used in the ecdsa lib).

### DIFF
--- a/tpm2tools/signer.go
+++ b/tpm2tools/signer.go
@@ -83,10 +83,10 @@ func (k *Key) GetSigner() (crypto.Signer, error) {
 		return nil, fmt.Errorf("unsupported key type: %v", k.pubArea.Type)
 	}
 	if sigScheme == nil {
-		return nil, fmt.Errorf("GetSigner called on key missing a signing scheme.")
+		return nil, fmt.Errorf("key missing required signing scheme")
 	}
 	if sigScheme.Alg != tpm2.AlgRSASSA && sigScheme.Alg != tpm2.AlgECDSA {
-		return nil, fmt.Errorf("unsupported signing algorithm: %v.", sigScheme.Alg)
+		return nil, fmt.Errorf("unsupported signing algorithm: %v", sigScheme.Alg)
 	}
 	hash, err := sigScheme.Hash.Hash()
 	if err != nil {
--- a/tpm2tools/signer.go
+++ b/tpm2tools/signer.go
@@ -49,11 +49,14 @@ func (signer *tpmSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts
 		return nil, err
 	}
 
-	if sig.RSA != nil {
+	switch sig.Alg {
+	case tpm2.AlgRSASSA:
 		return sig.RSA.Signature, nil
-	} else {
+	case tpm2.AlgECDSA:
 		sigStruct := struct{ R, S *big.Int }{sig.ECC.R, sig.ECC.S}
 		return asn1.Marshal(sigStruct)
+	default:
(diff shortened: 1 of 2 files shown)


josephlr:
@BenDHillier can you rebase this?

BenDHillier:
> @BenDHillier can you rebase this?

rebased

tpm2tools/signer.go
```diff
@@ -67,13 +72,23 @@ func (k *Key) GetSigner() (crypto.Signer, error) {
...
+	var sigScheme *tpm2.SigScheme
+	switch k.pubArea.Type {
+	case tpm2.AlgRSA:
+		sigScheme = k.pubArea.RSAParameters.Sign
+	case tpm2.AlgECC:
+		sigScheme = k.pubArea.ECCParameters.Sign
+	default:
+		return nil, fmt.Errorf("unsupported key type: %v", k.pubArea.Type)
+	}
+	if sigScheme == nil {
```
josephlr: So what I would do here is have the switch statement assign to a "supported algorithm" variable, and then check it below. Like:
```go
var sigScheme *tpm2.SigScheme
var sigAlg tpm2.Algorithm

switch k.pubArea.Type {
case tpm2.AlgRSA:
	sigScheme = k.pubArea.RSAParameters.Sign
	sigAlg = tpm2.AlgRSASSA
case tpm2.AlgECC:
	sigScheme = k.pubArea.ECCParameters.Sign
	sigAlg = tpm2.AlgECDSA
default:
	return nil, fmt.Errorf("unsupported key type: %v", k.pubArea.Type)
}

if sigScheme == nil {
	return nil, fmt.Errorf("key missing required signing scheme")
}
if sigScheme.Alg != sigAlg {
	return nil, fmt.Errorf("unsupported signing algorithm: %v", sigScheme.Alg)
}

hash, err := sigScheme.Hash.Hash()
if err != nil {
	return nil, err
}
return &tpmSigner{k, hash}, nil
```

tpm2tools/signer.go
```diff
@@ -46,7 +48,13 @@ func (signer *tpmSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts
...
 		return nil, err
 	}
-	return sig.RSA.Signature, nil
+
+	if sig.RSA != nil {
+		return sig.RSA.Signature, nil
+	} else {
+		sigStruct := struct{ R, S *big.Int }{sig.ECC.R, sig.ECC.S}
+		return asn1.Marshal(sigStruct)
+	}
```
josephlr: This should be a switch statement on `sig.Alg`. Doing one of:

- `AlgRSASSA` => `return sig.RSA.Signature, nil`
- `AlgECDSA` => `return asn1.Marshal(struct{ R, S *big.Int }{sig.ECC.R, sig.ECC.S})`
- panic

tpm2tools/signer.go
```diff
@@ -46,7 +48,16 @@ func (signer *tpmSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts
...
-	return sig.RSA.Signature, nil
+
+	switch sig.Alg {
+	case tpm2.AlgRSASSA:
+		return sig.RSA.Signature, nil
+	case tpm2.AlgECDSA:
+		sigStruct := struct{ R, S *big.Int }{sig.ECC.R, sig.ECC.S}
+		return asn1.Marshal(sigStruct)
+	default:
+		return nil, fmt.Errorf("unsupported signing algorithm: %v", sig.Alg)
```
josephlr: This should be a panic, as it would only happen if go-tpm-tools had a bug somewhere else.

PR #64, opened 2019-12-12
cirrus: Use golangci-lint for presubmits

josephlr:
This will check: formatting, style, imports, usage, etc...


josephlr:
@lulumel0n can you add the fixes that only the google3 internal liner complains about? Specifically:
  - Package comments
  - import "proto" things as "pb"

lulumel0n:
> @lulumel0n can you add the fixes that only the google3 internal liner complains about? Specifically:
> 
> * Package comments
> * import "proto" things as "pb"

Will be fixed in #65

PR #66, opened 2019-12-14
Use custom flags for PCRs and Algos

josephlr:
Both our alogrithms flags and our pcrs flags have custom validity
semantics. They are more than just plain int or string flags. This also
allows use to catch PCR range errors.

We also hardcode support for 24 PCRs and rename `utils.go` to `flags.go`


josephlr:
Note: this was Pair Programmed with @lulumel0n

PTAL, please also be sure to "Rebase" instead of "Squash"

gotpm/cmd/flags.go
```diff
@@ -14,11 +16,82 @@ var (
...
+func (f *algoFlag) Type() string {
+	return "algo"
+}
+
+func (f *algoFlag) String() string {
+	return algos[*f.value]
+}
+
+// Allowed gives a string list of the permitted algorithm values for this flag.
+func (f *algoFlag) Allowed() string {
```
lulumel0n: Should this method be private?

PR #67, opened 2019-12-19
Cleanup file structure and fix Windows

josephlr:
Notes: We can now just run `go build ./cmd/gotpm` and everything just works. Also, **don't** merge this PR online, I want to try some manual merging signature stuff.

### gitignore: Ignore built binaries in all directories

Before, we were only ignoring built binaries (tests, gotpm, etc...) in
specific directories. This makes future changes less fragile.

### windows: Fix build and fix cross-compiling

Right now it is hard to cross compile testing binaries as our simulator
uses C code, and golang cross compiling doesn't work with CGO.

We also fix some Windows bugs and add a check in the CI for building
on Windows.

### cmd: Move ./gotpm to ./cmd/gotpm

Our current filesystem structure is non-standard. Normally, go modules
have a "cmd" directory containing all of the binaries built as
subdirectories. We invert this.

Fix to be more standard.


simulator/internal/internal_cross.go
```diff
@@ -0,0 +1,20 @@
...
+	"io"
+)
+
+// SetSeeds
+func SetSeeds(r io.Reader) {}
+
+// Reset
+func Reset(forceManufacture bool) {}
+
+// RunCommand passes cmd to the simulator and returns the simulator's response.
```
lulumel0n: Since the function always returns an error, the comments should just say this is a place holder function

josephlr: Added some comments

PR #68, opened 2019-12-20
Package gotpm for Arch Linux

josephlr:
This adds Arch Linux support.

We now support installing the binary as well as adding initramfs hooks for unlocking TPM secrets at boot time. This process has some Arch Linux specific components.

Note that incorporating Debian configs is not yet ready (as Debian packaging is much more complex. 

Note that (right now) we use my work PGP key for verifying the package. This allows us to avoid updating sha256 sums every time we bump the package.

### DIFF
--- a/files/boot-unseal.sh
+++ b/files/boot-unseal.sh
@@ -16,21 +16,24 @@ for device in $(blkid -o device); do
     # the key is unsealed by the TPM, does not mean it will unlock the disk. We
     # write the unsealed key to the in-memory rootfs, it is not written to disk.
     for f in /mnt/esp/*/disk_unlock_keys/*.sealed; do
-        if [ -f "$f" ] && gotpm unseal --input "$f" --output "/crypto_keyfile.bin" ; then
-            echo "gotpm: Unsealed $f"
-            key_found=1
-            break
+        if [ -f "$f" ]; then
+            if gotpm unseal --input "$f" --output "/crypto_keyfile.bin" ; then
+                echo "Unsealed ${f#/mnt/esp}"
+                key_found=1
+                break
+            else
+                echo "Failed to unseal ${f#/mnt/esp}"
+            fi
         fi
     done
     umount $device
 
     if [ $key_found -ne 0 ]; then
-        break
+        exit 0
     fi
 done
 
-if [ $key_found -eq 0 ]; then
-    echo "gotpm: Unable to unseal any TPM disk unlock key"
-fi
+echo "Unable to unseal any TPM disk unlock key"
+exit 1
 
 # vim: set ft=sh ts=4 sw=4 et:
--- a/files/boot-unseal.sh
+++ b/files/boot-unseal.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/ash
 
-unlocked=0
+key_found=0
 # Loop through all devices to find the ESP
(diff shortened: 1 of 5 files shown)


josephlr:
It's up: https://aur.archlinux.org/packages/gotpm/

files/boot-unseal.sh
```diff
@@ -0,0 +1,23 @@
...
+
+    # Temporarily mount the ESP to read disk unlock keys
+    mkdir -p /mnt/esp
+    mount -t vfat $device /mnt/esp
+
+    # Try all of the keys in the appropriate folder
+    for f in /mnt/esp/*/disk_unlock_keys/*.sealed; do
+        if gotpm unseal --input "$f" --output "/crypto_keyfile.bin" ; then
+            echo "gotpm: Unsealed $f"
+        fi
```
lulumel0n: What if the first unsealed key cannot decrypt the disk? Will the system allow the script to try the next key in disk_unlock_keys?

josephlr: I added some additional to avoid unsealing multiple keys, and just succeeding when the first one is unsealed.

However, due to the structure of Arch Linux's initramfs hooks, we cannot check the disk unlock status without modifying the default `encrypt` hook shipped with the OS (like we do for Debian).

The `encrypt` hook supports an in-memory keyfile (defaulting to `/crypto_keyfile.bin`), but does not support _multiple_ keyfiles, so we just have to only try the first key and then fall back to password if that doesn't work.

It's not ideal, but it allows us to avoid reimplementing the main Arch Linux `encrypt` hook.

files/boot-unseal.sh
```diff
@@ -0,0 +1,34 @@
...
+    part_type=$(blkid -p $device -s PART_ENTRY_TYPE -o value)
+    if [ "$part_type" != "c12a7328-f81f-11d2-ba4b-00a0c93ec93b" ]; then
+        continue
+    fi
+
+    # Temporarily mount the ESP to read disk unlock keys
+    mkdir -p /mnt/esp
+    mount -t vfat $device /mnt/esp
+
+    # Try all of the keys in the appropriate folder
```
lulumel0n: Then the comment should be something like "Try to find a candidate key", since we are not going through all keys

files/boot-unseal.sh
```diff
@@ -0,0 +1,34 @@
...
+    done
+    umount $device
+
+    if [ $unlocked -ne 0 ]; then
+        break
+    fi
+done
+
+if [ $unlocked -eq 0 ]; then
+    echo "gotpm: Unable to unseal TPM disk unlock key"
```
lulumel0n: We should have a different name for $unlocked, since the found key may not "unlock" the disk.

Can call it $found

josephlr: I went with `$key_found`

files/boot-unseal.sh
josephlr: Done, added an explanation of what exactly we're doing (any why it might not work).

josephlr:
@lulumel0n note that I also added a `if [ -f "$f" ]` check in the file loop this avoids issues with:
- the `/mnt/esp/*/disk_unlock_keys/*.sealed` not matching anything
- subdirectories of the  `disk_unlock_keys` directory.

PR #70, opened 2020-01-09
Import RSA signing keys with PCR policies

BenDHillier:
Add importing for RSA signing keys.

The import flow returns a *Key to the user, which can be used to get a tpmSigner to sign digests.

### DIFF
--- a/server/import.go
+++ b/server/import.go
@@ -17,6 +17,7 @@ import (
 	"github.com/google/go-tpm/tpmutil"
 
 	tpmpb "github.com/google/go-tpm-tools/proto"
+	"github.com/google/go-tpm-tools/tpm2tools"
 )
 
 // CreateImportBlob uses the provided public EK to encrypt the sensitive data.
@@ -28,10 +29,36 @@ func CreateImportBlob(ekPub crypto.PublicKey, sensitive []byte, pcrs *tpmpb.Pcrs
 	if err != nil {
 		return nil, err
 	}
-	private := createPrivate(sensitive, ek.NameAlg)
-	public := createPublic(private, ek.NameAlg, pcrs)
+	private := createPrivate(sensitive)
+	public := createPublic(private)
+
+	return createImportBlobHelper(ek, public, private, pcrs)
+}
+
+// CreateSigningKeyImportBlob uses the provided public EK to encrypt the signing
+// key into import blob format. The returned import blob can be used to import
+// the signing key into the TPM associated with the provided EK without exposing
+// the private area to the TPM's OS using the tpm2tools Key.ImportSigningKey()
+// method. A non-nil pcrs parameter adds a requirement that the TPM must have
+// specific PCR values to use the signing key.
+func CreateSigningKeyImportBlob(ekPub crypto.PublicKey, signingKey crypto.PrivateKey, pcrs *tpmpb.Pcrs) (*tpmpb.ImportBlob, error) {
+	ek, err := CreateEKPublicAreaFromKey(ekPub)
+	if err != nil {
+		return nil, err
+	}
+	public, private, err := createPublicPrivateSign(signingKey)
+	if err != nil {
+		return nil, err
+	}
+
+	return createImportBlobHelper(ek, public, private, pcrs)
+}
+
+func createImportBlobHelper(ek, public tpm2.Public, private tpm2.Private, pcrs *tpmpb.Pcrs) (*tpmpb.ImportBlob, error) {
+	setPublicAuth(&public, pcrs)
 
 	var seed, encryptedSeed []byte
+	var err error
 	switch ek.Type {
 	case tpm2.AlgRSA:
 		seed, encryptedSeed, err = createRSASeed(ek)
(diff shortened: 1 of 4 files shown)


josephlr:
@BenDHillier We should rebase this on #73 and use the consolidated handling to use the existing Signer code with these imported keys.

tpm2tools/import.go
```diff
@@ -6,44 +6,82 @@ import (
...
+
+	auth, err = k.session.Auth()
+	if err != nil {
+		return tpm2.HandleNull, err
+	}
+	handle, _, err := tpm2.LoadUsingAuth(rw, k.Handle(), auth, blob.PublicArea, private)
+	if err != nil {
+		return tpm2.HandleNull, fmt.Errorf("load failed: %s", err)
+	}
+	return handle, nil
```
josephlr: Let's move this above `Import`. It makes the diff waaayyy more readable.

tpm2tools/import.go
```diff
@@ -6,44 +6,82 @@ import (
...
 	if err != nil {
-		return nil, fmt.Errorf("load failed: %s", err)
+		return nil, fmt.Errorf("unseal failed: %s", err)
 	}
-	defer tpm2.FlushContext(rw, handle)
+	return out, nil
+}
 
-	unsealSession, err := newPCRSession(rw, PCRSelection(blob.Pcrs))
+func (k *Key) ImportSigningKey(rw io.ReadWriter, blob *tpmpb.ImportBlob) (*Key, error) {
```
josephlr: We need a comment for this function

josephlr: We don't need the `rw io.ReadWriter` parameter. We can just use `k.rw` as part of the key.

Yes, I realize we had it as part of `Import`, but that was a mistake.

josephlr: I'll remove the the param from Import as part of a breaking change.

server/import.go
```diff
@@ -32,7 +32,33 @@ func CreateImportBlob(ekPub crypto.PublicKey, sensitive []byte, pcrs *tpmpb.Pcrs
 	private := createPrivate(sensitive, ek.NameAlg)
 	public := createPublic(private, ek.NameAlg, pcrs)
 
+	return createImportBlobHelper(ek, public, private, pcrs)
+}
+
+// CreateSigningKeyImportBlob uses the provided public EK to encrypt the signing
```
josephlr: Like `CreateImportBlob`, we should mention the actual client-side command that can be used `tpm2tools Key.ImportSigningKey`.

server/import.go
```diff
@@ -89,6 +115,38 @@ func createPublic(private tpm2.Private, hashAlg tpm2.Algorithm, pcrs *tpmpb.Pcrs
 			Unique: publicHash.Sum(nil),
 		},
 	}
+	setPublicAuth(&public, pcrs)
+	return public
+}
+
+func createPrivatePublicRSA(rsaPriv *rsa.PrivateKey, nameAlg tpm2.Algorithm, pcrs *tpmpb.Pcrs) (tpm2.Private, tpm2.Public, error) {
```
josephlr: We should use public then private for consistancy with other methdos

server/import.go
```diff
@@ -89,6 +115,38 @@ func createPublic(private tpm2.Private, hashAlg tpm2.Algorithm, pcrs *tpmpb.Pcrs
...
+			KeyBits:     uint16(rsaPub.N.BitLen()),
+			ExponentRaw: uint32(rsaPub.E),
+			ModulusRaw:  rsaPub.N.Bytes(),
+			Sign: &tpm2.SigScheme{
+				Alg:  tpm2.AlgRSASSA,
+				Hash: tpm2.AlgSHA256,
+			},
+		},
+	}
+	setPublicAuth(&public, pcrs)
```
josephlr: Do we need this if it's already set in `createImportBlobHelper`

josephlr: Nevermind, this is part of createPublic

server/import.go
```diff
@@ -89,6 +115,38 @@ func createPublic(private tpm2.Private, hashAlg tpm2.Algorithm, pcrs *tpmpb.Pcrs
 			Unique: publicHash.Sum(nil),
 		},
 	}
+	setPublicAuth(&public, pcrs)
```
josephlr: Can we just call this in the `createImportBlobHelper`? That way we won't forget it as part of other import types.

josephlr: We can also then avoid needing to pass around `pcrs` as much

server/import.go
josephlr: We could make this:

`createPublicPrivateSign(priv crypto.PrivateKey) -> (tpm2.Public, tpm2.Private, error)`

This method would then do the appropriate casting of crypto.PrivateKey to the correct type. This also makes adding ECC support easier.

server/import.go
```diff
@@ -32,7 +32,33 @@ func CreateImportBlob(ekPub crypto.PublicKey, sensitive []byte, pcrs *tpmpb.Pcrs
...
+func CreateSigningKeyImportBlob(ekPub crypto.PublicKey, signingKey crypto.PrivateKey, pcrs *tpmpb.Pcrs) (*tpmpb.ImportBlob, error) {
+	ek, err := CreateEKPublicAreaFromKey(ekPub)
+	if err != nil {
+		return nil, err
+	}
+	rsaKey, ok := signingKey.(*rsa.PrivateKey)
+	if !ok {
+		return nil, fmt.Errorf("unsupported signing key type: %T", signingKey)
+	}
+	private, public, err := createPrivatePublicRSA(rsaKey, ek.NameAlg, pcrs)
```
josephlr: I think we can avoid passing the NameAlg around in this function completely. It can't be changed by the user, so we can just do:

`var defaultNameAlg = tpm2tools.DefaultEKTemplateRSA().NameAlg`

and just use `defaultNameAlg` wherever we need it.

server/import.go
josephlr: A helper function like this should be moved to `key_conversion.go` to avoid cluttering `import.go`

PR #71, opened 2020-01-09
Add PSS support for Signer

BenDHillier:
This change is mostly just writing test code to verify RSAPSS signing works as expected, particularly for salt lengths.

### DIFF
--- a/tpm2tools/signer.go
+++ b/tpm2tools/signer.go
@@ -35,8 +35,13 @@ func (signer *tpmSigner) Public() crypto.PublicKey {
 // digestSize in which case, saltLen will be digestSize. The only normal case
 // where saltLen is not digestSize is when using 1024 keyBits with SHA512.
 func (signer *tpmSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
-	if _, ok := opts.(*rsa.PSSOptions); ok {
-		return nil, fmt.Errorf("signing with non-default PSS options not supported")
+	if pssOpts, ok := opts.(*rsa.PSSOptions); ok {
+		if signer.Key.pubArea.RSAParameters.Sign.Alg != tpm2.AlgRSAPSS {
+			return nil, fmt.Errorf("invalid options. PSSOptions cannot be used with signing alg: %v", signer.Key.pubArea.RSAParameters.Sign.Alg)
+		}
+		if pssOpts.SaltLength != rsa.PSSSaltLengthAuto {
+			return nil, fmt.Errorf("salt length must be rsa.PSSSaltLengthAuto")
+		}
 	}
 	if opts.HashFunc() != signer.Hash {
 		return nil, fmt.Errorf("opts hash: %v does not match the keys signing hash: %v", opts.HashFunc(), signer.Hash)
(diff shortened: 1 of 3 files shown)


tpm2tools/signer.go
```diff
@@ -82,7 +88,10 @@ func (k *Key) GetSigner() (crypto.Signer, error) {
 	switch k.pubArea.Type {
 	case tpm2.AlgRSA:
 		sigScheme = k.pubArea.RSAParameters.Sign
-		sigAlg = tpm2.AlgRSASSA
+		sigAlg = sigScheme.Alg
+		if sigAlg != tpm2.AlgRSAPSS && sigAlg != tpm2.AlgRSASSA {
+			return nil, fmt.Errorf("unsupported signing algorithm: %v", sigAlg)
+		}
```
josephlr: This seems a little strange, maybe we should do the scheme error checking inside the switch. Something like:

```go
var sigScheme *tpm2.SigScheme
switch k.pubArea.Type {
case tpm2.AlgRSA:
	sigScheme = k.pubArea.RSAParameters.Sign
	if sigScheme.Alg != tpm2.AlgRSAPSS && sigScheme.Alg != tpm2.AlgRSASSA {
		return nil, fmt.Errorf("unsupported signing algorithm: %v", sigScheme.Alg)
	}
case tpm2.AlgECC:
	sigScheme = k.pubArea.ECCParameters.Sign
	if sigScheme.Alg != tpm2.AlgECDSA {
		return nil, fmt.Errorf("unsupported signing algorithm: %v", sigScheme.Alg)
	}
default:
	return nil, fmt.Errorf("unsupported key type: %v", k.pubArea.Type)
}

hash, err := sigScheme.Hash.Hash()
if err != nil {
	return nil, err
}
return &tpmSigner{k, hash}, nil
```

This makes the error checking clearer and shorter

tpm2tools/signer.go
```diff
@@ -30,9 +30,13 @@ func (signer *tpmSigner) Public() crypto.PublicKey {
...
 // Concurrent use of Sign is thread safe, but it is not safe to access the TPM
 // from other sources while Sign is executing.
+// For RSAPSS signatures, you cannot specify custom salt lengths. The salt
+// length will be (keyBits/8) - digestSize - 2, unless that is less than the
+// digestSize in which case, saltLen will be digestSize. The only normal case
+// where saltLen is not digestSize is when using 1024 keyBits with SHA512.
 func (signer *tpmSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
 	if _, ok := opts.(*rsa.PSSOptions); ok {
-		return nil, fmt.Errorf("signing with PSS not supported")
+		return nil, fmt.Errorf("signing with non-default PSS options not supported")
```
josephlr: I think we need more here. When we want an RSA signature, I think we still pass [PSSOptions](https://golang.org/pkg/crypto/rsa/#PSSOptions). What we really want to do is just check that:

 - The key gives us a PSS signature if we get a `PSSOptions`
 - `SaltLength` in the options matches what we expect based on the key.
   - I think that means checking that it's [`PSSSaltLengthAuto`](https://golang.org/pkg/crypto/rsa/#pkg-constants)

PR #72, opened 2020-01-24
Add ECC AIK template and cache; Work around for CertifyCreation

lulumel0n:
Certain TPM (the one on my Lenovo X1 carbon) cannot handle CeritfyCreation with a null signing handle correctly. It will throw a TPM_RC_INSUFFICIENT error when we trying to do that.

One work around is that we pass in a non-null signing handle into CertifyCreation.
This PR implement this work around as well as add a ECC AIK template.

AIK ECC key will be cached for performance concerns.
I picked the handle 0x81008FFF (I picked randomly from "Storage -> Available" from Table 11 from  https://trustedcomputinggroup.org/wp-content/uploads/RegistryOfReservedTPM2HandlesAndLocalities_v1p1_pub.pdf

### DIFF
--- a/tpm2tools/handles.go
+++ b/tpm2tools/handles.go
@@ -9,10 +9,17 @@ import (
 	"github.com/google/go-tpm/tpmutil"
 )
 
-// Reserved Handles from "TCG TPM v2.0 Provisioning Guidance" - v1r1 - Table 2
 const (
-	EKReservedHandle  = tpmutil.Handle(0x81010001)
-	SRKReservedHandle = tpmutil.Handle(0x81000001)
+	// Reserved Handles from "TCG TPM v2.0 Provisioning Guidance" - v1r1 - Table 2
+	EKReservedHandle     = tpmutil.Handle(0x81010001)
+	EKECCReservedHandle  = tpmutil.Handle(0x81010002)
+	SRKReservedHandle    = tpmutil.Handle(0x81000001)
+	SRKECCReservedHandle = tpmutil.Handle(0x81000002)
+
+	// Picked available handles from TPM 2.0 Handles and Localities 2.3.1 - Table 11
+	// go-tpm-tools will use handles in the range from 0x81008F00 to 0x81008FFF
+	DefaultAIKECCHandle = tpmutil.Handle(0x81008F00)
+	DefaultAIKRSAHandle = tpmutil.Handle(0x81008F01)
 )
 
 func isHierarchy(h tpmutil.Handle) bool {
--- a/tpm2tools/keys.go
+++ b/tpm2tools/keys.go
@@ -29,7 +29,7 @@ func EndorsementKeyRSA(rw io.ReadWriter) (*Key, error) {
 
 // EndorsementKeyECC generates and loads a key from DefaultEKTemplateECC.
 func EndorsementKeyECC(rw io.ReadWriter) (*Key, error) {
-	return NewKey(rw, tpm2.HandleEndorsement, DefaultEKTemplateECC())
+	return NewCachedKey(rw, tpm2.HandleEndorsement, DefaultEKTemplateECC(), EKECCReservedHandle)
 }
 
 // StorageRootKeyRSA generates and loads a key from SRKTemplateRSA.
@@ -37,14 +37,24 @@ func StorageRootKeyRSA(rw io.ReadWriter) (*Key, error) {
 	return NewCachedKey(rw, tpm2.HandleOwner, SRKTemplateRSA(), SRKReservedHandle)
 }
 
-// AttestationIdentityKeyRSA generates and loads a key from AIKTemplateRSA
-func AttestationIdentityKeyRSA(rw io.ReadWriter, nonces []byte) (*Key, error) {
-	return NewKey(rw, tpm2.HandleOwner, AIKTemplateRSA(nonces))
-}
-
 // StorageRootKeyECC generates and loads a key from SRKTemplateECC.
 func StorageRootKeyECC(rw io.ReadWriter) (*Key, error) {
-	return NewKey(rw, tpm2.HandleOwner, SRKTemplateECC())
+	return NewCachedKey(rw, tpm2.HandleOwner, SRKTemplateECC(), SRKECCReservedHandle)
+}
+
+// AttestationIdentityKeyRSA generates and loads a key from AIKTemplateRSA
+func AttestationIdentityKeyRSA(rw io.ReadWriter, nonces []byte) (*Key, error) {
+	// If nonce is null, then try to use the cached key
+	if nonces == nil {
+		return NewCachedKey(rw, tpm2.HandleOwner, AIKTemplateRSA(nonces), DefaultAIKRSAHandle)
+	} else {
+		return NewKey(rw, tpm2.HandleOwner, AIKTemplateRSA(nonces))
+	}
+}
+
+// AttestationIdentityKeyECC generates and loads a key from AIKTemplateECC
+func AttestationIdentityKeyECC(rw io.ReadWriter) (*Key, error) {
+	return NewCachedKey(rw, tpm2.HandleOwner, AIKTemplateECC(), DefaultAIKECCHandle)
 }
 
 // EndorsementKeyFromNvIndex generates and loads an endorsement key using the
--- a/tpm2tools/keys.go
+++ b/tpm2tools/keys.go
@@ -282,9 +282,28 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
 		}
 		creationHash := sessionHashAlg.New()
 		creationHash.Write(in.GetCreationData())
-		if _, _, err = tpm2.CertifyCreation(k.rw, "", sealed, tpm2.HandleNull, nil, creationHash.Sum(nil), tpm2.SigScheme{}, ticket); err != nil {
-			return nil, fmt.Errorf("failed to certify creation: %v", err)
+
+		_, _, certErr := tpm2.CertifyCreation(k.rw, "", sealed, tpm2.HandleNull, nil, creationHash.Sum(nil), tpm2.SigScheme{}, ticket)
+		// There is a bug in some older TPMs, where they are unable to
+		// CertifyCreation when using a Null signing handle (despite this
+		// being allowed by all versions of the TPM spec). To work around
+		// this bug, we use a temporary signing key and ignore the signed
+		// result. To reduce the cost of this workaround, we use a cached
+		// ECC signing key.
+		// We can detect this bug, as it triggers a RCInsufficient
+		// Unmarshalling error.
+		if paramErr, ok := certErr.(tpm2.ParameterError); ok && paramErr.Code == tpm2.RCInsufficient {
+			signer, err := AttestationIdentityKeyECC(k.rw)
+			if err != nil {
+				return nil, fmt.Errorf("failed to create fallback signing key: %v", err)
+			}
+			defer signer.Close()
+			_, _, certErr = tpm2.CertifyCreation(k.rw, "", sealed, signer.Handle(), nil, creationHash.Sum(nil), tpm2.SigScheme{}, ticket)
 		}
+		if certErr != nil {
+			return nil, fmt.Errorf("failed to certify creation: %v", certErr)
+		}
+
 		// verify certify PCRs haven't been modified
 		decodedCreationData, err := tpm2.DecodeCreationData(in.GetCreationData())
 		if err != nil {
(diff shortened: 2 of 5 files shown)


tpm2tools/keys.go
```diff
@@ -273,7 +283,18 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
 		creationHash := sessionHashAlg.New()
 		creationHash.Write(in.GetCreationData())
 		if _, _, err = tpm2.CertifyCreation(k.rw, "", sealed, tpm2.HandleNull, nil, creationHash.Sum(nil), tpm2.SigScheme{}, ticket); err != nil {
```
josephlr: Is there a way to only trigger this fallback behavior on a specific type of error? If not that's ok. I would just like to only fallback if we hit the bug

tpm2tools/keys.go
```diff
@@ -273,7 +283,18 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
...
-			return nil, fmt.Errorf("failed to certify creation: %v", err)
+			// Fail to CertifyCreation using a null key handle. It is possible that some firmware cannot take care of the
+			// null key handle.
+			// Try to CertifyCreation again with a dummy AIK key.
+			// Using the AIK is just a workaround, we still ignore the signature and attstation returned by CertifyCreation
+			signer, err := AttestationIdentityKeyECC(k.rw)
+			if err != nil {
+				return nil, fmt.Errorf("fail to create a AIK ECC key: %v", err)
```
josephlr: "failed to create fallback signing key: %v"

tpm2tools/keys.go
```diff
@@ -273,7 +283,18 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
 		creationHash := sessionHashAlg.New()
 		creationHash.Write(in.GetCreationData())
 		if _, _, err = tpm2.CertifyCreation(k.rw, "", sealed, tpm2.HandleNull, nil, creationHash.Sum(nil), tpm2.SigScheme{}, ticket); err != nil {
-			return nil, fmt.Errorf("failed to certify creation: %v", err)
+			// Fail to CertifyCreation using a null key handle. It is possible that some firmware cannot take care of the
+			// null key handle.
+			// Try to CertifyCreation again with a dummy AIK key.
+			// Using the AIK is just a workaround, we still ignore the signature and attstation returned by CertifyCreation
```
josephlr: Let's rephrase this, something like
```
                There is a bug in some older vTPMs, where they are unable to
                CertifyCreation when using a Null signing handle (despite this
                being allowed by all versions of the TPM spec). To work around
                this bug, we use a temporary signing key and ignore the signed
                result. To reduce the cost of this workaround, we use a cached
                ECC signing key.
```

We should also wrap this comment so that it stays within 80-char lines (when using 8 space tabs)

tpm2tools/keys.go
lulumel0n: The err being returned is:

> parameter 3, error code 0x1a : the TPM was unable to unmarshal a value because there were not enough octets in the input buffer

We can try to use strings.contains() to extract the error code from the err string, but I don't think it is a good idea, since we can not get the error code directly from the err object.

tpm2tools/handles.go
```diff
@@ -9,10 +9,16 @@ import (
...
 const (
-	EKReservedHandle  = tpmutil.Handle(0x81010001)
-	SRKReservedHandle = tpmutil.Handle(0x81000001)
+	// Reserved Handles from "TCG TPM v2.0 Provisioning Guidance" - v1r1 - Table 2
+	EKReservedHandle     = tpmutil.Handle(0x81010001)
+	EKECCReservedHandle  = tpmutil.Handle(0x81010002)
+	SRKReservedHandle    = tpmutil.Handle(0x81000001)
+	SRKECCReservedHandle = tpmutil.Handle(0x81000002)
+
+	// Pickedavailable handles from TPM 2.0 Handles and Localities 2.3.1 (Table 11)
```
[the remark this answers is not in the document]
lulumel0n: Done

tpm2tools/keys.go
lulumel0n: done

tpm2tools/keys.go
lulumel0n: done

tpm2tools/keys.go
lulumel0n: done

tpm2tools/keys.go
```diff
@@ -273,7 +283,27 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
...
+			// There is a bug in some older vTPMs, where they are unable to
+			// CertifyCreation when using a Null signing handle (despite this
+			// being allowed by all versions of the TPM spec). To work around
+			// this bug, we use a temporary signing key and ignore the signed
+			// result. To reduce the cost of this workaround, we use a cached
+			// ECC signing key.
+			// The bug will cause the TPM to return a 0x1A error code. We will
+			// only try the workaround when we got the error code.
+			paramErr, ok := err.(tpm2.ParameterError)
+			if ok && paramErr.Code == 0x1A {
```
josephlr: Use `tpm2.RCInsufficient` instead of the 0x1A literal.

tpm2tools/keys.go
```diff
@@ -273,7 +283,27 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
...
-			return nil, fmt.Errorf("failed to certify creation: %v", err)
+			// There is a bug in some older vTPMs, where they are unable to
+			// CertifyCreation when using a Null signing handle (despite this
+			// being allowed by all versions of the TPM spec). To work around
+			// this bug, we use a temporary signing key and ignore the signed
+			// result. To reduce the cost of this workaround, we use a cached
+			// ECC signing key.
+			// The bug will cause the TPM to return a 0x1A error code. We will
+			// only try the workaround when we got the error code.
+			paramErr, ok := err.(tpm2.ParameterError)
```
josephlr: Combine lines and simplify like:
```go
if paramErr, ok := err.(tpm2.ParameterError); ok && paramErr.Code == tpm2.RCInsufficient {
        // fallback
}
```

tpm2tools/keys.go
```diff
@@ -273,7 +283,27 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
...
 		if _, _, err = tpm2.CertifyCreation(k.rw, "", sealed, tpm2.HandleNull, nil, creationHash.Sum(nil), tpm2.SigScheme{}, ticket); err != nil {
-			return nil, fmt.Errorf("failed to certify creation: %v", err)
+			// There is a bug in some older vTPMs, where they are unable to
+			// CertifyCreation when using a Null signing handle (despite this
+			// being allowed by all versions of the TPM spec). To work around
+			// this bug, we use a temporary signing key and ignore the signed
+			// result. To reduce the cost of this workaround, we use a cached
+			// ECC signing key.
+			// The bug will cause the TPM to return a 0x1A error code. We will
+			// only try the workaround when we got the error code.
```
josephlr: We want to avoid comments that just describe the code. Just say:
```go
// We can detect this bug, as it triggers a RCInsufficient Unmarshalling error.
```

tpm2tools/keys.go
```diff
@@ -273,7 +283,27 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
...
+				if err != nil {
+					return nil, fmt.Errorf("failed to create fallback signing key: %v", err)
+				}
+				defer signer.Close()
+				if _, _, err = tpm2.CertifyCreation(k.rw, "", sealed, signer.Handle(), nil, creationHash.Sum(nil), tpm2.SigScheme{}, ticket); err != nil {
+					return nil, fmt.Errorf("failed to certify creation after the fallback: %v", err)
+				}
+			} else {
+				return nil, fmt.Errorf("failed to certify creation: %v", err)
+			}
```
josephlr: This is a little hard to read, it might be easier if we are better about naming our errors, consider something like:

```go
_, _, certErr := tpm2.CertifyCreation(/* use nil signer */)

// Comment block explaining this fallback
if paramErr, ok := certErr.(tpm2.ParameterError); ok && paramErr.Code == tpm2.RCInsufficient {
	signer, err := AttestationIdentityKeyECC(k.rw)
	if err != nil {
		return nil, fmt.Errorf("failed to create fallback signing key: %v", err)
	}
	defer signer.Close()
	_, _, certErr = tpm2.CertifyCreation(/* use signer */)
}

if certErr != nil {
	return nil, fmt.Errorf("failed to certify creation: %v", certErr)
}
```

This also makes the error paths much easier to read.

tpm2tools/keys.go
lulumel0n: Done

tpm2tools/keys.go
lulumel0n: Done

tpm2tools/keys.go
lulumel0n: Done

tpm2tools/keys.go
lulumel0n: Done

tpm2tools/keys.go
```diff
@@ -272,9 +282,28 @@ func (k *Key) Unseal(in *tpmpb.SealedBytes, cOpt CertifyOpt) ([]byte, error) {
...
+		_, _, certErr := tpm2.CertifyCreation(k.rw, "", sealed, tpm2.HandleNull, nil, creationHash.Sum(nil), tpm2.SigScheme{}, ticket)
+		// There is a bug in some older TPMs, where they are unable to
+		// CertifyCreation when using a Null signing handle (despite this
+		// being allowed by all versions of the TPM spec). To work around
+		// this bug, we use a temporary signing key and ignore the signed
+		// result. To reduce the cost of this workaround, we use a cached
+		// ECC signing key.
+		// We can detect this bug, as it triggers a RCInsufficient
+		// Unmarshalling error.
+		if paramErr, ok := err.(tpm2.ParameterError); ok && paramErr.Code == tpm2.RCInsufficient {
```
josephlr: The check should be `certErr.(tpm2.ParameterError)`

lulumel0n: Done

savely-krasovsky:
> I'm not familiar with how admin rights affecting TPM on windows, does it work with admin rights?

Basically Windows doesn't allow you to get ECC SRK without admin rights for some reason. But after getting it for the first time (for example by spawning one-shot Windows service which just calling `client.StorageRootKeyECC()`), it will be available for other users without requiring admin rights, weird.

I didn't try with admin rights with production-ready piece of software because it does not meant to run under admin.

> Also maybe your TPM doesn't support the default AIKTemplateECC parameter (CurveNISTP256)?

Probably, but it successfully passing those lines:
```
signer, err := AttestationKeyECC(k.rw)
if err != nil {
	return nil, fmt.Errorf("failed to create fallback signing key: %w", err)
}
```
Otherwise it would return error here, isn't it?

> Or maybe you can try to use the RSA SRK to do the workaround.

We are already using RSA SRK and for the majority of users it works flawlessly, sealing, unsealing, etc. But during testing getting error at unsealing stage with this Lenovo Yoga.

Does it worth a try to change this line from: `signer, err := AttestationKeyECC(k.rw)` to `signer, err := AttestationKeyRSA(k.rw)`?

jkl73:
> > I'm not familiar with how admin rights affecting TPM on windows, does it work with admin rights?
> 
> Basically Windows doesn't allow you to get ECC SRK without admin rights for some reason. But after getting it for the first time (for example by spawning one-shot Windows service which just calling `client.StorageRootKeyECC()`), it will be available for other users without requiring admin rights, weird.
> 
> I didn't try with admin rights with production-ready piece of software because it does not meant to run under admin.
> 
> > Also maybe your TPM doesn't support the default AIKTemplateECC parameter (CurveNISTP256)?
> 
> Probably, but it successfully passing those lines:
> 
> ```
> signer, err := AttestationKeyECC(k.rw)
> if err != nil {
> 	return nil, fmt.Errorf("failed to create fallback signing key: %w", err)
> }
> ```
> 
> Otherwise it would return error here, isn't it?
>
> > Or maybe you can try to use the RSA SRK to do the workaround.
> 
> We are already using RSA SRK and for the majority of users it works flawlessly, sealing, unsealing, etc. But during testing getting error at unsealing stage with this Lenovo Yoga.
> 
> Does it worth a try to change this line from: `signer, err := AttestationKeyECC(k.rw)` to `signer, err := AttestationKeyRSA(k.rw)`?

Yes, I think it's worth to try, at least it should reveal if the problem is related to the ecc  algo or not.

PR #73, opened 2020-02-27
tpm2tools: Consolidate session handling

josephlr:
This PR cleans up session handling to make different types of key
authorization easier to use. Refreshing a key's authorization is
now part of each key operation.

We can now sign with keys that need authorization and can Import under
parent keys that don't need authorization.

Tests/comments are updated accordingly.

### DIFF
--- a/tpm2tools/keys.go
+++ b/tpm2tools/keys.go
@@ -2,6 +2,7 @@
 package tpm2tools
 
 import (
+	"bytes"
 	"crypto"
 	"crypto/subtle"
 	"fmt"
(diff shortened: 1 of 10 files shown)


tpm2tools/session.go
```diff
@@ -0,0 +1,92 @@
...
+	// - don't bind the session to any particular key
+	session, _, err := tpm2.StartAuthSession(
+		rw,
+		/*tpmkey=*/ tpm2.HandleNull,
+		/*bindkey=*/ tpm2.HandleNull,
+		/*nonceCaller=*/ make([]byte, sessionHashAlg.Size()),
+		/*encryptedSalt=*/ nil,
+		/*sessionType=*/ tpm2.SessionPolicy,
+		/*symmetric=*/ tpm2.AlgNull,
+		/*authHash=*/ sessionHashAlgTpm)
```
BenDHillier: Here and the later call to StartAuthSession (line 62-65) should have the parameter name comments on the same side

tpm2tools/keys.go
```diff
@@ -12,14 +13,16 @@ import (
 	"github.com/google/go-tpm/tpmutil"
 )
 
-// Key wraps an active TPM2 key. Users of Key should be sure to call Close()
-// when finished using the Key, so that the underlying TPM handle can be freed.
+// Key wraps an active asymmetric TPM2 key. This can either be a signing key or
```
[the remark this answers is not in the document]
josephlr: Yes, we never had a way to create symetric keys

tpm2tools/session.go
josephlr: Fixed, and moved into a dedicated helper function.

PR #74, opened 2020-02-27
Add read nvdata cmd

lulumel0n:
Add a gotpm cmd for tpm2.NVReadEX()
password will be empty and block size will be 0 (which will default to TPM_PT_NV_BUFFER_MAX)

### DIFF
--- a/cmd/read.go
+++ b/cmd/read.go
@@ -5,6 +5,8 @@ import (
 
 	pb "github.com/golang/protobuf/proto"
 	"github.com/google/go-tpm-tools/tpm2tools"
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
 	"github.com/spf13/cobra"
 )
 
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
 	},
 }
 
+var nvReadCmd = &cobra.Command{
+	Use:   "nvdata",
+	Short: "Read TPM NVData",
+	Long: `Read NVData at a particular NVIndex
+
+Based on the --index flag, this reads all of the NVData present at that NVIndex.
+The read is authenticated with the owner hierarchy and an empty password.`,
+	Args: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		rwc, err := openTpm()
+		if err != nil {
+			return err
+		}
+		defer rwc.Close()
+
+		data, err := tpm2.NVReadEx(rwc, tpmutil.Handle(nvIndex), tpm2.HandleOwner, "", 0)
+		if err != nil {
+			return err
+		}
+		if _, err := dataOutput().Write(data); err != nil {
+			return fmt.Errorf("cannot output NVData: %v", err)
+		}
+		return nil
+	},
+}
+
 func init() {
 	RootCmd.AddCommand(readCmd)
 	readCmd.AddCommand(pcrCmd)
+	readCmd.AddCommand(nvReadCmd)
 	addOutputFlag(pcrCmd)
 	addPCRsFlag(pcrCmd)
 	addHashAlgoFlag(pcrCmd)
+	addIndexFlag(nvReadCmd)
+	nvReadCmd.MarkPersistentFlagRequired("index")
+	addOutputFlag(nvReadCmd)
 }
--- a/tpm2tools/keys.go
+++ b/tpm2tools/keys.go
@@ -68,7 +68,7 @@ func EndorsementKeyFromNvIndex(rw io.ReadWriter, idx uint32) (*Key, error) {
 // (possibly a hierarchy root tpm2.Handle{Owner|Endorsement|Platform|Null})
 // using the template stored at the provided nvdata index.
 func KeyFromNvIndex(rw io.ReadWriter, parent tpmutil.Handle, idx uint32) (*Key, error) {
-	data, err := tpm2.NVRead(rw, tpmutil.Handle(idx))
+	data, err := tpm2.NVReadEx(rw, tpmutil.Handle(idx), parent, "", 0)
 	if err != nil {
 		return nil, fmt.Errorf("read error at index %d: %v", idx, err)
 	}


cmd/read.go
```diff
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
 	},
 }
 
+var nvReadCmd = &cobra.Command{
+	Use:   "nvdata <endorsement | owner | platform | null>",
```
josephlr: For now, let's not bother with heierarchies for this command. We should just use the owner heierarchy and be done with it.

The only other heierarchy that can be used here is  the platfrom heierachy, and it generally wouldn't work with the empty password anyway. (as platform NV Indicies generally used enhanced auth).

cmd/read.go
```diff
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
...
+	Short: "Read NVData from the TPM",
```
josephlr: Change to
```
Read TPM NVData
```

cmd/read.go
```diff
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
...
+	Long: `Read NVData from the TPM
+
+Based on the given hierarchy, read the contents of the TPM's NVData.
+An empty password will be passed in, and the size will be the default
+to TPM_PT_NV_BUFFER_MAX`,
```
josephlr: Change to:
```
Read NVData at a particular NVIndex

Based on the --index flag, this reads all of the NVData present at that NVIndex. The read is authenticated with the owner hierarchy and an empty password.

File: cmd/read.go [hunk 4]
```diff
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
...
+	ValidArgs: getKeys(hierarchyNames),
```
josephlr: We can delete this

cmd/read.go
```diff
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
...
+	Args:      cobra.ExactValidArgs(1),
```
josephlr: `Args: cobra.NoArgs,`

cmd/read.go
```diff
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		rwc, err := openTpm()
+		if err != nil {
+			return err
+		}
+		defer rwc.Close()
+
+		data, err := tpm2.NVReadEx(rwc, tpmutil.Handle(nvIndex), hierarchyNames[args[0]], "", 0)
```
josephlr: Just pass `tpm2.HandleOwner` directly here.

josephlr: We should also change the existing `tpm2.NVRead` in `tpm2tools` to use `tpm2.NVReadEx`

cmd/read.go
```diff
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
...
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(dataOutput(), string(data))
```
josephlr: We should just do:

```go
if _, err := dataOutput().Write(data); err != nil {
        return fmt.Errorf("outputting NVData: %v", err)
}
```

like we do in `seal`.

cmd/read.go
```diff
@@ -52,10 +54,40 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
...
+}
+
 func init() {
 	RootCmd.AddCommand(readCmd)
 	readCmd.AddCommand(pcrCmd)
+	readCmd.AddCommand(nvReadCmd)
 	addOutputFlag(pcrCmd)
 	addPCRsFlag(pcrCmd)
 	addHashAlgoFlag(pcrCmd)
+	addIndexFlag(nvReadCmd)
```
josephlr: We need to make the index flag required here.

cmd/flags.go
```diff
@@ -56,6 +57,13 @@ var algos = map[tpm2.Algorithm]string{
 	tpm2.AlgSHA512: "sha512",
 }
 
+var hierarchyNames = map[string]tpmutil.Handle{
+	"endorsement": tpm2.HandleEndorsement,
+	"owner":       tpm2.HandleOwner,
+	"platform":    tpm2.HandlePlatform,
+	"null":        tpm2.HandleNull,
+}
+
```
josephlr: Move this back to where it was (now that we don't need it here).

cmd/flags.go
```diff
@@ -92,6 +100,14 @@ func (f *algoFlag) Allowed() string {
...
 }
 
+func getKeys(m map[string]tpmutil.Handle) []string {
+	keys := make([]string, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
```
josephlr: We also no longer need this helper function.

We actually do this "get the keys from a map" operation twice in `gotpm`. Unfortunately, it's with a `map[string]tpmutil.Handle` and a `map[string][]tpm2.HandleType`. As Go lacks generics, we can use a helper function (we would need two helper functions.

cmd/read.go
lulumel0n: done

cmd/read.go
lulumel0n: done

cmd/read.go
lulumel0n: done

cmd/read.go
lulumel0n: done

cmd/read.go
lulumel0n: done

cmd/read.go
lulumel0n: done

cmd/read.go
lulumel0n: done

cmd/flags.go
lulumel0n: done

cmd/flags.go
lulumel0n: done

cmd/read.go
lulumel0n: done:
```
data, err := tpm2.NVReadEx(rw, tpmutil.Handle(idx), parent, "", 0)
```

cmd/read.go
lulumel0n: done

PR #75, opened 2020-02-28
internal: Only open the physical TPM once

josephlr:
This prevents some flakes we've seen on Linux

### DIFF
--- a/internal/test_tpm.go
+++ b/internal/test_tpm.go
@@ -1,0 +1,46 @@
+package internal
+
+import (
+	"io"
+	"sync"
+	"testing"
+
+	"github.com/google/go-tpm-tools/simulator"
+)
+
+// Only open the TPM device once. Reopening the device causes issues on Linux.
+var (
+	tpm  io.ReadWriteCloser
+	lock sync.Mutex
+)
+
+type NoClose struct {
+	io.ReadWriter
+}
+
+func (n NoClose) Close() error {
+	return nil
+}
+
+// GetTPM is a cross-platform testing helper function that retrives the
+// appropriate TPM device from the flags passed into "go test".
+func GetTPM(tb testing.TB) io.ReadWriteCloser {
+	tb.Helper()
+	if useRealTPM() {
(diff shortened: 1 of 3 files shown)


internal/test_tpm.go
```diff
@@ -0,0 +1,46 @@
...
+// appropriate TPM device from the flags passed into "go test".
+func GetTPM(tb testing.TB) io.ReadWriteCloser {
+	tb.Helper()
+	if useRealTPM() {
+		lock.Lock()
+		defer lock.Unlock()
+		if tpm == nil {
+			var err error
+			if tpm, err = getRealTPM(); err != nil {
+				tb.Fatalf("Failed to open TPM")
```
lulumel0n: We should expose the useful error msg here:
```
tb.Fatalf("Failed to open TPM: %v", err)
```

internal/test_tpm.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (n NoClose) Close() error {
+	return nil
+}
+
+// GetTPM is a cross-platform testing helper function that retrives the
+// appropriate TPM device from the flags passed into "go test".
+func GetTPM(tb testing.TB) io.ReadWriteCloser {
+	tb.Helper()
+	if useRealTPM() {
```
lulumel0n: The thing is we don't distinguish the /dev/tpm0 and /dev/tpmrm0 here. Seems like currently it is fine, but it could be an issue if we mix use tpm0, tpmrm0 in future tests.

internal/test_tpm.go
josephlr: Fixed

internal/test_tpm.go
josephlr: So this will be ok even if `/dev/tpmrm0` is used. When testing with a real TPM, the `--tpm-path` flag _must_ be used (otherwise a simulator is used). And this flag value doesn't change throughout the program. This means that returning the same TPM for every call will give correct behavior.

PR #76, opened 2020-02-28
Import: Remove io.ReadWriter parameter

josephlr:
This is a breaking change, so we will need to release a 0.2 version
after this.

### DIFF
--- a/server/import_test.go
+++ b/server/import_test.go
@@ -43,7 +43,7 @@ func TestImport(t *testing.T) {
 				t.Fatalf("creating import blob failed: %v", err)
 			}
 
-			output, err := ek.Import(rwc, blob)
+			output, err := ek.Import(blob)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


josephlr:
This should be merged after #77 and #75

josephlr:
So it looks like we've already had a breaking change since `v0.1.2`. When @BenDHillier did #46, we changed:
```go
func CreateImportBlob(ekPub crypto.PublicKey, sensitive []byte) (*tpmpb.ImportBlob, error)
```
to 
```go
func CreateImportBlob(ekPub crypto.PublicKey, sensitive []byte, pcrs *tpmpb.Pcrs) (*tpmpb.ImportBlob, error) 
```
which is backwards-incompatible.

So unless we want to undo that (which I don't), we should just release `v0.2.0` _after_ we have make all the backwards-incompatible changes we want to.

PR #93, opened 2020-12-06
signer: Prevent nil dereference w/ PSSOptions

josephlr:
Attempting to use PSSOptions with a non-RSA key results in a segfault
when attempting to access the RSAParameters. Fix this by adding
additional checks. Add a test for this case.

### DIFF
--- a/tpm2tools/signer_test.go
+++ b/tpm2tools/signer_test.go
@@ -206,3 +206,40 @@ func TestSignPSS(t *testing.T) {
 		})
 	}
 }
+
+/// Make sure signing fails when using PSS params with a non-PSS key
+func TestFailSignPSS(t *testing.T) {
+	rwc := internal.GetTPM(t)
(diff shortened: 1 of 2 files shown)


tpm2tools/signer_test.go
```diff
@@ -206,3 +206,40 @@ func TestSignPSS(t *testing.T) {
...
+/// Make sure signing fails when using PSS params with a non-PSS key
+func TestFailSignPSS(t *testing.T) {
+	rwc := internal.GetTPM(t)
+	defer CheckedClose(t, rwc)
+	tests := []struct {
+		name     string
+		template tpm2.Public
+	}{
+		{"SSA", templateSSA(tpm2.AlgSHA256)},
+		// {"PSS", templatePSS(tpm2.AlgSHA256)},
```
lulumel0n, line 219: Remove the commented out

lulumel0n (approved):

PR #100, opened 2021-04-28
Add event log fetching and parsing

alexmwu:
Here we add support for grabbing the Linux TCG event log and use the work https://github.com/google/go-attestation does to parse the event log.
This includes a unit test against a cached event log and  a system test, supporting a real event log and TPM.

Tested:
`sudo ../server.test  -test.run TestSystemParseEventLog -tpm-path /dev/tpmrm0 -event-log-path /sys/kernel/security/tpm0/binary_bios_measurements`
on a physical machine and a Shielded VM.


alexmwu:
Rebased on the tpm2tools->client package name change

server/test/debian_10_event_log.yml
```diff
@@ -0,0 +1,164 @@
+// Fetched via `tpm2_eventlog debian_10_binary_bios_measurements`
```
josephlr, line 1: Let's just remove this for now, as we do not use it in our tests.

server/test/debian_10_event_log.yml
```diff
@@ -0,0 +1,164 @@
...
+// Note this is invalid yaml.
+---
+events:
```
josephlr, line 4: We should use output from a Newer GCP VM to at least get the EV_NONHOST_INFO stuff.

alexmwu: Done.

server/test/debian_10_event_log.yml
alexmwu: Done.

josephlr:
@alexmwu 

Rebased on latest master, and slightly change the dependancy importing commit, will review the rest tomorrow.

It also seems like there's a test import cycle, can you fix that?

alexmwu:
> @alexmwu
> 
> Rebased on latest master, and slightly change the dependancy importing commit, will review the rest tomorrow.
> 
> It also seems like there's a test import cycle, can you fix that?

weird, I thought I pushed that change. fixed it now

josephlr:
@googlebot I consent.

josephlr:
@alexmwu I was able to simplify the circular dependency issue by moving the client tests to a `client_test` crate (while keeping everything in the same place) see https://github.com/google/go-tpm-tools/pull/100/commits/9a6cfa4fe6a11ba073cb71558d77b3f3badce770.

This required some other changes (see the commits before that one):
  - Making the `tpmpb.Pcrs` functions methods instead
  - Making some values public

I also shortened and cleaned up some of the eventlog_test code (mostly just moving stuff inline)

Also, do you have the corresponding PCRs for the RHEL tests?

josephlr (approved):

PR #105, opened 2021-06-18
Add client.ReadAllPCRs and use it in "gotpm read pcr"

josephlr:
This change also includes some other minor cleanups for the command line tool:
  - Formatting of PCRs is improved (to look more like what `tpm2-tools` does)
  - Now `gotpm read pcr` and `gotpm seal` can still both use `--hash-algo`, but have different default values
  - We now use `/dev/tpmrm0` by default.

### DIFF
--- a/cmd/read.go
+++ b/cmd/read.go
@@ -16,6 +16,8 @@ var readCmd = &cobra.Command{
 	Args:  cobra.NoArgs,
 }
 
+var pcrHashAlgo = tpm2.AlgUnknown
+
 var pcrCmd = &cobra.Command{
 	Use:   "pcr",
 	Short: "Read PCRs from the TPM",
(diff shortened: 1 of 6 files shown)


cmd/read.go
```diff
@@ -35,22 +35,29 @@ If --pcrs is not provided, all pcrs are read for that hash algorithm.`,
...
+			if err != nil {
+				return err
+			}
+			return pcrList.PrettyFormat(dataOutput())
 		}
 
-		fmt.Fprintf(debugOutput(), "Reading pcrs (%v)\n", sel.PCRs)
-		pcrList, err := client.ReadPCRs(rwc, sel)
+		fmt.Fprintln(debugOutput(), "Reading all PCRs")
+		banks, err := client.ReadAllPCRs(rwc)
```
jkl73, line 53: Here we are ignoring the user input --pcrs flag and print out all pcrs. We should print the --pcrs flag selection

josephlr: I added a check requiring `--hash-algo` to be specified if `--pcrs` is.

josephlr: Note this is a breaking change but that's fine.

jkl73 (approved):

PR #106, opened 2021-06-18
Add client.Attest

josephlr:
This change adds an `Attest` method to `Key`, as well as some smaller organizational changes around `Quote.Verify`. See the individual commits for more information.

Things that are still needed (will be in other PRs):
  - Server side verification (#103)
  - Updated documentation (will do after all code changes have landed)
  - Updated tests (will mostly be in #103, might add more tests in a later PR)


josephlr:
@iKevinY PTAL, this should be all the code needed client-side

iKevinY:
> @iKevinY PTAL, this should be all the code needed client-side

Changes LGTM overall (will let others do the formal code review).

client/pcr.go
```diff
@@ -33,6 +33,26 @@ func min(a, b int) int {
 	return b
 }
 
+// Get a list of selections corresponding to the TPM's implemented PCRs
+func implementedPCRs(rw io.ReadWriter) ([]tpm2.PCRSelection, error) {
+	caps, moreData, err := tpm2.GetCapability(rw, tpm2.CapabilityPCRs, math.MaxUint32, 0)
+	if err != nil {
+		return nil, fmt.Errorf("listing implemented PCR banks: %w", err)
+	}
+	if moreData {
```
jkl73, line 42: Why moreData results in error here? Is this a temporary solution?

client/keys.go
```diff
@@ -393,6 +393,32 @@ func (k *Key) Quote(selpcr tpm2.PCRSelection, extraData []byte) (*tpmpb.Quote, e
...
+		return nil, err
+	}
+
+	attestation := tpmpb.Attestation{}
+	if attestation.AkPub, err = k.PublicArea().Encode(); err != nil {
+		return nil, fmt.Errorf("failed to encode public area: %w", err)
+	}
+	attestation.Quotes = make([]*tpmpb.Quote, len(sels))
+	for i, sel := range sels {
+		if attestation.Quotes[i], err = k.Quote(sel, nonce); err != nil {
```
jkl73, line 412: What's the reason to quote each pcr separately? Can we just quote the sels one time?

jkl73 (commented):
Are we adding tests for Attest() later?

client/keys.go
josephlr: This is a limitation of [`QuoteRaw`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#QuoteRaw), it can only take a single [`PCRSelection`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#PCRSelection) despite the underlying command supporting a list of selections.

Having separate quotes also helps with quote verification (so one PCR set getting messed up doesn't ruin the whole thing).

client/pcr.go
josephlr: `moreData` should never be set for this type of call (per the spec), so erroring out here is fine. I'll add a comment explaining.

josephlr:
> Are we adding tests for Attest() later?

Yes, there will be two types of tests:
  - Simple `Attest()` tests, will be added once I make some other testing improvements
  - Integration against the Server-side verify (will be added in #103)

client/pcr.go
jkl73: I see, then we should also make sure the go-tpm library https://github.com/google/go-tpm/blob/7b1a3bb951f3f6af1a0a5d0190278480e522d752/tpm2/tpm2.go#L288 won't return that

jkl73 (approved):

PR #107, opened 2021-06-19
Improve testing (and add tests for client.Attest)

josephlr:
This PR implements multiple testing fixes:
  - Only uses PCRs 16 and 23 (to avoid breaking real systems)
  - Change `client.GetEventLog` to take the TPM as an argument (allowing the TPM to override the implementation)
  - Add tests for `client.Attest`

### DIFF
--- a/client/eventlog.go
+++ b/client/eventlog.go
@@ -1,0 +1,19 @@
+package client
+
+import "io"
+
+// GetEventLog grabs the crypto-agile TCG event log for the system. The TPM can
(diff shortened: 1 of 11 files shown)


josephlr:
@jkl73 I confirmed that running the tests against a real TPM only changed PCR 16, and kept the TPM in a usable state.

jkl73 (approved):
LGTM. This also reminds we probably should reset those extended PCRs in the test since the tests could use a real TPM.

PR #116, opened 2021-06-25
Add GceInstanceInfo structure to Attestation

josephlr:
Also add comments explaining what the attestation fields do.

Also add an `InstanceURL()` method to make it easier to call [`getShieldedInstanceIdentity`](https://cloud.google.com/compute/docs/reference/rest/v1/instances/getShieldedInstanceIdentity)


proto/tpm.proto
```diff
@@ -47,8 +50,24 @@ message Pcrs {
   map<uint32, bytes> pcrs = 2;
 }
 
+// Information uniquely identifying a GCE instnace. Can be used to create an
```
iKevinY, line 53: s/instnace/instance/ (also in PR description).

iKevinY (approved):

proto/tpm.proto
josephlr: Bahhhhh, thank you

iKevinY: Oh also s/Attestaion/Attestation in the PR title (in case that gets used as the squashed commit message lol).

josephlr: Lol, should be fixed now

PR #117, opened 2021-06-25
Add client.AttestOpts

josephlr:
This will allow us to add functionality to `client.Atttest` in a backwards compatible way.

Note that like many similar Go interfaces (e.g. [`PublicKey`](https://golang.org/pkg/crypto/#PublicKey), [`PrivateKey`](https://golang.org/pkg/crypto/#PrivateKey), [`DecrypterOpts`](https://golang.org/pkg/crypto/#DecrypterOpts), etc...) this `AttestOpts` interface has no methods. This just means we have to use casts to switch functionality, rather than using common methods. 

Also, rename `SealOpt` to `SealOpts` and `CertifyOpt` to `CertifyOpts`. This is the more consistent naming scheme.

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"crypto"
 	"crypto/subtle"
+	"errors"
 	"fmt"
 	"io"
 
@@ -407,11 +408,19 @@ func (k *Key) Quote(selpcr tpm2.PCRSelection, extraData []byte) (*tpmpb.Quote, e
 	return &quote, nil
 }
 
+// AttestOpts allows for optional Attest functionality to be enabled.
+type AttestOpts interface{}
+
 // Attest generates an Attestation containing the TCG Event Log and a Quote over
 // all PCR banks. The provided nonce can be used to guarantee freshness of the
 // attestation. This function will return an error if the key is not a
 // restricted signing key.
-func (k *Key) Attest(nonce []byte) (*tpmpb.Attestation, error) {
+//
+// An optional AttestOpts can also be passed. Currently, this parameter must be nil.
+func (k *Key) Attest(nonce []byte, opts AttestOpts) (*tpmpb.Attestation, error) {
+	if opts != nil {
+		return nil, errors.New("provided AttestOpts must be nil")
+	}
 	sels, err := implementedPCRs(k.rw)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


jkl73 (approved):

iKevinY (approved):
One pattern I've noticed in my (limited) experience with Go is methods taking in a `opts ...FooOpt` as the last parameter to allow for a variadic number of option objects to be passed in. Does that makes sense here in the context of each of the XOpts in `go-tpm-tools` rather than pluralizing all of the option structs?

If the current approach is preferable, then LGTM.

josephlr:
> One pattern I've noticed in my (limited) experience with Go is methods taking in a `opts ...FooOpt` as the last parameter to allow for a variadic number of option objects to be passed in. Does that makes sense here in the context of each of the XOpts in `go-tpm-tools` rather than pluralizing all of the option structs?

Ya I've seen this too, usually under the name "Functional Options" (see [this Blog post by Rob Pike](https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html)). Basically, each `opt` object in the variadic opts can configure the server, allowing you to feed in many options (or none). This is a good pattern when you have a bunch of orthogonal options and it makes sense to allow them to be specified independently.

However, this isn't the best pattern if:
  - you only want one or zero options
  - specifying multiple options would conflict

This is the case for `SealOpts` and `CertifyOpts` (e.g. [`CertifyCurrent`](https://pkg.go.dev/github.com/google/go-tpm-tools@v0.3.0-alpha6/client#CertifyCurrent) and [`CertifyExpected`](https://pkg.go.dev/github.com/google/go-tpm-tools@v0.3.0-alpha6/client#CertifyExpected) would conflict). The Go `crypto` package has [`DecrypterOpts`](https://golang.org/pkg/crypto/#DecrypterOpts) which works similarly. There is also the issue with `client.Reseal` which takes both sealing and certifying parameters, so it might be hard to implement that using variadic functions. 

Our eventual `VerifyOpts` will also work similarly, a user will only specify one way to get/check the `PublicKey` (`nil` won't be allowed in that case). I think the case for `AttestOpts` is less clear; we currently only have two use cases in mind:
  - `nil`: Don't do anything extra
  - `GetGCEMetadata{ *metadata.Client }`: Use the `metadata.Client` to retrieve the additional GCE metadata.

So I think using `Opts` instead of `Opt` is the right move for now (for consistency). But this is a _really good point_ to bring up. We should see how the code evolves as we add more functionality. It may turn out we like the variadic approach better and will switch to it with an eventual `v0.4` or `v1.0`.

iKevinY:
Makes sense; I wasn't sure how each of the options would interact with one another or potentially expand in the future (on the point of orthogonality), so thanks for clarifying that!

PR #118, opened 2021-06-25
Regenerate tpm.pb.go

josephlr:
Also, improve the documentation and add a `go:generate` comment. Now we can regenerate the protobuf code just by running:
```bash
go generate ./...
```

This also changes `SealedBytes.pcrs` to be an `uint32` instead of a `int32`. This is binary compatible and makes the use of PCRs more consistent.


proto/doc.go
```diff
@@ -1,3 +1,5 @@
 // Package proto contains protocol buffers that are exchanged between the client
 // and server. Note, some of these types have additional helper methods.
 package proto
+
+//go:generate protoc --go_out=. --go_opt=module=github.com/google/go-tpm-tools/proto tpm.proto
```
alexmwu, line 25: Not important now: do we want to keep it *.proto, in case we add protos outside of the tpm.proto file in the future?

alexmwu (approved):

proto/doc.go
josephlr: I was initially going to do this, but `go:generate` doesn't expand wildcards

PR #119, opened 2021-06-25
Make simulator double close a no-op, as a temp fix

jkl73:
Double closing the simulator results in mutex being unlocked twice which causes panic. Temporarily make this double closing as a no-op until we fixed the downstream.

### DIFF
--- a/simulator/simulator.go
+++ b/simulator/simulator.go
@@ -19,6 +19,7 @@ package simulator
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"math/rand"
 	"sync"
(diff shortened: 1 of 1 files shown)


simulator/simulator.go
```diff
@@ -106,6 +106,10 @@ func (s *Simulator) Read(responseBuffer []byte) (int, error) {
 // Close cleans up and stops the simulator, Close() should always be called when
 // the Simulator is no longer needed, freeing up other callers to use Get().
 func (s *Simulator) Close() error {
+	if s.closed {
+		fmt.Println("The Simulator has already been closed!")
```
josephlr, line 110: Nit: Can we print to stderr:
```go
os.Stderr.WriteString("The Simulator has already been closed!")
```

josephlr (commented):
Could we also add asserts that no simulator method is being called once its closed for the first time. This would still cause a panic, but only if a Simulator was _used_ after being closed (which is just not allowed).

I think the following methods need to have this check:
 - `Reset`
 - `ManufactureReset`
 - `Write`
 - `Read`

josephlr:
@googlebot I consent.

jkl73:
lgtm

josephlr (approved):

PR #123, opened 2021-06-29
cmd: Change default tpm path to use correct defaults

josephlr:
This makes sure `gotpm` does the right thing on systems that don't
support the RM.


jkl73 (commented):
LGTM, but why can't we have the default tpmpath logic here (instead of go-tpm)? I feel like that's a somewhat very specific use case in the cmd tool.

josephlr:
> LGTM, but why can't we have the default tpmpath logic here (instead of go-tpm)? I feel like that's a somewhat very specific use case in the cmd tool.

This is a good point. I updated the PR to not depend on https://github.com/google/go-tpm/pull/256. Instead it just implements the logic in this package. This should make things easier, and it still solves the problem.

I also updated our dependencies and ran `go mod tidy`.

jkl73 (approved):

iKevinY (approved):

PR #124, opened 2021-06-30
Move all helper code out of proto package, and split proto files

### LINKED ISSUE #114: Should we rename proto to tpmpb?
Having packages end in `pb` seems to be the standard convention.

Examples:
  - The example in the official docs: https://developers.google.com/protocol-buffers/docs/gotutorial#defining-your-protocol-format
  - All the subpackages of: https://github.com/protocolbuffers/protobuf-go/tree/master/types

Most notably, it conflicts with https://pkg.go.dev/google.golang.org/protobuf@v1.26.0/proto also called `proto`

josephlr:
This change does the following:

- Splits the `proto` package (containing `proto/tpm.proto`) into:
  - `proto/tpm`: Containing `proto/tpm.proto`
  - `proto/attest`: Containing `proto/attest.proto`
  - The `GCEInstanceInfo` and `Attestation` messages are moved to `proto.attest.proto`
- Moves all non-generated code out of `proto` (needed for Bazel compatibility), performing the following renames: 
  - `proto.Pcrs.PrettyFormat` -> `internal.FormatPCRs`
  - `proto.Pcrs.CheckIfSubsetOf` -> `internal.CheckSubset`
  - `proto.Pcrs.PCRSelection` -> `internal.PCRSelection`
  - `proto.Pcrs.HasSamePCRSelection` -> `internal.SamePCRSelection`
  - `proto.Pcrs.ComputePCRSessionAuth` -> `internal.PCRSessionAuth`
  - `proto.Pcrs.ComputePCRDigest` -> `internal.PCRDigest`
  - `proto.Quote.Verify` -> `internal.VerifyQuote`
  - `proto.GceInstanceInfo.InstanceURL` -> `server.GCEInstanceURL`
  - Note: `proto/doc.go` will only be built externally (as the package only contains stuff for godoc)
- When possible, imports protobuf packages as `pb`
  - This is made easier by moving the attestation related code to its own files
- Internal testing code is moved to `internal/test`
  - This required renaming some local variables
- Renames the following messages to be consistent with Go Naming conventions:
  - `tpm.Pcrs` -> `tpm.PCRs`
  - `attest.GceInstanceInfo` -> `attest.GCEInstanceInfo`

Fixes #114


iKevinY (approved):
Thanks for refactoring on such short notice, @josephlr! I'm assuming Cirrus CI covers all the building and unit testing, so LGTM since there aren't any internal breakages after the refactor. :shipit:

josephlr:
> I'm assuming Cirrus CI covers all the building and unit testing

Yes, cirrus checks: building (for Linux and Windows), testing (just Linux), generating (i.e. running protoc), and linting

iKevinY (commented):
One question: is it intentional that the `.pb.go` files now live in their own subdirectories (ie. `proto/attest/attest.pb.go` and `proto/tpm/tpm.pb.go`)? Previously it seems like `foo.pb.go` and `foo.proto` lived next to each other, so I just want to make sure that this is desired.

jkl73 (approved):

josephlr:
> One question: is it intentional that the `.pb.go` files now live in their own subdirectories (ie. `proto/attest/attest.pb.go` and `proto/tpm/tpm.pb.go`)? Previously it seems like `foo.pb.go` and `foo.proto` lived next to each other, so I just want to make sure that this is desired.

Yes this is expected and intentional. Basically, it allows us to import thing similarly with Bazel and with the Go build system.

With go build:
  - `import pb "github.com/google/go-tpm-tools/proto/tpm"`
  - Before we were just importing `github.com/google/go-tpm-tools/proto`, but now we need to disabiguate.

With Bazel:
  - We will just have one `BUILD` file in the proto directory
  - The subdirectories will not be used
  - We will have `proto_library` rules `tpm_proto`/`attest_proto`
  - We will have `go_proto_library` rules `tpm_go_proto`/`attest_go_proto`
  - `import pb "google3/foo/bar/baz/tpm_tools/proto/tpm_go_proto"`

iKevinY:
> Yes this is expected and intentional. Basically, it allows us to import thing similarly with Bazel and with the Go build system.

Awesome, thanks for the clarification! 😀

PR #125, opened 2021-07-01
Add proto generation script

josephlr:
This avoids having to deal with the limitations `go generate`. We can now use globs and will always run from the repository root.

We also change the proto import paths to be repository-relative, making Bazel support easier.


iKevinY (approved):
It seems like the generated code has changed somewhat non-trivially (see the raw descriptor byte arrays for the protos); assuming this isn't going to break anything, then LGTM.

PR #127, opened 2021-07-09
Fix some tests failing on a physical TPM

jkl73:
Fix some tests might fail when using a real TPM or simulator with different config.

Some cases in `TestReadPCRs`, `TestSign` will be skipped if the hash algorithm is not supported.
Also fixed TestReadPCRs. Previously it relys on the test PCR to be all zeros at beginning of the test (true for the simulator, but not for a real TPM)

Remove the VerifyPSS with the specified salt length, as it behaves differently between TPM implementations.


internal/test/test_tpm.go
```diff
@@ -40,6 +41,26 @@ func (s simulatedTpm) EventLog() ([]byte, error) {
 	return s.eventLog, nil
 }
 
+// SkipOnAlgorithm skips the test if the algorithm is not found in the TPM
+// capability.
+func SkipOnAlgorithm(tb testing.TB, tpm io.ReadWriteCloser, alg tpm2.Algorithm) {
+	tb.Helper()
+
+	caps, moreData, err := tpm2.GetCapability(tpm, tpm2.CapabilityAlgs, math.MaxUint32, 0)
```
alexmwu, line 49: While I suspect `math.MaxUint32` as propertyCount works in theory, the spec says the following (part 3, 30.2.1):
```
When all of the properties of the requested type have been returned, the moreData parameter in the
response will be set to NO. Otherwise, it will be set to YES.
NOTE 3 The moreData parameter will be YES if there are more properties even if the requested number of
capabilities has been returned.
The TPM is not required to return more than one value at a time. It is not required to provide the same
number of values in response to subsequent requests.
EXAMPLE 2 A TPM may return 4 properties in response to a TPM2_GetCapability(capability =
TPM_CAP_TPM_PROPERTY, property = TPM_PT_MANUFACTURER, propertyCount = 8 ) and for a
latter request with the same parameters, the TPM may return as few as one and as many as 8
values.
```

So, a TPM may in fact only return 1 property with moreData set to true and we would get the Fatal `received extra data from GetCapability`

Unless we have a good reason for not, we should use the moreData field to collect all properties.

internal/test/test_tpm.go
```diff
@@ -40,6 +41,26 @@ func (s simulatedTpm) EventLog() ([]byte, error) {
 	return s.eventLog, nil
 }
 
+// SkipOnAlgorithm skips the test if the algorithm is not found in the TPM
+// capability.
```
alexmwu, line 45: If we think we'll want to use go-tpm's skip function to reduce duplication (which we would expose as a test package), we should add a TODO.

internal/test/test_tpm.go
```diff
@@ -40,6 +41,26 @@ func (s simulatedTpm) EventLog() ([]byte, error) {
...
+func SkipOnAlgorithm(tb testing.TB, tpm io.ReadWriteCloser, alg tpm2.Algorithm) {
```
alexmwu, line 46: I like go-tpm's name `skipOnUnsupportedAlgorithm`, as it makes clear that unsupported algorithms on a TPM will not run. Thoughts?

client/signer_test.go
```diff
@@ -237,13 +252,9 @@ func TestSignPSS(t *testing.T) {
...
-			// Verify with expected salt length.
-			err = rsa.VerifyPSS(signer.Public().(*rsa.PublicKey), k.opts.HashFunc(), digest[:], sig, &rsa.PSSOptions{SaltLength: k.saltLen, Hash: k.opts.HashFunc()})
-			if err != nil {
-				t.Error(err)
-			}
 
-			// Verify with default salt length.
+			// Different implementations may specify different salt length. Some have "keyBytes - digestSize - 2", some have
+			// just "digestSize". Therefore here we just verify with default salt length.
 			err = rsa.VerifyPSS(signer.Public().(*rsa.PublicKey), k.opts.HashFunc(), digest[:], sig, nil)
```
alexmwu, line 258: What's the reason we don't use the hash function? `&rsa.PSSOptions{Hash: k.opts.HashFunc()}` or `&rsa.PSSOptions{SaltLength: PSSSaltLengthAuto, Hash: k.opts.HashFunc()}` to be explicit. This allows us to check against the expected HashFunc for the test.

alexmwu:
@jkl73 could you also add the PR message into the commit? it's easier to git blame then

internal/test/test_tpm.go
jkl73: Yes, I will use the implementation in GoTPM's skipOnUnsupportedAlgorithm which handles moreData.

client/signer_test.go
jkl73: Seems like the `Hash` in `rsa.PSSOptions` is not used in `rsa.VerifyPSS()`. And passing PSSSaltLengthAuto (0) has the same effect as passing the nil PSSOptions ptr. So I fell they are testing the same there, what do you think?

internal/test/test_tpm.go
jkl73: Yes, we may also need to extend its functionality like supporting skipping over a ECC curve (CapabilityECCCurves). I will create an issue for that

internal/test/test_tpm.go
jkl73: changed

jkl73:
> @jkl73 could you also add the PR message into the commit? it's easier to git blame then

Added

client/signer_test.go
alexmwu: make sense to me. let's keep it as is

alexmwu (approved):

internal/test/test_tpm.go
alexmwu: makes sense. please resolve when you do.

PR #128, opened 2021-07-13
Add attest command to gotpm CLI

alexmwu:
The Attest command supports generating a proto containing the quote, AK, and TCG event log using
client.Attest. The user can pass in which key type (GCE or go-tpm-tools generated AK) as well as the
key algo and a nonce to generate the attestation.


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/google/go-tpm-tools
 go 1.16
 
 require (
+	github.com/golang/protobuf v1.5.0
```
jkl73, line 6: This shouldn't be needed, run `go mod tidy` to clean

cmd/attest.go
```diff
@@ -0,0 +1,93 @@
...
+	Args: cobra.ExactValidArgs(1),
+	ValidArgs: func() []string {
+		// The keys from the hierarchyNames map are our valid arguments
+		aks := make([]string, len(attestationKeys))
+		for ak := range attestationKeys {
+			aks = append(aks, ak)
+		}
+		return aks
+	}(),
+	RunE: func(cmd *cobra.Command, args []string) error {
```
jkl73, line 55: args is not used, so the keyType will always default to "default", is this intended?

cmd/open.go
```diff
@@ -22,7 +25,11 @@ func (ic ignoreClose) Close() error {
 
 func openTpm() (io.ReadWriteCloser, error) {
 	if ExternalTPM != nil {
-		return ignoreClose{ExternalTPM}, nil
+		ignoreRwc := ignoreClose{ExternalTPM, nil}
```
jkl73, line 28: I feel like the logic should be in test.GetTPM(). In a test the ExternalTPM will always != nil because ExternalTPM will be assigned by test.GetTPM(). But I am not 100% sure.

cmd/attest.go
alexmwu: nope. oversight on my part. thanks!

cmd/open.go
jkl73: ExternalTPM could mean a real TPM when setting the test flag for a using a real TPM for test `--tpm-path='/dev/tpmrm0'` So we can't add EventLogGetter to ignoreClose, otherwise the test against the real TPM will fail.
I think we need to refactor all the open TPM related code. We can do after this pr if we want want to leave the real TPM attest test failing here.
My first though is to separate GetEventLog from the TPM itself. Because it is not directly depend on the TPM

cmd/attest_test.go
```diff
@@ -0,0 +1,107 @@
...
+				if err := internal.VerifyQuote(quote, matchingKeyPub.PublicKey(), nonce); err != nil {
+					t.Errorf("failed to verify quote: %v", err)
+				}
+			}
+
+			// Replay Event Log.
+			replaySuccess := false
+			for _, quote := range attestation.GetQuotes() {
+				if _, err := server.ParseAndVerifyEventLog(attestation.GetEventLog(), quote.GetPcrs()); err == nil {
+					replaySuccess = true
```
jkl73, line 82: Also just noticed.. the replay will succeed as long as one set of PCRs is verified with EventLog? If this is intended we can just break after we found one replaySuccess here.

alexmwu: good point, thanks!

cmd/attest_test.go
```diff
@@ -0,0 +1,107 @@
...
+				}
+			}
+			if !replaySuccess {
+				t.Errorf("failed to replay event log: %v", err)
```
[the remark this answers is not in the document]
alexmwu: will remove.

cmd/open.go
alexmwu: I agree that the current state is confusing for simulator vs real TPM. We want to tie the event log and TPM together for a simulated TPM because we want to extend the event log events in the simulator.

I suspect the best approach if we want to get rid of the weirdness is to refactor our testing to have a `test.GetTPMAndEventLog` replace `test.GetTPM`. This can return some `TPMAndEventLog` structure that tests can optionally take one or both of. We then need to revisit `client.Attest` and probably introduce an `eventLog []byte` parameter again that calls `GetEventLog` when nil. I'm also open to other suggestions as this would make `Attest` less usable to favor testing.
Thoughts, @jkl73 and @josephlr?

Also, I tested this on a real TPM. What failure are you getting when running the command on a real TPM?

jkl73: I'm running `sudo go test --tpm-path="/dev/tpmrm0"` under the cmd folder. And it is a seg fault because of the "Real" ExternalTPM.  
```
--- FAIL: TestAttest (0.99s)
    --- FAIL: TestAttest/AttestRSADefaultArg (0.99s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x7dd0ef]
...
	/usr/local/google/home/jiankun/githubprojects/go-tpm-tools/client/eventlog.go:9 +0x62
github.com/google/go-tpm-tools/client.(*Key).Attest(0xc0001a0090, 0xc000128f60, 0x4, 0x8, 0x0, 0x0, 0x2, 0x1f, 0x0)
	/usr/local/google/home/jiankun/githubprojects/go-tpm-tools/client/attest.go:39 +0x328
github.com/google/go-tpm-tools/cmd.glob..func2(0xd6c5c0, 0xc00016e380, 0x1, 0x8, 0x0, 0x0)
	/usr/local/google/home/jiankun/githubprojects/go-tpm-tools/cmd/attest.go:90 +0x3f5
...
exit status 2
FAIL	github.com/google/go-tpm-tools/cmd	1.047s

```

jkl73: I think passing an `eventlog` parameter to `client.Attest` makes sense

alexmwu: created an issue to discuss more: https://github.com/google/go-tpm-tools/issues/129. Also, I fixed the issue you mentioned. PTAL.

alexmwu:
Closing since we merged https://github.com/google/go-tpm-tools/pull/293, which has more testing and review

PR #131, opened 2021-08-19
Add policy with secure boot

alexmwu:
This PR includes the branched extra policy bits from #103 minus the complicated SecureBootPolicy.

Users with an attestation server may want to apply a policy against a validated event log. A common use case might be validating secure configuration against expected values. For example, a server may want to check that a machine has updated its dbx and is not vulnerable to Boothole.

The policy object includes a firmware and secure boot related subpolicies. The firmware policy is mostly GCE instance related.


alexmwu:
Superseded by #139

PR #133, opened 2021-08-31
Change SealOpts and UnsealOpts to a struct instead of interface

alexmwu:
SealOpts was previously an interface that implemented PCRsForSealing.
This did not allow users the option to seal to both a current PCR
selection and a target PCR selection without first reading PCRs manually
and merging into the target. 

CertifyOpts was previously an interface that implemented CertifyPCRs.
This did not allow users the option to seal to both a current PCR
selection and an expected PCR selection without manually reading PCRs
and merging into the target.

These structs allows these use cases more
easily and support future options.

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -229,9 +229,9 @@ func (k *Key) Seal(sensitive []byte, opts SealOpts) (*pb.SealedBytes, error) {
 	var err error
 	var auth []byte
 
-	pcrs, err = sealOptsToPcrs(k.rw, opts)
+	pcrs, err = mergePCRSelAndProto(k.rw, opts.Current, opts.Target)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid SealOpts: %v", err)
 	}
 	if len(pcrs.GetPcrs()) > 0 {
 		auth = internal.PCRSessionAuth(pcrs, SessionHashAlg)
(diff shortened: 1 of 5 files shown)


josephlr (commented):
@alexmwu this is the correct model, feel free to add `UnsealOpts` to this PR as well (removing all the `Cerify*` structs/interface).

jkl73 (approved):
LGTM

client/pcr.go
```diff
@@ -99,64 +98,22 @@ func ReadAllPCRs(rw io.ReadWriter) ([]*pb.PCRs, error) {
...
-// CertifyPCRs will compare the digest with given expected PCRs values.
-func (p CertifyExpected) CertifyPCRs(_ io.ReadWriter, pcrs *pb.PCRs) error {
-	if len(p.Pcrs.GetPcrs()) == 0 {
-		panic("CertifyExpected contains 0 PCRs")
-	}
-	return internal.CheckSubset(p.Pcrs, pcrs)
+type UnsealOpts struct {
+	// CertifyCurrent certifies that a selection of current PCRs have the same
+	// value when sealing.
+	// Hash Algorithm in the selection should be CertifyHashAlgTpm.
```
josephlr, line 112: Can we also update the docs on `CertifyHashAlgTpm` to be:
```go
/// CertifyHashAlgTpm is the hard-coded algorithm that must be used with UnsealOpts.
```

client/seal_test.go
```diff
@@ -329,10 +331,130 @@ func BenchmarkSeal(b *testing.B) {
...
+				t.Errorf("received PCRs (%v) do not match expected PCRs (%v)",
+					outPcrsMap, testcase.expectedPcrs)
+			}
+		})
+	}
+
+	// Run empty SealOpts.
+	_, err = srk.Seal([]byte("secretzz"),
+		client.SealOpts{})
+	if err != nil {
```
josephlr, line 408: nit: make one line

```go
if _, err = srk.Seal([]byte("secretzz"), client.SealOpts{}); err != nil {
	
}
```

client/seal_test.go
```diff
@@ -329,10 +331,130 @@ func BenchmarkSeal(b *testing.B) {
...
+		ret := make(map[uint32]struct{})
+		for _, val := range a {
+			ret[val] = struct{}{}
+		}
+		return ret
+	}
+	for _, testcase := range opts {
+		t.Run(testcase.name, func(t *testing.T) {
+			sealed, err := srk.Seal([]byte("secretzz"),
+				client.SealOpts{Current: testcase.current, Target: testcase.target})
```
josephlr, line 393: Here and elsewhere in this file, this might be more readable as:
```go
opts := client.SealOpts{Current: testcase.current, Target: testcase.target}
sealed, err := srk.Seal([]byte("secretzz"), opts)
```

client/keys.go
```diff
@@ -313,7 +349,11 @@ func (k *Key) Unseal(in *pb.SealedBytes, opts CertifyOpts) ([]byte, error) {
 	}
 	defer tpm2.FlushContext(k.rw, sealed)
 
-	if opts != nil {
+	pcrs, err := mergePCRSelAndProto(k.rw, opts.CertifyCurrent, opts.CertifyExpected)
+	if err != nil {
+		return nil, fmt.Errorf("invalid UnsealOpts: %v", err)
+	}
+	if len(pcrs.GetPcrs()) > 0 {
```
josephlr, line 356: Should we move the `CheckSubset` to after this line. That way we fail early on bad `UnsealOpts`. For example, if a user uses a hash algorithm in `UnsealOpts` that's not `CertifyHashAlgTpm`, we want to get the error
```
failed to certify PCRs: PCR hash algo not matching: ...
```
sooner rather than even bothering with the process of certification.

josephlr (approved):
Looks great! Only nits

cmd/seal.go
```diff
@@ -117,9 +117,9 @@ machine state when sealing took place.
 		fmt.Fprintln(debugOutput(), "Unsealing data")
 
 		certifySel := tpm2.PCRSelection{Hash: client.CertifyHashAlgTpm, PCRs: pcrs}
-		var opts client.CertifyOpts
+		var opts client.UnsealOpts
 		if len(certifySel.PCRs) > 0 {
-			opts = client.CertifyCurrent{PCRSelection: certifySel}
+			opts = client.UnsealOpts{CertifyCurrent: certifySel}
 		}
```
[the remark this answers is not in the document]
alexmwu: I also removed the opts parameter, as that's unneeded now as well.

cmd/seal.go
```diff
@@ -51,7 +51,7 @@ state (like Secure Boot).`,
 		fmt.Fprintf(debugOutput(), "Sealing to PCRs: %v\n", sel.PCRs)
 		var opts client.SealOpts
 		if len(sel.PCRs) > 0 {
-			opts = client.SealCurrent{PCRSelection: sel}
+			opts = client.SealOpts{Current: sel}
 		}
```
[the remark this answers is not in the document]
alexmwu: Done.

client/keys.go
```diff
@@ -250,6 +250,42 @@ func (k *Key) Seal(sensitive []byte, opts SealOpts) (*pb.SealedBytes, error) {
 	return sb, nil
 }
 
+func mergePCRSelAndProto(rw io.ReadWriter, sel tpm2.PCRSelection, proto *pb.PCRs) (*pb.PCRs, error) {
```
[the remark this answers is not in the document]
alexmwu: yup, makes sense!

client/pcr.go
alexmwu: done.

client/seal_test.go
alexmwu: done.

client/seal_test.go
alexmwu: oh looks like you prefer this to  the inline creation method. I have no preference so I'll do this and revert the changes I mentioned in cmd/seal.go above.

client/keys.go
alexmwu: this makes sense to me! nice call.

PR #144, opened 2021-11-18
Add Kernel cmdline parser

alexmwu:
- Add LinuxKernelState to MachineState.
- Parse Linux kernel command line from the GRUB state (only works with GRUB at the moment).


server/kernel_cmdline.go
```diff
@@ -0,0 +1,130 @@
...
+		// use equals to differentiate.
+		if equals == 0 {
+			param = param[:len(param)-1]
+		} else {
+			val = val[:len(val)-1]
+		}
+	}
+
+	if args[i] != 0 {
+		args[i] = 0
```
jkl73, line 95: I saw this in the kernel C code as well, but I don't quite understand why need to assign 0 to args[i]

server/eventlog.go
```diff
@@ -182,3 +234,130 @@ func convertToPbEvents(hash crypto.Hash, events []attest.Event) []*pb.Event {
...
+	return nil, fmt.Errorf("supported dm-verity arg not found")
+}
+
+func getLinuxKernelStateFromGRUB(grub *pb.GrubState, parseVerity bool) (*pb.LinuxKernelState, error) {
+	var cmdline string
+	var err error
+	var verity *pb.DmVerityState
+	seen := false
+
+	for _, command := range grub.GetCommands() {
```
jkl73, line 340: What will happen if no kernel cmdline is found? Should throw an error?

server/eventlog.go
```diff
@@ -182,3 +234,130 @@ func convertToPbEvents(hash crypto.Hash, events []attest.Event) []*pb.Event {
...
+				hasher.Write(rawData[suffixAt:])
+				if !bytes.Equal(event.Digest, hasher.Sum(nil)) {
+					return nil, fmt.Errorf("invalid digest seen for GRUB event log in event %d: %s", idx, hex.EncodeToString(event.Digest))
+				}
+			}
+			hasher.Reset()
+			commands = append(commands, string(rawData))
+		}
+	}
+	return &pb.GrubState{Files: files, Commands: commands}, nil
```
jkl73, line 281: What if the files or commands is nil/empty, should we throw an error?

server/eventlog.go
```diff
@@ -41,10 +76,27 @@ func ParseMachineState(rawEventLog []byte, pcrs *tpmpb.PCRs) (*pb.MachineState,
...
 
+	var grub *pb.GrubState
+	var kernel *pb.LinuxKernelState
+	if opts.Loader == GRUB {
+		grub, err = getGrubState(cryptoHash, rawEvents)
+		if err != nil {
+			// TODO(wuale): replace with SecureBoot changes for GroupedError
+			return &pb.MachineState{}, err
+		}
+		kernel, err = getLinuxKernelStateFromGRUB(grub, opts.ParseVerity)
```
jkl73, line 87: Instead of parsing both kernel cmdline and dmverity stuff in the same function, can they be done in two separate functions? I fell like it can make things more clear.

server/eventlog.go
```diff
@@ -41,10 +76,27 @@ func ParseMachineState(rawEventLog []byte, pcrs *tpmpb.PCRs) (*pb.MachineState,
...
+		// TODO(wuale): replace with changes for GroupedError
+		if err != nil {
+			return &pb.MachineState{}, err
```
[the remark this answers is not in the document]
alexmwu: This is temporary (e.g., it should get overwritten with changes in #139), but more importantly we want to return a valid MachineState with GroupedError down the line. This allows downstream consumers to have some things parsed out from the event log.

server/kernel_cmdline.go
alexmwu: Idk whether you're asking why the kernel does it or why we do it, so I'll explain both:
The kernel sets the null terminator at the point i because that byte represents the next space found in args. As a result, the val/param return values in the kernel can be properly null-terminated but still point to the same memory as in args initially (no more mem allocs). 

For golang: we try and modify args as closely as the kernel does. Even if golang doesn't have C-style/null-terminated strings, we interpret the byte array the same way, since we also have logic that deals with if any byte in arg == 0.

server/eventlog.go
alexmwu: like returning the parsed command line from its own function and then parsing dm-verity? That makes sense to me, will do.

server/kernel_cmdline.go
```diff
@@ -0,0 +1,130 @@
...
+		return true
+	}
+	return false
+}
+
+func skipSpaces(str []byte) []byte {
+	i := 0
+	for i < len(str) && isspace(str[i]) {
+		i++
+	}
```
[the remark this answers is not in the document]
jkl73: ah yes..

server/eventlog.go
jkl73: yes
```
kernel.commandline, err = getKernelCMDFromGRUB...
kernel.verity, err = getDmVerityFromGRUB...
```

server/eventlog.go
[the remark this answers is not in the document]
jkl73: Gotcha

server/eventlog.go
alexmwu: I don't know what issues there could be with GRUB potentially used with non-Linux systems or some Linux setup with an auto or empty command line. I also didn't want to add a "ParseKernelCmdline" option. That said, we can document that we only support certain common configurations (GRUB booting the kernel) and error here.

server/eventlog.go
alexmwu: Yes, this is an oversight on my part. Thanks!

server/kernel_cmdline.go
jkl73: I see arg[i] should be a space before, so it can be set to 0.

server/kernel_cmdline.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// parseArgs takes a Linux kernel command line and returns a key value mapping
+// for each command line argument.
+func parseArgs(commandline []byte) map[string]string {
+	var args []byte
+	// nextArgs must receive a null-terminated string.
+	if len(commandline) == 0 || commandline[len(commandline)-1] != 0 {
+		args = make([]byte, len(commandline)+1)
+		args[len(args)-1] = 0
```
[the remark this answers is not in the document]
alexmwu: Yeah we could do that. Actually looking closer, technically line 114 is unnecessary regardless because the last value is the zero-value (byte 0x00) anyway.

server/eventlog.go
alexmwu: Done in #143.

server/kernel_cmdline.go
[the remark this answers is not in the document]
alexmwu: Pulled out in https://github.com/google/go-tpm-tools/pull/217

server/eventlog_test.go
```diff
@@ -633,18 +669,71 @@ func generateNonCosCelEvent(hashAlgoList []crypto.Hash) (cel.Record, error) {
...
+				if msState.LinuxKernel == nil || len(msState.LinuxKernel.CommandLine) == 0 {
+					t.Errorf("expected %s to have a LinuxKernelState", log.name)
+				}
+
+				if msState.LinuxKernel.CommandLine != log.expectedCmdline {
+					got := len(msState.LinuxKernel.CommandLine)
+					want := len(log.expectedCmdline)
+					if got != want {
+						t.Log(got)
+						t.Log(want)
```
jkl73, line 701: This only log a number, could make the message more specific/detail

server/eventlog_test.go
```diff
@@ -633,18 +669,71 @@ func generateNonCosCelEvent(hashAlgoList []crypto.Hash) (cel.Record, error) {
...
+					for i := 0; i < got; i++ {
+						gotAt := msState.LinuxKernel.CommandLine[i]
+						wantAt := log.expectedCmdline[i]
+						if gotAt != wantAt {
+							t.Log(i)
+							t.Logf("%c", gotAt)
+							t.Logf("%c", wantAt)
+							t.Log(gotAt)
+							t.Log(wantAt)
+						}
```
jkl73, line 712: IIUC, why need to print the commandline char by char? Also if `len(log.expectedCmdline) < len(msState.LinuxKernel.CommandLine)` line 705 can have  a out of bound err

jkl73 (approved):
Just comments about the test

server/eventlog_test.go
alexmwu: These are left over debugging lines; thanks for catching. Removed

server/eventlog_test.go
alexmwu: Debug printlines. Removed

PR #145, opened 2021-12-03
cel: Some ideas for ergonomics improvements

josephlr:
Some ideas about how we might make CEL stuff easier to work with.

@alexmwu @jkl73 what do you think? Worth doing?

### DIFF
--- a/cel/canonical_eventlog.go
+++ b/cel/canonical_eventlog.go
@@ -35,7 +35,7 @@ type TLV struct {
 }
 
 // MarshalBinary marshals a TLV to a byte slice.
-func (tlv *TLV) MarshalBinary() (data []byte, err error) {
+func (tlv TLV) MarshalBinary() (data []byte, err error) {
 	buf := make([]byte, len(tlv.Value)+tlvTypeFieldLength+tlvLengthFieldLength)
 
 	buf[0] = tlv.Type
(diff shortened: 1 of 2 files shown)


cel/canonical_eventlog.go
```diff
@@ -35,7 +35,7 @@ type TLV struct {
 }
 
 // MarshalBinary marshals a TLV to a byte slice.
-func (tlv *TLV) MarshalBinary() (data []byte, err error) {
+func (tlv TLV) MarshalBinary() (data []byte, err error) {
```
alexmwu, line 38: typically they are pointer receivers: https://cs.opensource.google/search?q=MarshalBinary()&ss=go%2Fgo:src%2F&start=1

alexmwu (approved):
Makes a lot of sense to me. Thanks!

jkl73 (approved):
LGTM, this way should be easier to use Record.

cel/canonical_eventlog.go
josephlr: So it seems like:
  - the digest types all use pointer receivers
  - `time.Time` [uses a value receiver](https://cs.opensource.google/go/go/+/master:src/time/time.go;l=1171;drc=90462dfc3aa99649de90bb587af56a9cb0214665)
  - `ip.Addr`/`ip.AddrPort`/`ip.Prefix` [use a value receiver](https://cs.opensource.google/go/go/+/master:src/net/netip/netip.go;l=993;drc=e83a2047e0332e45c20fb1bcdb984e9cc74ffb1f)
  - Some of the `gob` adapters use pointer receivers
  - [`json.RawMessage`](https://pkg.go.dev/encoding/json@go1.17.3#RawMessage) uses value receivers for marshaling and pointer receivers for unmarshaling.

So I think it's fine to use a value receiver here. The type is small, marshaling doesn't change the value, and if I change it to a pointer receiver the code doesn't compile.

This is because you then can't do something like:
```go
createRecNumField(r.RecNum).MarshalBinary()
```
as temporary values do not have an address.

PR #146, opened 2021-12-07
Add cos_tlv and ability to parse CEL to MachineState

jkl73:
Add more COS events ops related to CEL

Now able to append an CEL event given the type and its content, and parse the event after decoding.

Add digest verification for a CEL record when doing replay and decode.


cel/cos_tlv.go
```diff
@@ -1,31 +1,51 @@
...
-	// CosTypeValue indicate the CELR event is a COS content
+	// CosEventType indicate the CELR event is a COS content
 	// TODO: the value needs to be reserved in the CEL spec
-	CosTypeValue uint8 = 80
+	CosEventType uint8 = 80
+)
+
+// Type for COS nested events
+const (
+	ImageRefType      uint8 = iota
```
alexmwu, line 16: you can introduce a 
`type CosType uint8`
and then have `ImageRefType  CosType = iota`

cel/cos_tlv.go
```diff
@@ -1,31 +1,51 @@
...
+	ImageDigestType   uint8 = iota
```
alexmwu, line 17: nit: here and below, no need for `= iota`

cel/cos_tlv.go
```diff
@@ -37,3 +57,17 @@ func (c CosTlv) GenerateDigest(hashAlgo crypto.Hash) ([]byte, error) {
 	}
 	return hash.Sum(nil), nil
 }
+
+// NewCosTlvFromTlv constructs a CosTlv from a TLV. It will check for the correct COS event
+// type, and unmarshal the nested event.
+func NewCosTlvFromTlv(tlv TLV) (CosTlv, error) {
+	if tlv.Type != CosEventType {
+		return CosTlv{}, fmt.Errorf("not a COS event")
```
alexmwu, line 65: would be helpful to see the seen type in the error: `return CosTlv{}, fmt.Errorf("TLV type %v not a COS event", tlv.Type)`

cel/cos_tlv.go
```diff
@@ -37,3 +57,17 @@ func (c CosTlv) GenerateDigest(hashAlgo crypto.Hash) ([]byte, error) {
 	}
 	return hash.Sum(nil), nil
 }
+
+// NewCosTlvFromTlv constructs a CosTlv from a TLV. It will check for the correct COS event
+// type, and unmarshal the nested event.
+func NewCosTlvFromTlv(tlv TLV) (CosTlv, error) {
```
alexmwu, line 63: why do we need this? And if we do, should we export it?

cel/canonical_eventlog_test.go
```diff
@@ -11,51 +11,66 @@ func TestCELEncodingDecoding(t *testing.T) {
...
+			t.Errorf("SHA256 digest length doesn't match for record %d", i)
+		}
+		if len(decodedcel.Records[i].Digests[crypto.SHA256]) != 32 {
+			t.Errorf("SHA256 digest length doesn't match for record %d", i)
+		}
+		if len(decodedcel.Records[i].Digests[crypto.SHA1]) != 20 {
+			t.Errorf("SHA1 digest length doesn't match for record %d", i)
+		}
+
+		extractedCos, err := NewCosTlvFromTlv(decodedcel.Records[i].Content)
```
alexmwu, line 62: seems like we're testing both CEL encode decode with COS functionality in this test. let's break it up into another test in a `cos_tlv_test.go` file

cel/cos_tlv.go
jkl73: done

cel/cos_tlv.go
jkl73: done

cel/cos_tlv.go
jkl73: done

cel/cos_tlv.go
jkl73: I think yes, we need a way to parse a regular TLV (record.content) into a CosTlv, after then we can read its content

alexmwu: Oh I see, thanks! I got confused reading this, since I was thinking higher level: something that goes through a CEL and per supported event type tried to convert to a content specific value (`CEL`, `PCCLIENT_STD`, `COS_TYPE`, etc.). Regardless, we need this.

Let's turn this into a receiver function on TLV.

jkl73: Yep, changing to ParseToCosTlv()

server/eventlog_test.go
```diff
@@ -238,6 +241,8 @@ var Debian10GCE = eventLog{
 			5: decodeHex("019079179dbc0eb5992c500dcf8a095910ac590d"),
 			6: decodeHex("b2a83b0ebf2f8374299a5b2bdfc31ea955ad7236"),
 			7: decodeHex("9e6c57e850f371c2a7fe02bca552149363952318"),
+			// PCR 13 here is to hold CEL COS event digests for a specified test
+			13: decodeHex("5d2017fdcdbf56ac66b333e526b9169315a897d3"),
```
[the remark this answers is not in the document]
alexmwu: with the new changes, you should be able to test separately even easier. with the stuff below.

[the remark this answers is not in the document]
jkl73: Done

server/eventlog_test.go
```diff
@@ -350,6 +355,56 @@ func TestParseSecureBootState(t *testing.T) {
...
+
+	var buf bytes.Buffer
+	err := coscel.EncodeCEL(&buf)
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+
+	// test on Debian 10 only, preferred to switch to COS 93 later if possible
+	for _, bank := range Debian10GCE.Banks {
+		if msState, err := ParseMachineState(test.Debian10EventLog, buf.Bytes(), bank); err != nil {
```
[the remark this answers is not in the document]
jkl73: Thanks! this is more clear

server/eventlog.go
```diff
@@ -26,7 +27,7 @@ import (
 // It is the caller's responsibility to ensure that the passed PCR values can be
 // trusted. Users can establish trust in PCR values by either calling
 // client.ReadPCRs() themselves or by verifying the values via a PCR quote.
-func ParseMachineState(rawEventLog []byte, pcrs *tpmpb.PCRs) (*pb.MachineState, error) {
+func ParseMachineState(rawEventLog []byte, rawCanonicalEventLog []byte, pcrs *tpmpb.PCRs) (*pb.MachineState, error) {
```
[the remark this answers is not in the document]
jkl73: Done

server/eventlog.go
```diff
@@ -62,6 +69,45 @@ func contains(set [][]byte, value []byte) bool {
...
+
+	cosState := &pb.AttestedCosState{}
+	cosState.Container = &pb.ContainerState{}
+
+	for _, record := range coscel.Records {
+		cosTlv, err := cel.NewCosTlvFromTlv(record.Content)
+		// igonore non COS CEL events
+		if err != nil {
+			continue
+		}
```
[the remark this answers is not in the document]
jkl73: good point, I created a function IsCosTlv() to do the check.

cel/canonical_eventlog_test.go
```diff
@@ -124,6 +113,29 @@ func TestCELReplayFailTampered(t *testing.T) {
...
+	defer client.CheckedClose(t, tpm)
+
+	cel := &CEL{}
+	measuredHashes := []crypto.Hash{crypto.SHA256, crypto.SHA1, crypto.SHA512}
+
+	cosEvent := CosTlv{ImageRefType, []byte("docker.io/bazel/experimental/test:latest")}
+	cosEvent2 := CosTlv{ImageDigestType, []byte("sha256:781d8dfdd92118436bd914442c8339e653b83f6bf3c1a7a98efcfb7c4fed7483")}
+	appendOrFatal(t, cel, tpm, test.ApplicationPCR, measuredHashes, cosEvent)
+	appendOrFatal(t, cel, tpm, test.ApplicationPCR, measuredHashes, cosEvent2)
+	modifieddRecord := CosTlv{ImageDigestType, []byte("sha256:000000000000000000000000000000000000000000000000000000000000000")}
```
alexmwu, line 127: nit: `modifiedRecord`

client/example_test.go
```diff
@@ -137,7 +137,7 @@ func ExampleKey_Attest() {
 	// On verifier, replay event log.
 	// TODO: decide which hash algorithm to use in the quotes. SHA1 is
 	// typically undesirable but is the only event log option on some distros.
-	_, err = server.ParseMachineState(attestation.EventLog, attestation.Quotes[0].Pcrs)
+	_, err = server.VerifyAttestation(attestation, server.VerifyOpts{Nonce: nonce})
```
alexmwu, line 140: Since `VerifyAttestation` calls `VerifyQuote`, the code should probably be rewritten as follows:

```
	// TODO: establish trust in the AK (typically via an AK certificate signed
	// by the manufacturer).
	// On verifier, verify the Attestation message. This:
        // - checks the quote(s) against a stored public key/AK
	// certificate's public part and the expected nonce.
        // - replays the event log against the quoted PCRs
        // - extracts events into a MachineState message.
	// TODO: decide which hash algorithm to use in the quotes. SHA1 is
	// typically undesirable but is the only event log option on some distros.
	_, err = server.VerifyAttestation(attestation, server.VerifyOpts{Nonce: nonce, TrustedAKs: []crypto.PublicKey{ak.GetPublicKey()}})
```

and then remove lines 126 to 135 (calling internal.VerifyQuote and its comments).

go.mod
```diff
@@ -6,6 +6,7 @@ go 1.16
 require (
 	github.com/google/certificate-transparency-go v1.1.1
 	github.com/google/go-attestation v0.3.2
+	github.com/google/go-cmp v0.5.5
```
alexmwu, line 9: any reason we add this dependency over just using `reflect.DeepEqual`? All the fields are exported in CosTlv and we don't override equality checks.

server/verify.go
```diff
@@ -77,14 +78,22 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
...
 			continue
 		}
 
+		celState, err := parseCanonicalEventLog(attestation.CanonicalEventLog, pcrs)
+		if err != nil {
+			lastErr = fmt.Errorf("failed to validate the Canonical event log: %w", err)
+			continue
+		}
+
+		proto.Merge(state, celState)
```
alexmwu, line 95: nit: we discussed having the CEL be the source because if we parse anything from a CEL that's generally safer. That said, this is not too important now (but if we decide to change in the future that could potentially be a breaking change in certain circumstances).

server/eventlog.go
```diff
@@ -62,6 +74,49 @@ func contains(set [][]byte, value []byte) bool {
...
+	// Validate the COS event log first.
+	if err := coscel.Replay(pcrs); err != nil {
+		return nil, err
+	}
+
+	cosState := &pb.AttestedCosState{}
+	cosState.Container = &pb.ContainerState{}
+
+	for _, record := range coscel.Records {
+		// igonore non COS CEL events
```
alexmwu, line 92: nit: "Ignore" and period at end.

server/eventlog_test.go
```diff
@@ -347,7 +350,66 @@ func TestParseSecureBootState(t *testing.T) {
 			t.Error("expected to see both WinProdPCA and ThirdPartyUEFI certs")
 		}
 	}
+}
+
+func TestParsingCELEventLog(t *testing.T) {
```
alexmwu, line 355: consider testing an empty CEL and a CEL without any COS types (but other random TLVs)

server/eventlog_test.go
```diff
@@ -347,7 +350,66 @@ func TestParseSecureBootState(t *testing.T) {
...
+	defer client.CheckedClose(t, tpm)
+
+	coscel := &cel.CEL{}
+
+	testEvents := []struct {
+		cosNestedEventType cel.CosType
+		pcr                int
+		eventPayload       []byte
+	}{
+		{cel.ImageRefType, 13, []byte("docker.io/bazel/experimental/test:latest")},
```
alexmwu, line 366: don't actually use PCR13 because `AppendEvent` has the side effect of extending real PCRs, and we sometimes test against real TPMs. Prefer Debug or Application

alexmwu: If we expect COS events to always be in PCR13 and want to add that check in `parseCEL`, then we can tag this or similar tests such that we only run them on simulated TPMs.

server/verify_test.go
```diff
@@ -280,3 +283,64 @@ func TestVerifySHA1Attestation(t *testing.T) {
...
+	}
+	defer ak.Close()
+
+	coscel := &cel.CEL{}
+	testEvents := []struct {
+		cosNestedEventType cel.CosType
+		pcr                int
+		eventPayload       []byte
+	}{
+		{cel.ImageRefType, 13, []byte("docker.io/bazel/experimental/test:latest")},
```
alexmwu, line 303: change to debug or application PCR

server/verify_test.go
```diff
@@ -280,3 +283,64 @@ func TestVerifySHA1Attestation(t *testing.T) {
...
+		{cel.ImageDigestType, 13, []byte("sha256:781d8dfdd92118436bd914442c8339e653b83f6bf3c1a7a98efcfb7c4fed7483")},
+		{cel.RestartPolicyType, 13, []byte(attestpb.RestartPolicy_ALWAYS.String())},
+	}
+	hashAlgoList := []crypto.Hash{crypto.SHA256, crypto.SHA1, crypto.SHA512}
+	for _, testEvent := range testEvents {
+		cos := cel.CosTlv{EventType: testEvent.cosNestedEventType, EventContent: testEvent.eventPayload}
+		err := coscel.AppendEvent(rwc, testEvent.pcr, hashAlgoList, cos)
+		if err != nil {
+			t.Fatal(err.Error())
```
alexmwu, line 312: nit: here and below, this can just be `t.Fatal(err)`

cel/cos_tlv_test.go
```diff
@@ -0,0 +1,62 @@
...
+	}{
+		{ImageRefType, 13, []byte("docker.io/bazel/experimental/test:latest")},
+		{ImageDigestType, 14, []byte("sha256:781d8dfdd92118436bd914442c8339e653b83f6bf3c1a7a98efcfb7c4fed7483")},
+		{RestartPolicyType, 15, []byte(pb.RestartPolicy_NEVER.String())},
+	}
+
+	for _, testEvent := range testEvents {
+		cos := CosTlv{testEvent.cosNestedEventType, testEvent.eventPayload}
+		if err := cel.AppendEvent(tpm, testEvent.pcr, hashAlgoList, cos); err != nil {
+			t.Fatal(err.Error())
```
alexmwu, line 34: nit: `t.Fatal(err)`

cel/canonical_eventlog.go
```diff
@@ -346,28 +352,35 @@ func DecodeToCELR(buf *bytes.Buffer) (r Record, err error) {
 	if err != nil {
 		return Record{}, err
 	}
+
+	if err := r.VerifyDigest(); err != nil {
```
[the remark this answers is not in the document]
jkl73: yes, I forgot the digest can depends on the content type...

cel/canonical_eventlog.go
```diff
@@ -389,5 +402,27 @@ func (c *CEL) replay(bank *pb.PCRs) error {
...
 	}
 
-	return fmt.Errorf("CEL replay failed for these PCRs in bank %v: %v", tpm2Alg, failedReplayPcrs)
+	return fmt.Errorf("CEL replay failed for these PCRs in bank %v: %v", cryptoHash, failedReplayPcrs)
+}
+
+// VerifyDigest checks all digests in the Record, to make sure they are equal to
+// the digests generated by hashing the content.
+func (r Record) VerifyDigest() error {
+	for hash, digest := range r.Digests {
```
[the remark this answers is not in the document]
jkl73: yes, I forgot digest can depend on the actual content

cel/canonical_eventlog_test.go
jkl73: done

cel/canonical_eventlog_test.go
```diff
@@ -124,6 +113,29 @@ func TestCELReplayFailTampered(t *testing.T) {
 		/*shouldSucceed=*/ false)
 }
 
+func TestCELReplayFailTamperedContent(t *testing.T) {
+	tpm := test.GetTPM(t)
```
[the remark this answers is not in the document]
jkl73: moving to verify_test

client/example_test.go
jkl73: Done

go.mod
jkl73: I was following the [style guide](https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#types-of-equality), but seems like using DeepEqual is fine here, and can remove a dependency?

alexmwu: Ah if it's explicitly called out, then let's migrate everything over, ideally in another PR.

jkl73: Okay, so I will keep using cmp here. And change all reflect.DeepEqual to cm.Equal in next PR.

server/eventlog_test.go
jkl73: I changed to use test.DebugPCR. If we want use real TPM for testing we need to figure out a way to reset the debug pcr before/after the test. But seems like go-tpm hasn't implement this pct.PCRReset API yet.

server/verify_test.go
jkl73: Done

server/eventlog_test.go
jkl73: Added an empty CEL test, note an empty CEL can pass Verify with any PCR values as no record can be check against with.

Test with non COS type is tricky as currently we only have one type (CosTlv) that implements the Content interface, I can try to manually create a Record and manually generate it digests.

jkl73: Added the random TLV test. Also I realized we can read PCR value directly from the TPM to do the test.

cel/cos_tlv_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	hashAlgoList := []crypto.Hash{crypto.SHA256, crypto.SHA1, crypto.SHA512}
+	cel := &CEL{}
+
+	testEvents := []struct {
+		cosNestedEventType CosType
+		pcr                int
+		eventPayload       []byte
+	}{
+		{ImageRefType, 13, []byte("docker.io/bazel/experimental/test:latest")},
```
alexmwu, line 26: Sorry missed this: use debug or application here and below as well.

cel/canonical_eventlog.go
```diff
@@ -389,5 +394,20 @@ func (c *CEL) replay(bank *pb.PCRs) error {
 		return nil
 	}
 
-	return fmt.Errorf("CEL replay failed for these PCRs in bank %v: %v", tpm2Alg, failedReplayPcrs)
+	return fmt.Errorf("CEL replay failed for these PCRs in bank %v: %v", cryptoHash, failedReplayPcrs)
+}
+
+// VerifyDigest checks the digest generated by the given recornd content to make sure they are equal to
```
alexmwu, line 400: nit: "generated by the given recornd content to" -> "generated by the given record's content to"

server/example_test.go
```diff
@@ -1,14 +1,15 @@
...
 	"log"
 
 	"github.com/google/go-tpm-tools/client"
 	"github.com/google/go-tpm-tools/simulator"
-	"github.com/google/go-tpm/tpm2"
 )
 
-func ExampleParseMachineState() {
+func ExampleVerifyAttestation() {
 	// On client machine, generate the TPM quote.
```
alexmwu, line 13: I wonder if it makes sense to keep this test around given that we already have the Attest example. The only reason I can think of is for clear examples for both `server.VerifyAttestation` and `client.Attest`, using many of the available opts.

Just some bikeshedding here. Either way, don't need to worry about it in this PR.

server/eventlog_test.go
```diff
@@ -347,7 +354,133 @@ func TestParseSecureBootState(t *testing.T) {
...
+	defer client.CheckedClose(t, tpm)
+
+	coscel := &cel.CEL{}
+	hashAlgoList := []crypto.Hash{crypto.SHA256, crypto.SHA1, crypto.SHA384, crypto.SHA512}
+	emptyCosState := attestpb.ContainerState{}
+
+	var buf bytes.Buffer
+	// encode an empty CEL
+	if err := coscel.EncodeCEL(&buf); err != nil {
+		t.Fatalf(err.Error())
```
alexmwu, line 370: nit: `t.Fatal(err)`

server/eventlog_test.go
alexmwu: thanks!
small nit: maybe break these up into different tests as they are starting to be difficult to read
feel free to leave if you disagree.

alexmwu (approved):
Thanks Jiankun!

cel/cos_tlv_test.go
jkl73: Done, I missed as well

cel/canonical_eventlog.go
jkl73: Done

server/example_test.go
jkl73: ya, seems like it is a little bit repetitive. Problem is that these two tests are touching both client and server. We can find a better place one of them later

server/eventlog_test.go
jkl73: Done

cel/canonical_eventlog.go
```diff
@@ -389,5 +394,20 @@ func (c *CEL) replay(bank *pb.PCRs) error {
 		return nil
 	}
 
-	return fmt.Errorf("CEL replay failed for these PCRs in bank %v: %v", tpm2Alg, failedReplayPcrs)
+	return fmt.Errorf("CEL replay failed for these PCRs in bank %v: %v", cryptoHash, failedReplayPcrs)
+}
+
+// VerifyDigest checks the digest generated by the given record's content to make sure they are equal to
+// the digests in the digestMap.
+func VerifyDigest(c Content, digestMap map[crypto.Hash][]byte) error {
```
josephlr, line 402: Nit: This should be called `VerifyDigests` (as it verifies multiple digests)

server/verify.go
```diff
@@ -77,14 +78,22 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
...
 		pcrs := quote.GetPcrs()
-		state, err := ParseMachineState(attestation.GetEventLog(), pcrs)
+		state, err := parsePCClientEventLog(attestation.GetEventLog(), pcrs)
 		if err != nil {
-			lastErr = fmt.Errorf("failed to validate the event log: %w", err)
+			lastErr = fmt.Errorf("failed to validate the PCClient event log: %w", err)
 			continue
 		}
 
+		celState, err := parseCanonicalEventLog(attestation.CanonicalEventLog, pcrs)
```
josephlr, line 89: Nit: `attestation.GetCanonicalEventLog()` for consistency with above code

server/eventlog.go
```diff
@@ -53,6 +54,17 @@ func ParseMachineState(rawEventLog []byte, pcrs *tpmpb.PCRs) (*pb.MachineState,
...
+func parseCanonicalEventLog(rawCanonicalEventLog []byte, pcrs *tpmpb.PCRs) (*pb.MachineState, error) {
+	cosState, err := getVerifiedCosState(rawCanonicalEventLog, pcrs)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pb.MachineState{
+		Cos: cosState,
+	}, err
+}
```
josephlr, line 66: Nit: Why not just have this be one method instead of two methods (where one doesn't do anything.

Specifically, why do we not just have the implementation of `getVerifiedCosState` live in `parseCanonicalEventLog`

server/eventlog.go
```diff
@@ -62,6 +74,54 @@ func contains(set [][]byte, value []byte) bool {
...
+
+		switch cosTlv.EventType {
+		case cel.ImageRefType:
+			cosState.Container.ImageReference = string(cosTlv.EventContent)
+
+		case cel.ImageDigestType:
+			cosState.Container.ImageDigest = cosTlv.EventContent
+
+		case cel.RestartPolicyType:
+			restartPolicy, ok := pb.RestartPolicy_value[string(cosTlv.EventContent)]
```
josephlr, line 115: So are the measured values in the CEL log going to look like: `"OnFailure"` (the style from [Pod Lifecycle](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/)) or `"ON_FAILURE"` (the style from [our protobuf](https://github.com/google/go-tpm-tools/blob/bc3fe1ecfe70da837b051fc8d81dfaa22bfa5170/proto/attest.proto#L114-L122))

josephlr (approved):
Looks good modulo nits

cel/canonical_eventlog.go
jkl73: Done

server/verify.go
jkl73: Done

server/eventlog.go
jkl73: I am thinking maybe later if there are more canonical event log states other than the COS state, we can add them here.

server/eventlog.go
jkl73: It will use protobuf ON_FAILURE, when the launcher is writing the log, it needs to use the protobuf enum string.

josephlr: Ok, I have an idea about how to make this more ergonomic, but that can be in a followup PR

server/eventlog.go
josephlr: Ok that makes sense. If we are doing that split, then we might want to do something like:

In `parseCanonicalEventLog`:
  - Call DecodeToCEL
  - Do the PCR replay

Then have `getVerifiedCosState` only take a `cel.CEL`, and it would:
  - Loop though all the records
  - Create (and return) the container/cos state

IMO, this sepearates the COS specific stuff from the generic CEL stuff

jkl73: make sense, Done

server/eventlog.go
```diff
@@ -62,6 +83,44 @@ func contains(set [][]byte, value []byte) bool {
...
+		if err := cel.VerifyDigests(cosTlv, record.Digests); err != nil {
+			return nil, err
+		}
+
+		switch cosTlv.EventType {
+		case cel.ImageRefType:
+			cosState.Container.ImageReference = string(cosTlv.EventContent)
+
+		case cel.ImageDigestType:
+			cosState.Container.ImageDigest = cosTlv.EventContent
```
josephlr, line 111: It seems like you synced on Alex's changes, so this needs to be `cosState.Container.ImageDigest = string(cosTlv.EventContent)`

jkl73: yes done

PR #147, opened 2021-12-07
Add COS measurements to MachineState

alexmwu:
Allows MachineState to specify COS-specific container measurements such as the image ref, digest, and restart policy (as a start).


proto/attest.proto
```diff
@@ -108,6 +108,26 @@ message SecureBootState {
...
 
+// The container's restart policy.
+// See https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy
+// for more details.
+enum RestartPolicy {
+  NO = 0;
+  ON_FAILURE = 1;
+  ALWAYS = 2;
+  UNLESS_STOPPED = 3;
+}
```
jkl73, line 119: I am not sure we should use docker's policy here, it seems more complex. I prefer the [kubelet](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#:~:text=The%20spec%20of%20a%20Pod%20has%20a%20restartPolicy%20field%20with%20possible%20values%20Always%2C%20OnFailure%2C%20and%20Never.%20The%20default%20value%20is%20Always.) (which is also the [konlet](https://github.com/GoogleCloudPlatform/konlet/blob/9cb9106daf07123c2641159cb8bcc9d6f4960ec2/gce-containers-startup/types/api.go)) style , which is 

Always, OnFailure, and Never

alexmwu: Done, thanks!

jkl73 (approved):

PR #148, opened 2021-12-09
Add CEL replay and TPM extension in AppendEvent

alexmwu:


cel/canonical_eventlog_test.go
```diff
@@ -3,49 +3,52 @@ package cel
...
+		t.Fatal(err)
 	}
 	decodedcel, err := DecodeToCEL(&buf)
 	if err != nil {
-		t.Fatalf(err.Error())
+		t.Fatal(err)
 	}
 	if len(decodedcel.Records) != 2 {
-		t.Errorf("should have two records")
+		t.Errorf("should have three records")
```
jkl73, line 40: Two or three?

cel/canonical_eventlog_test.go
```diff
@@ -56,7 +59,99 @@ func TestCELEncodingDecoding(t *testing.T) {
...
+	appendOrFatal(t, cel, tpm, test.ApplicationPCR, measuredHashes, cosEvent)
+	appendOrFatal(t, cel, tpm, test.ApplicationPCR, measuredHashes, cosEvent)
+
+	modifiedRecord := cel.Records[3]
+	for hash := range modifiedRecord.Digests {
+		newDigest := make([]byte, hash.Size())
+		rand.Read(newDigest)
+		modifiedRecord.Digests[hash] = newDigest
+	}
+	replay(t, cel, tpm, measuredHashes,
```
jkl73, line 117: We should test replay using a subset/superset of the measuredHashes, to see its behavior

cel/canonical_eventlog.go
```diff
@@ -340,3 +348,39 @@ func DecodeToCELR(buf *bytes.Buffer) (r Record, err error) {
...
+		if !bytes.Equal(bank.Pcrs[uint32(simulatedPcr)], digest) {
+			failedReplayPcrs = append(failedReplayPcrs, simulatedPcr)
+		}
+	}
+
+	if len(failedReplayPcrs) == 0 {
+		return nil
+	}
+
+	return fmt.Errorf("CEL replay failed for these PCRS: %v", failedReplayPcrs)
```
jkl73, line 385: We should add hash to the error msg.

cel/canonical_eventlog.go
```diff
@@ -340,3 +348,39 @@ func DecodeToCELR(buf *bytes.Buffer) (r Record, err error) {
 	}
 	return r, nil
 }
+
+// Replay takes the digests from a Canonical Event Log and carries out the
+// extend sequence for each PCR in the log. It then compares the final digests
+// against a bank of PCR values to see if they match.
+func (c *CEL) Replay(bank *pb.PCRs) error {
```
jkl73, line 355: I am wondering what's the behavior of Replay.

* I am thinking should `Replay()` just return a list of all final PCRs values involved in the CEL, and we do the validation in later/in another function

cel/canonical_eventlog.go
```diff
@@ -340,3 +348,39 @@ func DecodeToCELR(buf *bytes.Buffer) (r Record, err error) {
...
+	replayed := make(map[uint8][]byte)
+	// initBank := make([]byte, cryptoHash.Size())
+	for _, record := range c.Records {
+		if _, ok := replayed[record.PCR]; !ok {
+			replayed[record.PCR] = make([]byte, cryptoHash.Size())
+		}
+		hasher := cryptoHash.New()
+		digestsMap := record.Digests
+		hasher.Write(replayed[record.PCR])
+		hasher.Write(digestsMap[cryptoHash])
```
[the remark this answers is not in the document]
alexmwu: yes, good point thanks!

cel/canonical_eventlog.go
alexmwu: I'm  what the point of replay is without comparing it. What's your underlying concern? I can rename it `Verify` if you don't like the name and pull replay out to an unexported function if you want to use it for something else.

jkl73: My main concern is the hash algorithm selection.
When a user call Replay(), they won't have an idea about the what pcr hashalgo they are facing (they can try to read the Record directly to get the hashalgo but it won't be straight forward), they might be confused about what hashalgo to set in *pb.PCRs.

alexmwu: talked offline:
We will unexport Replay for now, since there is no need for this to be user facing at the moment. It will be integrated into VerifyAttestation or ParseMachineState where the caller will try all the given banks to replay. If the hashAlgo is not found in even one record, the whole replay will fail.

Also, we discussed how this replay behavior differs from go-attestation. go-attestation checks that the intersecting PCRs of the log and the bank match digests. Our approach will be to see if PCRs seen in the CEL are a subset of the ones passed in. If so, we check that the subset's PCRs match digests with the passed in bank.

cel/canonical_eventlog_test.go
alexmwu: Done.

cel/canonical_eventlog_test.go
```diff
@@ -56,7 +59,120 @@ func TestCELEncodingDecoding(t *testing.T) {
...
+	cosEvent2 := CosTlv{someEvent2}
+	appendOrFatal(t, cel, tpm, test.DebugPCR, measuredHashes, cosEvent)
+	appendOrFatal(t, cel, tpm, test.DebugPCR, measuredHashes, cosEvent2)
+
+	appendOrFatal(t, cel, tpm, test.ApplicationPCR, measuredHashes, cosEvent2)
+	appendOrFatal(t, cel, tpm, test.ApplicationPCR, measuredHashes, cosEvent)
+	appendOrFatal(t, cel, tpm, test.ApplicationPCR, measuredHashes, cosEvent)
+
+	replay(t, cel, tpm, measuredHashes,
+		[]int{test.DebugPCR, test.ApplicationPCR} /*shouldSucceed=*/, true)
```
jkl73, line 88: nit: should the inline comment locate after the comma?
[]int{test.DebugPCR, test.ApplicationPCR}, /*shouldSucceed=*/ true)

cel/canonical_eventlog.go
```diff
@@ -340,3 +348,46 @@ func DecodeToCELR(buf *bytes.Buffer) (r Record, err error) {
...
+		if !bytes.Equal(bankDigest, replayDigest) {
+			failedReplayPcrs = append(failedReplayPcrs, replayPcr)
+		}
+	}
+
+	if len(failedReplayPcrs) == 0 {
+		return nil
+	}
+
+	return fmt.Errorf("CEL replay failed for these PCRS in bank %v: %v", tpm2Alg, failedReplayPcrs)
```
jkl73, line 392: nit: PCRs

client/attest.go
```diff
@@ -19,6 +19,11 @@ type AttestOpts struct {
 	// attestation protocols:
 	// https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.70.4562&rep=rep1&type=pdf
 	Nonce []byte
+	// TCG Canonical Event Log to add to the attestation.
+	// Currently, we only support PCR replay for PCRs orthogonal to those in the
+	// firmware event log, where PCRs 0-9 and 14 are often measured. If the two
+	// logs overlap, server-side verification using this library may fail.
+	CanonicalEventLog []byte
```
jkl73, line 26: Naming question, should we call it CanonicalEventLog (the format) or COSEventLog (the actual content)

jkl73 (approved):

cel/canonical_eventlog_test.go
alexmwu: I agree, however the VSCode golang formatter overwrites putting it after the ",". I can put it on the next line if it bothers you.

client/attest.go
alexmwu: I asked this question and I believe Joe wanted to separate based on format. I think mostly because the type in each TLV should give us guidance that they are COS events vs other potential events

cel/canonical_eventlog_test.go
jkl73: Never mind, lets follow the formatter.

client/attest.go
jkl73: SGTM

cel/canonical_eventlog_test.go
alexmwu: nvm I just went ahead and put it on the next line.

PR #149, opened 2021-12-13
Add AK cert support in VerifyAttestation

alexmwu:
client: Added a cert field to client.Key and ak_cert to `Attestation`. Eventually, users will be able to manually add a cert and we can add support for fetching EK certs from the known index. For GCE AKs, this fetches the cert from the known NV indexes.

server: VerifyAttestation now takes in a Root and Intermediate CertPool. Used to verify the ak_cert against trusted CAs.
This adds a dependency on https://pkg.go.dev/github.com/google/certificate-transparency-go/x509 due to a different SAN extension and unknown OID extended key usage for EKs/AKs.


iKevinY (approved):
Code style LGTM, but pending @jkl73's and @josephlr's reviews on implementation specifics as usual.

client/keys.go
```diff
@@ -427,3 +447,28 @@ func (k *Key) Reseal(in *pb.SealedBytes, uOpts UnsealOpts, sOpts SealOpts) (*pb.
...
 	return k.pubArea.Attributes&attr != 0
 }
+
+// Cert returns the parsed certificate for the given key.
+func (k *Key) Cert() *x509.Certificate {
+	return k.cert
+}
+
+// CertRaw provides the ASN.1 DER content of the key's certificate.
+func (k *Key) CertRaw() []byte {
```
josephlr, line 457: Let's explicitly call this `CertDERBytes()` or something more specific than `Raw()`.

server/verify.go
```diff
@@ -22,14 +25,21 @@ type VerifyOpts struct {
...
 	// However, setting this to true may be necessary if the client only
 	// supports the legacy event log format. This is the case on older Linux
 	// distributions (such as Debian 10).
 	AllowSHA1 bool
+	// A collection of trusted root CAs that are used to sign AK certificates.
+	// The TrustedAKs are used first, followed by TrustRoots and Intermediates.
+	// Adding a specific TPM manufacturer's root or intermediate CA means all
+	// TPMs signed by that CA will be trusted.
+	TrustedRoots         *x509.CertPool
+	TrustedIntermediates *x509.CertPool
```
josephlr, line 42: Should we just use an `x509.VerifyOptions` structure (https://pkg.go.dev/crypto/x509#VerifyOptions) it already includes Roots and Intermediates? and just override the incompatible fields?

alexmwu: That's fine: we'd just have to document that we only take in `Roots` and `Intermediates` from that struct. At that point, it may just be easier to be clear that we only accept those two via top level fields. The only other field of `VerifyOptions` that seems to make sense to expose is `CurrentTime`.

I personally think it's a cleaner abstraction to accept the two fields over a `VerifyOptions` because adding that as a field signals to users of the API that we use other things in `VerifyOptions`

josephlr: Actually I think this is fine. However, we should change the names to `TrustedRootCerts` and `IntermediateCerts` (as the intermediates are not trusted here.

client/keys.go
```diff
@@ -67,14 +69,32 @@ func EndorsementKeyFromNvIndex(rw io.ReadWriter, idx uint32) (*Key, error) {
...
 // GceAttestationKeyECC generates and loads the GCE ECC AK. Note that this
 // function will only work on a GCE VM. Unlike AttestationKeyECC, this key uses
 // the Endorsement Hierarchy and its template loaded from GceAKTemplateNVIndexECC.
 func GceAttestationKeyECC(rw io.ReadWriter) (*Key, error) {
-	return EndorsementKeyFromNvIndex(rw, GceAKTemplateNVIndexECC)
+	akEcc, err := EndorsementKeyFromNvIndex(rw, GceAKTemplateNVIndexECC)
+	if err != nil {
+		return nil, err
+	}
+	akCert, err := getCertificateFromNvram(rw, GceAKCertNVIndexECC)
```
josephlr, line 92: Here (and for any other GCE keys), we shoudn't fail if the cert isn't present, just don't set it.

client/keys.go
```diff
@@ -67,14 +69,32 @@ func EndorsementKeyFromNvIndex(rw io.ReadWriter, idx uint32) (*Key, error) {
 // function will only work on a GCE VM. Unlike AttestationKeyRSA, this key uses
 // the Endorsement Hierarchy and its template loaded from GceAKTemplateNVIndexRSA.
 func GceAttestationKeyRSA(rw io.ReadWriter) (*Key, error) {
-	return EndorsementKeyFromNvIndex(rw, GceAKTemplateNVIndexRSA)
+	akRsa, err := EndorsementKeyFromNvIndex(rw, GceAKTemplateNVIndexRSA)
```
josephlr, line 72: We should also get the EK Certs for the standard EK keys (if present)

server/policy_constants.go
```diff
@@ -60,6 +60,14 @@ var (
...
 )
 
+// Known Privacy CA certs.
+var (
+	//go:embed ca-certs/tpm_ek_root_1.cer
+	GceEKRootCA []byte
+	//go:embed ca-certs/tpm_ek_intermediate_2.crt
+	GceEKIntermediateCA2 []byte
+)
+
```
josephlr, line 70: Nit, these should both be `*x509.CertPool`s (created from the embedded data), and they should be named something like: `GceEKRoots` and `GceEKIntermediates`

client/keys.go
alexmwu: Done.

PR #150, opened 2021-12-22
proto: Add rest of AttestedCosState measurements

alexmwu:


proto/attest.proto
```diff
@@ -125,10 +125,15 @@ message ContainerState {
   string image_reference = 1;
   bytes image_digest = 2;
   RestartPolicy restart_policy = 3;
+  string image_id = 4;
+  repeated string args = 5;
+  map<string, string> env_vars = 6;
```
alexmwu, line 130: in the launcher we use a slice of `EnvVar`s. Do we want something similar here (maybe because other metadata will be attached to a specific kv entry) or just leave it as an actual map.

josephlr: A map is fine, no reason to make it overly complicated

proto/attest.proto
```diff
@@ -125,10 +125,15 @@ message ContainerState {
   string image_reference = 1;
   bytes image_digest = 2;
   RestartPolicy restart_policy = 3;
+  string image_id = 4;
```
josephlr, line 128: Having `image_id` be a string is good, as the data will look like `sha256:<hex>`. For similar reasons we should make the `image_digest` also be a string (as it has the same format).

Also, can you add a comment for these two fields explaining:
  - their format
  - how they are different

proto/attest.proto
```diff
@@ -125,10 +125,15 @@ message ContainerState {
...
+  repeated string args = 5;
+  map<string, string> env_vars = 6;
 }
 
 message AttestedCosState {
   ContainerState container = 1;
+  int32 cos_version = 2;
+  int32 launcher_version = 3;
```
josephlr, line 136: Why int32 instead of an unsigned number, or a major/minor helper struct?

josephlr (approved):

proto/attest.proto
alexmwu: Good catch on the signed. Do we expect to have major/minor versions? What will claims look like?

proto/attest.proto
alexmwu: Ack, thanks

proto/attest.proto
alexmwu: as this isn't yet used, should we do a back incompat change? or would you prefer a deprecated comment and new field?

proto/attest.proto
jkl73: Do we need to worry about the order of the env vars here? My intuition is no, so a Map is okay. That means Attestation service may read the Env Var in a different order than it is measured, but I don't that's an issue.

jkl73 (approved):

proto/attest.proto
josephlr: Backwards incompatible is fine, also [changing a `bytes` field to a `string` field is backwards compatible](https://developers.google.com/protocol-buffers/docs/overview#updating) in general (provided that the bytes messages have always contained valid UTF8. No really relevent here, but an interesting fact.

proto/attest.proto
josephlr: I agree that the ordering doesn't seem like an issue here.

josephlr (approved):

proto/attest.proto
alexmwu: cool, TIL!

PR #153, opened 2022-01-14
server: Fix AK Cert checking bug 

josephlr:
Without this change, providing a good AK Cert but a bad AKPub will result in VerifyAttestation succeeding. Now, we check that the AKPub and AKCert (if present) have the same public key.

I also reorganized the code into one `checkAKTrusted` method, which should make things easier to understand.

### DIFF
--- a/server/verify.go
+++ b/server/verify.go
@@ -67,10 +67,8 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
 	if err != nil {
 		return nil, fmt.Errorf("failed to get AK public key: %w", err)
 	}
-	if err := checkAkTrusted(akPubKey, opts); err != nil {
-		if err := validateAkCert(attestation.AkCert, opts.IntermediateCerts, opts.TrustedRootCerts); err != nil {
-			return nil, fmt.Errorf("failed to validate attestation key: AKPub is untrusted and %v", err)
-		}
+	if err := checkAKTrusted(akPubKey, attestation.GetAkCert(), opts); err != nil {
+		return nil, fmt.Errorf("failed to validate AK: %w", err)
 	}
 
 	// Verify the signing hash algorithm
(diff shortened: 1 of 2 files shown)


server/verify.go
```diff
@@ -126,7 +124,7 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
 	return nil, fmt.Errorf("attestation does not contain a supported quote")
 }
 
-func pubKeysEqual(k1 crypto.PublicKey, k2 crypto.PublicKey) bool {
```
alexmwu, line 129: why change to `interface{}`? `crypto.PublicKey` is also an empty interface but it at least conveys that the input should be a PublicKey more than `interface{}`

server/verify.go
```diff
@@ -138,33 +136,40 @@ func pubKeysEqual(k1 crypto.PublicKey, k2 crypto.PublicKey) bool {
 }
 
 // Checks if the provided AK public key can be trusted
-func checkAkTrusted(ak crypto.PublicKey, opts VerifyOpts) error {
-	if len(opts.TrustedAKs) == 0 {
-		return fmt.Errorf("no mechanism for AK verification provided")
+func checkAKTrusted(ak crypto.PublicKey, akCertBytes []byte, opts VerifyOpts) error {
+	checkPub := len(opts.TrustedAKs) > 0
+	checkCert := len(opts.TrustedRootCerts.Subjects()) > 0
```
alexmwu, line 141: the tests are failing (https://github.com/google/go-tpm-tools/runs/4812554993?check_suite_focus=true) because you need to check that opts.TrustedRootCerts != nil

alexmwu (approved):
LGTM after fixing tests

server/verify.go
```diff
@@ -138,33 +136,40 @@ func pubKeysEqual(k1 crypto.PublicKey, k2 crypto.PublicKey) bool {
...
 	}
 
 	// Check against known AKs
-	for _, trusted := range opts.TrustedAKs {
-		if pubKeysEqual(ak, trusted) {
-			return nil
+	if checkPub {
+		for _, trusted := range opts.TrustedAKs {
+			if pubKeysEqual(ak, trusted) {
+				return nil
```
chrisfenner, line 150: I'm not sure if we expect a caller to ever provide both `checkPub` and `checkCert` or what such a caller should expect (boolean union or intersection of those checks)? Should we prefer to indent error flow here? https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

server/verify.go
josephlr: Whoops, this was because [`x509.Certificate.PublicKey`](https://pkg.go.dev/crypto/x509#Certificate) is an `interface{}` and that's what we pass to `pubKeysEqual`. I din't realize that `crypto.PublicKey` is also just an empty interface.

server/verify.go
josephlr: Fixed

server/verify.go
josephlr: I fixed the `checkAKTrusted` to require that exactly one of `TrustedAKs` or `TrustedRootCerts` is used.

Thanks for the advice.

josephlr:
I added some additional tests to make sure we don't regress w.r.t. this bug.

PR #157, opened 2022-01-25
server: Support all public key types in pubKeysEqual

josephlr:
Mostly just a style nit

Instead of checking the concrete types for `crypto.PublicKey`, we try to cast the key to an interface containing the expected `Equals` method. This approach is [explicitly mentioned in the `crypto.PublicKey` documentation](https://cs.opensource.google/go/go/+/refs/tags/go1.18beta1:src/crypto/crypto.go;l=154-161).

### DIFF
--- a/server/verify.go
+++ b/server/verify.go
@@ -123,7 +123,8 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
 }
 
 func pubKeysEqual(k1 crypto.PublicKey, k2 crypto.PublicKey) bool {
-	// Common interface for all public keys (see crypto.PublicKey documentation)
+	// Common interface for all the standard public key types, see:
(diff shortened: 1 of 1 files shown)


server/verify.go
```diff
@@ -125,14 +123,14 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
 }
 
 func pubKeysEqual(k1 crypto.PublicKey, k2 crypto.PublicKey) bool {
-	switch key := k1.(type) {
```
alexmwu, line 128: this makes sense, but could you link to the exact documentation (I couldn't find it).

alexmwu: I see a proposal here that was accepted and closed: https://github.com/golang/go/issues/21704

alexmwu (approved):

server/verify.go
josephlr: Done, it's in the 1.18 beta docs

PR #161, opened 2022-02-10
Add OverrideEnvType and OverrideArgType for cel event

jkl73:


josephlr (approved):
This looks good for the CEL types. The plan is still to just have two booleans in the machine state indicating if args (or respectively envs) were overwritten as part of container launch, correct?

jkl73:
> This looks good for the CEL types. The plan is still to just have two booleans in the machine state indicating if args (or respectively envs) were overwritten as part of container launch, correct?

Yes we will do that for now.
Later we can specify the actual overwritten args envs  (strings instead of a boolean) in the machine state. That way trusted party can set a more granular rule on envs and args.

PR #164, opened 2022-02-15
Add AK certificate chain retrieval

jessieqliu:
Adds an optional `intermediate_certs` field to `Attestation` to store the certificate chain for the AK cert. If the `validateCertChain` flag in `AttestOpts` is specified, `Key.Attest` will retrieve the certificate chain for the AK cert.

Also updates documentation with instructions to resolve a `program is not executable` error for `protoc-gen-go`, based on [this thread](https://stackoverflow.com/questions/57700860/protoc-gen-go-program-not-found-or-is-not-executable)


client/attest.go
```diff
@@ -24,6 +27,39 @@ type AttestOpts struct {
 	// firmware event log, where PCRs 0-9 and 14 are often measured. If the two
 	// logs overlap, server-side verification using this library may fail.
 	CanonicalEventLog []byte
+	// Indicates whether the AK certificate chain should be retrieved for validation.
```
josephlr, line 30: We should make it clearer here that this will potentially make external network calls. Maybe calling it `FetchCertChain`, and making it clear _which_ certs are being fetched (i.e. from `IssuingCertificateURL`) and how.

client/attest.go
```diff
@@ -24,6 +27,39 @@ type AttestOpts struct {
...
+			return nil, fmt.Errorf("certificate retrieval from %s returned non-OK status: %v", url, resp.StatusCode)
+		}
+		certBytes, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read response body from %s: %v", url, err)
+		}
+
+		// Verify that the bytes can be parsed into a certificate.
+		_, err = x509.ParseCertificate(certBytes)
+		if err != nil {
```
josephlr, line 55: Nit: one line

client/attest.go
```diff
@@ -24,6 +27,39 @@ type AttestOpts struct {
...
+			return nil, fmt.Errorf("failed to retrieve certificate at %v: %v", url, err)
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("certificate retrieval from %s returned non-OK status: %v", url, resp.StatusCode)
+		}
+		certBytes, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read response body from %s: %v", url, err)
```
josephlr, line 50: Here and elsewhere, use `%w` when wrapping another error.

client/attest.go
```diff
@@ -24,6 +27,39 @@ type AttestOpts struct {
 	// firmware event log, where PCRs 0-9 and 14 are often measured. If the two
 	// logs overlap, server-side verification using this library may fail.
 	CanonicalEventLog []byte
+	// Indicates whether the AK certificate chain should be retrieved for validation.
+	ValidateCertChain bool
+}
+
+// Constructs the certificate chain for the key's certificate, using the provided HTTP client.
+func (k *Key) getCertificateChain() ([][]byte, error) {
```
josephlr, line 35: So this seems like it would just get the first step in the cert chain, but not necessarily _all_ steps in the cert chain.

For example, if our Leaf cert contains one IssuingCertificateURL, we will just fetch that single cert, but we won't do this process recursively. What we really need to do is the following:

For each step in our cert chain:
  - Make sure there aren't too many IssuingCertificateURLs (3 max is fine).
  - For each URL in IssuingCertificateURL:
    - Try to `Get` the payload
    - Try to parse the cert
    - Check if the cert we got is what we expect (i.e. it signs the current cert in our chain)
    - If this doesn't work, move on to next url
  - append the cert to the chain
  - Take the cert we just appened in the previous step and repeat the process 
  - Cap the total depth we can go (4 max is fine)

client/attest.go
```diff
@@ -62,5 +98,14 @@ func (k *Key) Attest(opts AttestOpts) (*pb.Attestation, error) {
 	if len(opts.CanonicalEventLog) != 0 {
 		attestation.CanonicalEventLog = opts.CanonicalEventLog
 	}
+
+	// Construct certficate chain.
+	if opts.ValidateCertChain {
```
josephlr, line 103: We should also check that the key even has a cert (otherwise I think we will have a nil dereference).

client/attest_test.go
```diff
@@ -0,0 +1,110 @@
+package client
```
josephlr, line 1: For this package we should add two additional types of tests:

1. A test that actually calls `client.Attest` to make sure the `ValidateCertChain` option works correctly.
2. A test that calls `client.Attest` with ValidateCertChain = true`, but where the key does not have a certificate.
  a. We should decide whether to error or do-nothing in this case.

josephlr (commented):
@jkl73 can you finish up the review here? I'll be out of town this week.

client/attest.go
jessieqliu: If, at any depth, we go through a list of IssuingCertificateURLs and can't find the expected one, should it return an error or just provide the certificates so far (if any).

jkl73: IIUC, we should return an error if we cannot find a root cert. (Or it'll be checked elsewhere?)

jessieqliu: The root cert should be checked by the attestation service when it verifies the attestation, we aren't expecting to check it here.

go.mod
```diff
@@ -4,6 +4,7 @@ module github.com/google/go-tpm-tools
 go 1.16
 
 require (
+	github.com/golang/protobuf v1.5.2 // indirect
```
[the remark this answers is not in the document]
jessieqliu: Removed

client/attest.go
```diff
@@ -24,6 +33,78 @@ type AttestOpts struct {
...
+	}
+	// For each URL, fetch and parse the certificate, then verify whether it signed cert.
+	// If successful, return the parsed certificate. If any step in this process fails, try the next url.
+	for _, url := range cert.IssuingCertificateURL {
+		resp, err := http.Get(url)
+		if err != nil {
+			log.Printf("failed to retrieve certificate at %v: %v\n", url, err)
+			continue
+		}
+		defer resp.Body.Close()
```
jkl73, line 56: I think we should just read the response body and close it inside the current loop iteration, instead of deferring it (because if defer to func end, only the last `resp.Body` can be closed).

client/attest.go
```diff
@@ -24,6 +33,78 @@ type AttestOpts struct {
...
+	if len(k.cert.IssuingCertificateURL) > maxIssuingCertificateURLs {
+		return nil, fmt.Errorf("key cert contains too many issuing URLS: got %v, expect no more than %v", len(k.cert.IssuingCertificateURL), maxIssuingCertificateURLs)
+	}
+
+	var certs [][]byte
+	currentCert := k.cert
+	for i := 0; i < maxCertChainDepth; i++ {
+		issuingCert, err := fetchIssuingCertificate(currentCert)
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving certificate chain: %w", err)
```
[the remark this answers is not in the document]
jessieqliu: Per discussion over chat, removed error returns from `fetchIssuingCertificate` and `getCertificateChain`. If an error is encountered, `fetchIssuingCertificate` will return nil and `getCertificateChain` will return whatever part of the chain has been constructed so far.

We are avoiding returning errors here because it is possible to have issues retrieving intermediate certificates, but still have enough to allow `VerifyAttestation` to succeed later on. Therefore we do not want `key.Attest` to fail due to cert chain problems.

jkl73 (approved):
LGTM

PR #165, opened 2022-02-17
Fix tests for physical TPMs

jkl73:
Some tests were failing on some physical TPMs due to certain hash algos are not supported. 
Changing them to SHA1 and SHA256 which most TPM2 chips should support.

Reset the Test PCR for certain tests where requiring the PCR bank to be zero value at the beginning of the test.


josephlr (approved):
Nice fix, did you make sure these tests passed on a physical tpm?

jkl73:
> Nice fix, did you make sure these tests passed on a physical tpm?

All tests in `client` and `cel` passed. Some tests in `server` failed due to the PCR 0 replay issue on the system eventlog (maybe because I am using Debian).

PR #167, opened 2022-03-02
Add Key.SetCert() function

jessieqliu:
Add a SetCert() function to client.Key that allows setting the AK Certificate. The function will verify that the cert matches Key.pubKey.

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -471,11 +471,7 @@ func (k *Key) CertDERBytes() []byte {
 
 // SetCert assigns the provided certificate to the key after verifying it matches the key.
 func (k *Key) SetCert(cert *x509.Certificate) error {
-	certPubKey, ok := cert.PublicKey.(crypto.PublicKey)
-	if !ok {
-		return errors.New("certificate public key is not the appropriate type")
-	}
-
+	certPubKey := cert.PublicKey.(crypto.PublicKey) // This cast cannot fail
 	if !internal.PubKeysEqual(certPubKey, k.pubKey) {
 		return errors.New("certificate does not match key")
 	}
--- a/client/keys.go
+++ b/client/keys.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"reflect"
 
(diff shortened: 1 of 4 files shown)


jkl73 (approved):
LGTM

client/keys.go
```diff
@@ -468,6 +470,21 @@ func (k *Key) CertDERBytes() []byte {
...
 }
 
+// SetCert assigns the provided certificate to the key after verifying it matches the key.
+func (k *Key) SetCert(cert *x509.Certificate) error {
+	certPubKey, ok := cert.PublicKey.(crypto.PublicKey)
+	if !ok {
+		return errors.New("certificate public key is not the appropriate type")
+	}
+
+	if !reflect.DeepEqual(certPubKey, k.pubKey) {
```
josephlr, line 480: I think we already have a method to compare two `crypto.PublicKey`s

https://github.com/google/go-tpm-tools/blob/66ec981c39007db9718f14f34c15102cba642499/server/verify.go#L125

We should move that method from `server` to `internal` and then just use it here

client/keys.go
```diff
@@ -468,6 +470,21 @@ func (k *Key) CertDERBytes() []byte {
 	return k.cert.Raw
 }
 
+// SetCert assigns the provided certificate to the key after verifying it matches the key.
+func (k *Key) SetCert(cert *x509.Certificate) error {
```
josephlr, line 474: Nit: we should also be using this method internally for the various methods that set certificates, basically the 4 methods that call: `getCertificateFromNvram`.

josephlr: You may have to change that helper function to be a helper method, but it should be fine.

josephlr:
@jkl73 @jessieqliu I added two commits to Jessie's changes. Can you take a look (specifically at https://github.com/google/go-tpm-tools/pull/167/commits/11161ba2d64d6289c68c0d2cc6bef86b36ff85ed). If my changes are OK, merge this PR.

josephlr (approved):

jkl73 (approved):
lgtm

PR #168, opened 2022-03-08
client: Have client.Attest return an error on bad certs 

josephlr:
If a user explictly requests that we construct the EK/AK Cert chain, we
should surface any network/format errors on the client side, rather than
silently succeeding. This makes debugging cert chain issues much easier.

Also, we shouldn't be calling logging methods in our library (as it might interfere with the calling application's logging).


client/attest.go
```diff
@@ -39,74 +38,74 @@ type AttestOpts struct {
...
-		if i == maxIssuingCertificateURLs {
-			log.Printf("Reached the maximum number of attempts (%v)", maxIssuingCertificateURLs)
-			return nil
+		if i >= maxIssuingCertificateURLs {
+			break
 		}
 		resp, err := http.Get(url)
 		if err != nil {
-			log.Printf("failed to retrieve certificate at %v: %v\n", url, err)
+			certErr = fmt.Errorf("failed to retrieve certificate at %v: %w", url, err)
```
jessieqliu, line 61: This (and other err assignments below) could be overwritten by later loops. In the event of reaching max attempts only the last certErr would be returned. Is this the intended behavior? Why not append the errors so it doesn't get overwritten (ex. in the event the important failure point isn't the last certificate attempted)

josephlr: The overwriting here is intentional (and I've updated the code to reflect this). Eventually, we should return multiple errors, but it's probably easier to deal with that when we fix #169.

At the end of the day, it probably doesn't really matter. The vast, vast majority of certs will either have 0 or 1 URLs in `Cert.IssuingCertificateURL`, so it doesn't really matter.

PR #171, opened 2022-03-09
Change FetchCertChain bool option to HTTP client

jessieqliu:
Instead of a boolean flag, have an http.Client specified for fetching AK Certificate chains.

### DIFF
--- a/client/attest.go
+++ b/client/attest.go
@@ -35,7 +35,7 @@ type AttestOpts struct {
 	// If non-nil, will be used to fetch the AK certificate chain for validation.
 	// Key.Attest() will construct the certificate chain by making GET requests to
 	// the contents of Key.cert.IssuingCertificateURL using this client.
-	FetchCertChainClient *http.Client
+	CertChainFetcher *http.Client
 }
 
 // Given a certificate, iterates through its IssuingCertificateURLs and returns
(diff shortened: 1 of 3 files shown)


client/attest.go
```diff
@@ -32,17 +32,17 @@ type AttestOpts struct {
...
 	// logs overlap, server-side verification using this library may fail.
 	CanonicalEventLog []byte
-	// Indicates whether the AK certificate chain should be retrieved for validation.
-	// If true, Key.Attest() will construct the certificate chain by making GET requests to
-	// the contents of Key.cert.IssuingCertificateURL.
-	FetchCertChain bool
+	// If non-nil, will be used to fetch the AK certificate chain for validation.
+	// Key.Attest() will construct the certificate chain by making GET requests to
+	// the contents of Key.cert.IssuingCertificateURL using this client.
+	FetchCertChainClient *http.Client
```
josephlr, line 38: Nit on naming: `CertChainFetcher`? It already has type `Client`, so this name would avoid having the type in the name.  https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#repetitive-with-type

client/attest_test.go
```diff
@@ -62,7 +62,7 @@ func TestFetchIssuingCertificateSucceeds(t *testing.T) {
 
 	leafCert, _ := getTestCert(t, []string{"invalid.URL", ts.URL}, testCA, caKey)
 
-	cert, err := fetchIssuingCertificate(leafCert)
+	cert, err := fetchIssuingCertificate(&http.Client{}, leafCert)
```
josephlr, line 65: In this file we should probably have a Global variable called `localClient` that's just defined as:

```go
// The default client is fine as we are just talking to localhost
var localClient = http.DefaultClient
```

Just to make it clear that we don't need anything special.

client/attest_network_test.go
```diff
@@ -23,7 +24,7 @@ func TestNetworkFetchIssuingCertificate(t *testing.T) {
 
 	key := &Key{cert: akCert}
 
-	certChain, err := key.getCertificateChain()
+	certChain, err := key.getCertificateChain(&http.Client{})
```
josephlr, line 27: In this file, we should have a 

```go
var externalClient = http.DefaultClient
```

We can then define the client to be a different value inside google3

josephlr (approved):

PR #172, opened 2022-03-10
Automatically use attestation intermediates for AK cert verification

jessieqliu:
If an attestation contains intermediate certificates, automatically add them to the options for `checkAKTrusted`

### DIFF
--- a/server/verify.go
+++ b/server/verify.go
@@ -75,7 +75,7 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
 		for _, certBytes := range attestation.IntermediateCerts {
 			cert, err := x509.ParseCertificate(certBytes)
 			if err != nil {
-				return nil, fmt.Errorf("filed to parse intermediate certificate in attestation: %w", err)
+				return nil, fmt.Errorf("failed to parse intermediate certificate in attestation: %w", err)
 			}
 
 			opts.IntermediateCerts.AddCert(cert)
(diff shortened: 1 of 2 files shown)


server/verify.go
```diff
@@ -65,6 +65,22 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
...
+	// Add intermediate certs in the attestation if they exist.
+	if len(attestation.IntermediateCerts) != 0 {
+		if opts.IntermediateCerts == nil {
+			opts.IntermediateCerts = x509.NewCertPool()
+		}
+
+		for _, certBytes := range attestation.IntermediateCerts {
+			cert, err := x509.ParseCertificate(certBytes)
+			if err != nil {
+				return nil, fmt.Errorf("filed to parse intermediate certificate in attestation: %w", err)
```
jessieqliu, line 78: I was unsure if this should raise an error or proceed without adding a certificate (and let any validation issues be caught in `checkAKTrusted`).

server/verify_test.go
```diff
@@ -461,6 +461,22 @@ func TestVerifyAttestationWithCerts(t *testing.T) {
 	}
 }
 
+func TestVerifyAutomaticallyUsesIntermediatesInAttestation(t *testing.T) {
```
alexmwu, line 464: let's also add a test with overlapping `opts.IntermediateCerts` and `att.IntermediateCerts`

server/verify.go
alexmwu: I'm inclined to fail since the client-side already does this check. The proto should not have a malformed intermediate cert. This mirrors the behavior on a malformed AK in `checkAKTrusted`
Note that `ParseCertificate` returns nil on a parsing error. https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/crypto/x509/parser.go;l=989
If we then feed nil into `AddCert`, it will panic. So we need to handle the nil case properly either way (e.g., if we decide to not error out for whatever reason, we can skip)

also typo in the error (failed)

alexmwu (approved):
LGTM after adding the test

PR #182, opened 2022-03-25
Add impersonate service accounts to launcherspec

jkl73:
Create the metadata service field to input service accounts for impersonation.

Operator can leave the field blank, and the default GCE service account will be used;
Or, they can input just one service account which will be the impersonation target;
Or, they can input a list of service accounts, the last service account will be the impersonation target, service accounts before that will act as delegates.

The actual implementation will be in a followup PR

### DIFF
--- a/launcher/spec/launcher_spec.go
+++ b/launcher/spec/launcher_spec.go
@@ -29,11 +29,12 @@ const (
 )
 
 const (
-	imageRefKey        = "tee-image-reference"
-	restartPolicyKey   = "tee-restart-policy"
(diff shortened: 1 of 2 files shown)


jkl73:
> Change LGTM. I wonder though if should separate the impersonation account from the principals (`ImpersonatedServiceAccount` and `Delegates`)

I was thinking about that as well, but then I found https://cloud.google.com/sdk/gcloud/reference#--impersonate-service-account , so I just mimic that usage here.

PR #183, opened 2022-03-31
Add OIDC token refresh to container runner

jessieqliu:
Container runner starts a goroutine that uses a timer to refresh the OIDC token before the previous one expires.


launcher/container_runner_test.go
```diff
@@ -0,0 +1,238 @@
...
+		t.Errorf("expect TTL to be greater than 0, got %v", ttl)
+	}
+
+	if ttl > expectedTTL {
+		t.Errorf("TTL exceeds the expected TTL: got %v, want no greater than %v", ttl, expectedTTL)
+	}
+}
+
+func TestGetTTLError(t *testing.T) {
+	_, err := getTTL([]byte("not a valid token"))
```
alexmwu, line 86: can we test a TTL that's negative?

launcher/container_runner_test.go
```diff
@@ -0,0 +1,238 @@
...
+	expectedRefreshedToken := createJWTToken(t, 10*time.Second)
+	runner.attestAgent = &fakeAttestationAgent{
+		attestFunc: func(context.Context) ([]byte, error) {
+			return expectedRefreshedToken, nil
+		},
+	}
+
+	time.Sleep(ttl)
+
+	data, err = os.ReadFile(filepath)
```
alexmwu, line 230: nice test! would it be possible to test that we change the `attestFunc` but don't sleep until the TTL (then assert the data is the old data, not expectedRefreshedToken like in this test)

alexmwu:
It's looking good! Mostly readability and testing comments

launcher/container_runner.go
```diff
@@ -47,6 +58,12 @@ const (
 	snapshotID  = "tee-snapshot"
 )
 
+// Given a token TTL, calculates the duration to wait before refreshing the token.
+// Interval is shorter than TTL to ensure refresh succeeds before the previous token expires.
+func refreshInterval(ttl time.Duration) time.Duration {
+	return ttl * 9 / 10
```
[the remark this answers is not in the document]
jessieqliu: What should the behavior be if the flag is greater than 1?

launcher/container_runner_test.go
jessieqliu: Can you elaborate on what you mean here? Is it creating a token whose ExpiresAt has already passed?

launcher/container_runner_test.go
jessieqliu: Added a check after changing `attestFunc` but before sleeping, which should cover it.

launcher/container_runner.go
```diff
@@ -194,43 +211,106 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
 	token, err := r.attestAgent.Attest(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to retrieve attestation service token: %v", err)
 	}
 
-	// Write OIDC token for the container.
-	if err := os.MkdirAll(HostTokenPath, 0644); err != nil {
+	filepath := path.Join(HostTokenPath, attestationVerifierTokenFile)
+
+	ttl, err := getTTL(token)
```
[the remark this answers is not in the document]
jessieqliu: Added `refreshToken()`, kept `fetchAndWriteToken` in its own function for ease of testing.

The immediate goroutine seems a little complicated to set up. Don't think it's necessary since there isn't that much duplicated code here anymore.

launcher/container_runner_test.go
alexmwu: Yes, sorry for the confusion. ExpiresAt set to `time.Now() - 1`

launcher/container_runner.go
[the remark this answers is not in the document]
alexmwu: we can return an error or force the value to 11 * ttl`. An error is better, and, if you don't want to pass the error around, you can check early on initialization (add a `validateFlags` func or something)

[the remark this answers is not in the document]
jessieqliu: Did a little refactoring so that refreshToken takes the refresh multiplier and checks that 0 < refreshMultiplier <= 1.

launcher/container_runner.go
```diff
@@ -194,43 +208,112 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+	now := time.Now()
+	if !now.Before(claims.ExpiresAt.Time) {
+		return 0, errors.New("token is expired")
+	}
+
+	return time.Until(claims.ExpiresAt.Time), nil
+}
+
+func (r *ContainerRunner) refreshToken(ctx context.Context, refreshMultiplier float64) (time.Duration, error) {
+	if refreshMultiplier > 1 {
```
alexmwu, line 244: let's store this in the tokenRefresher struct and check it once when setting it.

alexmwu (approved):

go.sum
```diff
@@ -252,6 +252,8 @@ github.com/gogo/protobuf v1.2.1/go.mod h1:hp+jE20tsWTFYpLwKvXlhS1hjn+gTNwPg2I6zV
 github.com/gogo/protobuf v1.3.0/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
 github.com/gogo/protobuf v1.3.1/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
 github.com/gogo/protobuf v1.3.2/go.mod h1:P1XiOD3dCwIKUDQYPy72D8LYyHL2YPYrpS2s69NZV8Q=
+github.com/golang-jwt/jwt v3.2.2+incompatible h1:IfV12K8xAKAnZqdXVzCZ+TOjboZ2keLg81eXfW3O+oY=
```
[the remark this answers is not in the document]
josephlr: Just running `go mod tidy` on the root directory should clean this up.

launcher/go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	github.com/docker/go-events v0.0.0-20190806004212-e31b211e4f1c // indirect
 	github.com/gogo/googleapis v1.4.0 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
+	github.com/golang-jwt/jwt/v4 v4.4.1 // indirect
```
josephlr, line 30: We also should run `go mod tidy` in the `launcher` directory.

launcher/container_runner.go
```diff
@@ -26,12 +29,21 @@ import (
...
-	attestConn  *grpc.ClientConn
-	attestAgent *AttestationAgent
+	container      containerd.Container
+	launchSpec     spec.LauncherSpec
+	attestConn     *grpc.ClientConn
+	attestAgent    attestationAgent
+	tokenRefresher struct {
+		timer *time.Timer
+		done  chan bool
+	}
```
josephlr, line 46: It seems a little odd that the refresher is a member of the `ContainerRunner` struct. It seems like this should only exist inside `Run` method. After the `Run` call completes, the container is stopped, so we shouldn't need to be running a timer anymore.

This also then removes the need to have a separate `done` channel or do additional cleanup in `Close()`.

launcher/container_runner.go
```diff
@@ -194,43 +208,112 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+		for {
+			select {
+			case <-r.tokenRefresher.done:
+				return
+			case <-r.tokenRefresher.timer.C:
+				// Refresh token.
+				duration, err := r.refreshToken(ctx, defaultRefreshMultiplier)
+				if err != nil {
+					log.Printf("failed to refresh attestation service token: %v", err)
+					continue
```
josephlr, line 296: If we aren't going to reset the timer, should this be a `return` to make the behavior more clear?

launcher/container_runner.go
```diff
@@ -194,43 +208,112 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+
+	now := time.Now()
+	if !now.Before(claims.ExpiresAt.Time) {
+		return 0, errors.New("token is expired")
+	}
+
+	return time.Until(claims.ExpiresAt.Time), nil
+}
+
+func (r *ContainerRunner) refreshToken(ctx context.Context, refreshMultiplier float64) (time.Duration, error) {
```
josephlr, line 243: We should add a comment explaining what the return value is. Something like:
```golang
/// Returns how long we should wait until trying to refresh the token again.
```

josephlr (changes requested):

launcher/container_runner.go
```diff
@@ -194,43 +211,114 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
 	return nil
 }
 
-// Run the container
-// Doesn't support container restart yet
-// Container output will always be redirected to stdio for now
-func (r *ContainerRunner) Run(ctx context.Context) error {
-	if err := r.measureContainerClaims(ctx); err != nil {
-		return fmt.Errorf("failed to measure container claims: %v", err)
+func getTTL(token []byte) (time.Duration, error) {
```
josephlr, line 231: I think it might make more sense to just have `getTTL` be `getExpiration` and return the `time.Time` that the token expires. Then all the expiration calculations can be done at the end of `refreshToken`. Something like:

```go
func getExpiration(token []byte) (time.Time, error) {
	claims := &jwt.RegisteredClaims{}
	_, _, err := new(jwt.Parser).ParseUnverified(string(token), claims)
	if err != nil {
		return time.Time{}, fmt.Errorf("failed to parse token: %w", err)
	}
	return claims.ExpiresAt.Time, nil
}

func (r *ContainerRunner) refreshToken(ctx context.Context) (time.Duration, error) {
	...
	exp, err := getExpiration(token)
	if err != nil {
		return 0, fmt.Errorf("failed to get token expiration: %v", err)
	}
	return time.Duration(float64(time.Until(exp)) * defaultRefreshMultiplier), nil
}
```

launcher/container_runner.go
```diff
@@ -194,43 +211,114 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+
+	ttl, err := getTTL(token)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get token TTL: %v", err)
+	}
+
+	return time.Duration(float64(ttl) * r.refresher.refreshMultiplier), nil
+}
+
+func (r *ContainerRunner) fetchAndWriteToken(ctx context.Context, refreshMultiplier float64) error {
```
josephlr, line 265: I think this function can be simplified if we don't store the refresher in the `ContainerRunner`.
```go
func (r *ContainerRunner) fetchAndWriteToken(ctx context.Context) error {
	if err := os.MkdirAll(HostTokenPath, 0744); err != nil {
		return err
	}

	dur, err := r.refreshToken(ctx)
	if err != nil {
		return err
	}
	
	timer := time.NewTimer(dur)
	go func() {
		for {
			select {
			case <-ctx.Done():
				log.Printf("token refreshing stopped: %v", ctx.Err())
				return
			case <-timer.C:
				dur, err := r.refreshToken(ctx)
				if err != nil {
					log.Printf("failed to refresh attestation service token: %v", err)
					return
				}
				timer.Reset(dur)
			}
		}
	}()
	return nil
}
```

We then also shouldn't need a `done` channel as one is already built into the `ctx`.

jessieqliu: It looks like if we use `ctx.Done()`, we have to make sure the context can be cancelled (context.Background() is never cancelled by default so Done() will never be closed in that case).

This can be worked around by using `ctx, cancel := context.WithCancel(context.Background())`, then calling `cancel()` during cleanup, but not sure if this defeats the purpose of removing a dedicated `done` channel in the first place.

josephlr: Good catch, that's correct, we need to cancel the context once the container stops. Using `WithCancel` seems correct here as that will handle cancelling outstanding RPCs as well as cancelling the timer.

Putting something like:
```go
ctx, cancel := context.WithCanel(ctx)
defer cancel()
```
at the start of `Run` should work.

launcher/container_runner.go
jessieqliu: I ended removing the `getExpiration` method since most of the functionality was moved out at that point, and it's only used once in `refreshToken` so it didn't seem necessary to have a separate method.

josephlr (approved):

PR #184, opened 2022-04-01
Retrieve instance info from AK certificate

jessieqliu:
After validating the AK certificate, look for an extension with a specific ObjectIdentifier that contains GCE instance information and return it in a MachineState.

### DIFF
--- a/server/verify.go
+++ b/server/verify.go
@@ -46,22 +46,20 @@ type VerifyOpts struct {
 	IntermediateCerts []*x509.Certificate
 }
 
-type gceSecurityProperties struct {
-	SecurityVersion             int64 `asn1:"optional"`
-	IsProduction                bool  `asn1:"optional"`
-	TpmDataAlwaysEncrypted      bool  `asn1:"optional"`
-	SuspendResumeAlwaysDisabled bool  `asn1:"optional"`
-	VmtdAlwaysDisabled          bool  `asn1:"optional"`
-	AlwaysInYawn                bool  `asn1:"optional"`
-}
-
 type gceInstanceInfo struct {
-	Zone               string `asn1:"utf8"`
-	ProjectNumber      int64
-	ProjectId          string `asn1:"utf8"`
-	InstanceId         int64
-	InstanceName       string                `asn1:"utf8"`
-	SecurityProperties gceSecurityProperties `asn1:"optional"`
+	Zone               []byte
+	ProjectNumber      int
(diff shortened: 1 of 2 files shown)


josephlr (changes requested):

josephlr:
@jessieqliu can you also fix the conflicts?

server/verify.go
```diff
@@ -140,37 +164,67 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
...
+	}
+
+	parsedInstanceInfo := gceInstanceInfo{}
+	if _, err := asn1.Unmarshal(gceInstanceInfoBytes, &parsedInstanceInfo); err != nil {
+		return nil, fmt.Errorf("failed to parse GCE Instance Information Extension: %w", err)
+	}
+
+	// Check production.
+	if !parsedInstanceInfo.SecurityProperties.IsProduction {
+		return nil, nil
```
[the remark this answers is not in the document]
jessieqliu: When we discussed this previously you recommended returning an empty machine state instead of an error. If that doesn't seem fitting anymore it can be changed to an error.

server/verify.go
```diff
@@ -44,6 +47,24 @@ type VerifyOpts struct {
 	IntermediateCerts []*x509.Certificate
 }
 
+type gceSecurityProperties struct {
+	SecurityVersion             int64 `asn1:"optional"`
+	IsProduction                bool  `asn1:"optional"`
```
alexmwu, line 52: This field and the ones below are internal interfaces, should we have them in this library or call them reserved and provide a way to extract them from the library?

jessieqliu: I don't have context on the security implications but in this library we're only interested in IsProduction and perhaps SecurityVersion in the future. I should be able to remove the other fields from the struct (so they won't be unmarshaled) and only use the ones we need?

josephlr: Let's just omit the last 4 fields here (as they are optional anyway, and aren't really something customers can control).

So we would just have:
```go
type gceSecurityProperties struct {
	SecurityVersion             int64 `asn1:"optional"`
	IsProduction                bool  `asn1:"optional"`
}
```

server/verify.go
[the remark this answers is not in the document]
josephlr: Oh that's my bad. You are absolutely correct

josephlr (approved):

PR #189, opened 2022-04-18
Add cloudbuild config and scripts

jkl73:
Add the basci cloudbuild config and related scripts.

The workflow will build the launcher and install it in a COS image using the cos-customizer.
cos-customizer will also setup the entrypoint script for the machine.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,43 @@
+substitutions:
+  '_BASE_IMAGE': 'cos-dev-101-16963-0-0'
+  '_BASE_IMAGE_PROJECT': 'cos-cloud'
+  '_OUTPUT_IMAGE': ''
+  '_ATTEST_ENDPOINT': ''
+
+steps:
+  - name: golang:1.18
+    entrypoint: /bin/bash
+    args:
+      - -c
+      - |
+        cd launcher
+        go build  
+  - name: 'gcr.io/cos-cloud/cos-customizer'
+    args: ['start-image-build',
+           '-gcs-bucket=${PROJECT_ID}_cloudbuild',
+           '-gcs-workdir=customizer-${BUILD_ID}',
+           '-image-name=${_BASE_IMAGE}',
+           '-image-project=${_BASE_IMAGE_PROJECT}']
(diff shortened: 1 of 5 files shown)


launcher/preload.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+copy_launcher() {
+  cp launcher/launcher /usr/share/oem/cc_container_launcher
+}
+
+setup_launcher_systemd_unit() {
+  cp launcher/container-runner.service /usr/share/oem/container-runner.service
+  # set attest service endpoint
+  sed -i 's/\${ATTEST_ENDPOINT}/'${ATTEST_ENDPOINT}'/g' /usr/share/oem/container-runner.service
```
alexmwu, line 10: Can we specify this in a .conf file instead of `sed`ing it? https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines

See more at https://www.freedesktop.org/software/systemd/man/systemd.unit.html (search for .conf files)

alexmwu (approved):

cloudbuild.yaml
```diff
@@ -0,0 +1,42 @@
...
+    args: ['seal-oem']
+  - name: 'gcr.io/cos-cloud/cos-customizer'
+    args: ['run-script',
+           '-script=launcher/fixup_oem.sh']
+  - name: 'gcr.io/cos-cloud/cos-customizer'
+    args: ['finish-image-build',
+           '-oem-size=500M',
+           '-disk-size-gb=11',
+           '-image-name=confidential-space-image-${_OUTPUT_IMAGE}',
+           '-image-project=${PROJECT_ID}',
```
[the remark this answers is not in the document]
jkl73: yes, I was thinking promoting to prod image (creating an image with the prod image family) should use a different workflow, either a different cloudbuild trigger or inside google3

launcher/preload.sh
jkl73: I think it is possible, ATTEST_ENDPOINT is an env variable passed in when executing preload.sh, you mean we create a .conf file on the fly in this function?

PR #190, opened 2022-04-22
Start to use AK cert in attestation

jkl73:
Start to use AK cert in launcher/attestation

Fix ASN1 annotation in `gceInstanceInfo`
Previously it would ignore the optional `gceSecurityProperties` when unmarshaling a real AK cert.

### DIFF
--- a/server/verify.go
+++ b/server/verify.go
@@ -49,8 +49,8 @@ type VerifyOpts struct {
 
 // TODO: Change int64 fields to uint64 when compatible with ASN1 parsing.
 type gceSecurityProperties struct {
-	SecurityVersion int64 `asn1:"optional"`
-	IsProduction    bool  `asn1:"optional"`
+	SecurityVersion int64 `asn1:"explicit,tag:0,optional"`
+	IsProduction    bool  `asn1:"explicit,tag:1,optional"`
 }
 
 type gceInstanceInfo struct {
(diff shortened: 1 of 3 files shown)


server/verify.go
```diff
@@ -59,7 +59,7 @@ type gceInstanceInfo struct {
 	ProjectID          string `asn1:"utf8"`
 	InstanceID         int64
 	InstanceName       string                `asn1:"utf8"`
-	SecurityProperties gceSecurityProperties `asn1:"optional"`
+	SecurityProperties gceSecurityProperties `asn1:"explicit"`
```
josephlr, line 62: Should this be marked explicit and optional?

server/verify.go
```diff
@@ -49,8 +49,8 @@ type VerifyOpts struct {
 
 // TODO: Change int64 fields to uint64 when compatible with ASN1 parsing.
 type gceSecurityProperties struct {
-	SecurityVersion int64 `asn1:"optional"`
-	IsProduction    bool  `asn1:"optional"`
+	SecurityVersion int64 `asn1:"explicit"`
```
josephlr, line 52: Should this have a `tag:0`?

server/verify.go
```diff
@@ -49,8 +49,8 @@ type VerifyOpts struct {
...
+	IsProduction    bool  `asn1:"explicit,tag:1"`
```
josephlr, line 53: Should these also be explicit and optional?

josephlr (approved):
Thanks for fixing this!

josephlr:
@jkl73 I'm not sure what the new cloud-build failure is, but feel free to submit over it, provided that it doesn't matter.

server/verify.go
jkl73: added

server/verify.go
jkl73: added

server/verify.go
jkl73: added

jkl73:
> @jkl73 I'm not sure what the new cloud-build failure is, but feel free to submit over it, provided that it doesn't matter.

That's something I am trying to setup, it can be ignored for now

PR #193, opened 2022-04-27
Add support for fetching impersonated tokens to launcher

jessieqliu:
If the `ImpersonateServiceAccounts` field in the launcher spec is populated with an account (or delegation chain), the `principalFetcher` provided to the attestation verifier will retrieve an impersonated ID token for the provided account/chain.

Tested:
- Manually confirmed that an impersonated ID token can be retrieved for a service account, and verified with a validator from `google.golang.org/api/idtoken`

A manual test to confirm if this does what we want with group impersonators & the WIP is currently in progress, but sending this out for code review to get feedback for now.


launcher/impersonate.go
```diff
@@ -0,0 +1,159 @@
...
+
+		var err error
+		authToken, err = f.fetchAccessToken(sa, authToken, serviceAccounts[i+1:])
+		if err != nil {
+			return "", fmt.Errorf("unable to get access token: %v", err)
+		}
+	}
+
+	fmt.Println("Retrieving impersonated ID token")
+	idToken, err := f.fetchIDToken(idTokenAccount, authToken, audience)
```
jkl73, line 153: instead of fetching the access token one by one manually, is it possible to include SAs as delegates in the request? https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateIdToken#request-body

jkl73 (commented):
Can we use this library https://github.com/googleapis/google-api-go-client/blob/v0.76.0/impersonate/idtoken.go directly?

jessieqliu:
> Can we use this library https://github.com/googleapis/google-api-go-client/blob/v0.76.0/impersonate/idtoken.go directly?

Done. I had a misconception about how the API works, turns out it's a lot simpler than I thought.

jessieqliu:
> Can we use this library https://github.com/googleapis/google-api-go-client/blob/v0.76.0/impersonate/idtoken.go directly?

Done. I had a misconception about how the API works, turns out it's a lot simpler than I thought.

launcher/container_runner.go
```diff
@@ -141,19 +178,32 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 	}
 	// Fetch ID token with specific audience.
 	// See https://cloud.google.com/functions/docs/securing/authenticating#functions-bearer-token-example-go.
```
jkl73, line 180: this comment needs to move to line 193

launcher/container_runner.go
```diff
@@ -25,10 +25,47 @@ import (
...
+}
+
+type idTokenResp struct {
+	Token string `json:"token"`
+}
+
+func fetchImpersonatedToken(ctx context.Context, serviceAccounts []string, audience string, opts ...option.ClientOption) ([]byte, error) {
+	config := impersonate.IDTokenConfig{
+		Audience:        audience,
+		TargetPrincipal: serviceAccounts[len(serviceAccounts)-1],
```
jkl73, line 51: lets check `len(serviceAccounts)` >= 1 at beginning of this function

launcher/container_runner.go
```diff
@@ -141,19 +178,32 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
...
-			}.Encode(),
+	var principalFetcher func(audience string) ([][]byte, error)
+
+	if len(launchSpec.ImpersonateServiceAccounts) != 0 {
+		principalFetcher = func(audience string) ([][]byte, error) {
+			token, err := fetchImpersonatedToken(ctx, launchSpec.ImpersonateServiceAccounts, audience)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get impersonated tokens: %w", err)
+			}
+
```
jkl73, line 189: extra space

jkl73 (commented):
LGTM, could you rebase/sync on the master and fix the CI lint issue?

jessieqliu:
Manual group support test is passing.

launcher/container_runner.go
```diff
@@ -153,7 +176,21 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
...
-		return [][]byte{[]byte(idToken)}, nil
+
+		tokens := [][]byte{[]byte(idToken)}
+
+		if len(launchSpec.ImpersonateServiceAccounts) != 0 {
+			// Fetch impersonated ID tokens.
+			for _, sa := range launchSpec.ImpersonateServiceAccounts {
+				idToken, err := fetchImpersonatedToken(ctx, sa, audience)
+				if err != nil {
+					return nil, fmt.Errorf("failed to get impersonated tokens: %w", err)
```
jkl73, line 187: Maybe helpful for debug to include the sa email in the error.

jkl73 (approved):
lgtm

launcher/container_runner.go
jessieqliu: Joe commented on one of the attestation verifier CLs that the emails could be PII and to avoid printing them out. If it's the same case here, could instead print the index for the email (which is what we do in the attestation verifier).

go.work.sum
```diff
@@ -1 +1,3 @@
 cloud.google.com/go v0.100.2 h1:t9Iw5QH5v4XtlEQaCtUY7x6sCABps8sW0acw7e2WQ6Y=
+github.com/googleapis/gax-go v2.0.2+incompatible h1:silFMLAnr330+NRuag/VjIGF7TLp/LBrV2CJKFLWEww=
+github.com/googleapis/gax-go/v2 v2.1.1 h1:dp3bWCh+PPO1zjRRiCSczJav13sBvG4UhNyVTa1KqdU=
```
josephlr, line 3: I'm not sure why this keeps changing, can you run `go mod tidy` in each directory before submitting this.

launcher/container_runner.go
```diff
@@ -153,7 +176,21 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 		if err != nil {
 			return nil, fmt.Errorf("failed to get principal tokens: %w", err)
 		}
-		return [][]byte{[]byte(idToken)}, nil
+
+		tokens := [][]byte{[]byte(idToken)}
+
+		if len(launchSpec.ImpersonateServiceAccounts) != 0 {
+			// Fetch impersonated ID tokens.
+			for _, sa := range launchSpec.ImpersonateServiceAccounts {
```
josephlr, line 184: We don't need this if and for, just the for statement is fine, and we will just do nothing if the impersonateServiceaccounts slice is empty.

josephlr (approved):
Nice Job

PR #195, opened 2022-05-03
Add launch policy for cmd and env vars

jkl73:
Add the launch policy implementation for env vars and cmd

Update logger so it output everything from launcher to stdout and cloud logging (for better debug)

Update service file so it will sleep for 60 seconds before shutting down the machine.

Update cloudbuild.yaml with the new image name.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -31,8 +31,8 @@ steps:
     args: ['finish-image-build',
            '-oem-size=500M',
            '-disk-size-gb=11',
-           '-image-name=attest-cos-dev-${_OUTPUT_IMAGE}',
-           '-image-family=attest-cos-dev',
+           '-image-name=confidential-space-${_OUTPUT_IMAGE}',
+           '-image-family=confidential-space-dev',
            '-image-project=${PROJECT_ID}',
            '-zone=us-central1-a',
            '-project=${PROJECT_ID}']
--- a/launcher/container-runner.service
+++ b/launcher/container-runner.service
@@ -6,7 +6,8 @@ After=network-online.target gcr-online.target containerd.service
 [Service]
 ExecStart=/var/lib/google/cc_container_launcher --addr=${ATTEST_ENDPOINT}
 # Shutdown the host after the launcher exits
-ExecStopPost=systemctl poweroff
+ExecStopPost=/bin/sleep 60
+ExecStopPost=/usr/bin/systemctl poweroff
 Restart=no
 # RestartSec=90
 StandardOutput=journal+console
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -102,12 +102,20 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 	logger.Printf("Operator Override Env Vars : %v\n", envs)
 	logger.Printf("Operator Override Cmd      : %v\n", launchSpec.Cmd)
 
-	imagelabels, err := getImageLabels(ctx, image)
+	imageLabels, err := getImageLabels(ctx, image)
 	if err != nil {
 		logger.Printf("Failed to get image OCI labels %v\n", err)
-	} else {
-		logger.Printf("Image Labels               : %v\n", imagelabels)
 	}
+
+	logger.Printf("Image Labels               : %v\n", imageLabels)
+	launchPolicy, err := spec.GetLaunchPolicy(imageLabels)
+	if err != nil {
+		return nil, err
+	}
+	if err := launchPolicy.Verify(launchSpec); err != nil {
+		return nil, err
+	}
+
 	if imageConfig, err := image.Config(ctx); err != nil {
 		logger.Println(err)
 	} else {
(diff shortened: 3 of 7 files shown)


cloudbuild.yaml
```diff
@@ -31,8 +31,8 @@ steps:
     args: ['finish-image-build',
            '-oem-size=500M',
            '-disk-size-gb=11',
-           '-image-name=attest-cos-dev-${_OUTPUT_IMAGE}',
-           '-image-family=attest-cos-dev',
```
alexmwu, line 35: Replace "-dev" with "-debug"

launcher/container-runner.service
```diff
@@ -6,7 +6,8 @@ After=network-online.target gcr-online.target containerd.service
 [Service]
 ExecStart=/var/lib/google/cc_container_launcher --addr=${ATTEST_ENDPOINT}
 # Shutdown the host after the launcher exits
-ExecStopPost=systemctl poweroff
+ExecStopPost=/bin/sleep 60
```
alexmwu, line 9: I tried looking through the documentation (https://www.freedesktop.org/software/systemd/man/systemd.service.html) but couldn't find anything about ordering. Does systemd guarantee execution in order or declaration?

launcher/container_runner.go
```diff
@@ -345,10 +345,8 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to fetch and write OIDC token: %v", err)
 	}
 
-	loggerAndStdout := io.MultiWriter(os.Stdout, r.logger.Writer())
-	loggerAndStderr := io.MultiWriter(os.Stderr, r.logger.Writer())
```
alexmwu, line 341: do we no longer want the ability to distinguish b/w stdout/err?

launcher/spec/launch_policy.go
```diff
@@ -0,0 +1,70 @@
...
+		// default is false
+		launchPolicy.AllowedCmdOverride = false
+	}
+
+	return launchPolicy, nil
+}
+
+// Verify will use the LaunchPolicy to verify the given LauncherSpec. If the verification passed, will return nil.
+// If there is any issue, the function will return the error based on the first error.
+func (p LaunchPolicy) Verify(lp LauncherSpec) error {
```
alexmwu, line 50: lp -> ls. I got confused reading thinking lp was launchPolicy

launcher/spec/launch_policy.go
```diff
@@ -0,0 +1,70 @@
...
+		for _, env := range envs {
+			// strip out empty env name
+			if env != "" {
+				launchPolicy.AllowedEnvOverride = append(launchPolicy.AllowedEnvOverride, env)
+			}
+		}
+	}
+
+	if v, ok := imageLabels[cmdOverride]; ok {
+		if launchPolicy.AllowedCmdOverride, err = strconv.ParseBool(v); err != nil {
```
alexmwu, line 37: because ParseBool (https://pkg.go.dev/strconv#ParseBool) accepts so many different bool-like values, we should clearly document that this is the current interface. Alternatively, we can consider only allowing a string "true" and a string "false" and maybe different capitalizations

launcher/spec/launch_policy.go
```diff
@@ -0,0 +1,70 @@
...
+			return LaunchPolicy{}, fmt.Errorf("value of LABEL %s of the image is not a boolean %s", cmdOverride, v)
+		}
+	} else {
+		// default is false
```
alexmwu, line 41: this is the zero value of a boolean already. We shouldn't need this block. If you delete it, the tests should pass.

launcher/spec/launch_policy_test.go
```diff
@@ -0,0 +1,143 @@
+package spec
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestLauncPolicy(t *testing.T) {
```
alexmwu, line 9: nit: "TestLaunchPolicy"

launcher/spec/launch_policy_test.go
```diff
@@ -0,0 +1,143 @@
...
+	testCases := []struct {
+		testName       string
+		imageLables    map[string]string
+		expectedPolicy LaunchPolicy
+	}{
+		{
+			"case 1",
```
alexmwu, line 16: nit: perhaps instead of case 1 we have the test named single Env value.
for case 2, test multiple Env values.

alexmwu (approved):
Looks good. Mostly nits.

launcher/spec/launch_policy_test.go
jkl73: better names, done

launcher/spec/launch_policy_test.go
jkl73: done

launcher/spec/launch_policy.go
jkl73: yep, removed

launcher/spec/launch_policy.go
jkl73: I agree we should emphasis this in the codelab/doc for image authors, recommend using (true/false), but other value are also accept but may not work later.
I just don't want/too lazy to implement our own logic for checking bool val...

PR #196, opened 2022-05-06
Add Cloud Logging; Update the service file for launcher

jkl73:
All launcher log will be sent to Cloud Logging as INFO if running on GCE.

Stdout/Stderr of the container will still be redirected to Stdout/err and won't be sent to Cloud Logging.

Host will poweroff after the launcher exits.

Wrap the `main` function with `run()` so we can call os.exit with non-zero code and execute deferred function.

Update the container io: stdin changed to nil and stdout/stderr will be redirected to logger.


alexmwu (approved):
were you able to test the cloud logging changes?

jkl73:
> were you able to test the cloud logging changes?

Yes, using the image built I can see the log in the Cloud Logging.

jkl73:
@alexmwu I added a new commit that update the IO so the container output can be written to Cloud Log (instead of the serial console), PTAL.

alexmwu:
> PTAL

LGTM, thanks!

PR #198, opened 2022-05-11
Save binary SML at seal time.

Zabari:


Zabari:
It seems that adding a new field to the proto requires a rebuild somewhere, is that something that needs to be done on my side, or somewhere in the build/testing process?

jkl73:
> It seems that adding a new field to the proto requires a rebuild somewhere, is that something that needs to be done on my side, or somewhere in the build/testing process?

Try to run `go generate ./...` in the project dir

jkl73:
Could you elaborate the use case for having the eventlog in the `SealedBytes`? Ideally the proto should contain just enough information for unseal.

An workaround can be calling `GetEventlog()` when you call `Seal()`, and save the eventlog in a separate file.

josephlr:
> A workaround can be calling `GetEventlog()` when you call `Seal()`, and save the eventlog in a separate file.

I think I agree w/ @jkl73 here, given that the EventLog isn't used for Unsealing, it seems weird to include it in the Sealed Data proto, when a user can just include it own their own if they want.

PR #199, opened 2022-05-13
Write the container output to both stdout/err and the logger

jkl73:
Output will appear in stdout as well.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -333,13 +333,14 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 	if err := r.measureContainerClaims(ctx); err != nil {
 		return fmt.Errorf("failed to measure container claims: %v", err)
 	}
-
 	if err := r.fetchAndWriteToken(ctx); err != nil {
 		return fmt.Errorf("failed to fetch and write OIDC token: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


launcher/container_runner.go
```diff
@@ -333,13 +333,14 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
 	}
-
 	if err := r.fetchAndWriteToken(ctx); err != nil {
 		return fmt.Errorf("failed to fetch and write OIDC token: %v", err)
 	}
 
+	loggerAndStdout := io.MultiWriter(r.logger.Writer(), os.Stdout)
+	loggerAndStderr := io.MultiWriter(r.logger.Writer(), os.Stderr)
 	for {
-		task, err := r.container.NewTask(ctx, cio.NewCreator(cio.WithStreams(nil, r.logger.Writer(), r.logger.Writer())))
```
alexmwu, line 342: given that it looks like io.MultiWriter iterates through all the streams in order, let's put stdout first since Cloud Logging makes a network call that could fail.

alexmwu (approved):

launcher/container_runner.go
jkl73: done

PR #203, opened 2022-05-27
Introduce shim verifier client to the launcher

alexmwu:
The shim client will serve as a common abstraction between different attestation verifier implementations. For now, we just wrap the gRPC attestation verifier in the client. Eventually, with other attestation verifiers, we can support multiple protocols more easily as well as create a more robust fake.

This PR also renames `launcher/attest.go` to `launcher/agent.go`


jkl73 (approved):

launcher/internal/verifier/client.go
```diff
@@ -0,0 +1,56 @@
...
+// get challenge part of a remote attestation protocol. The challenge
+// will be verified as part of VerifyAttestation.
+type Challenge struct {
+	name   string
+	nonce  []byte
+	connID string
+}
+
+// Name is the attestation verifier-specific identifier for a challenge.
+func (c *Challenge) Name() string {
```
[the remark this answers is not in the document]
alexmwu: Good point. Done.

launcher/agent.go
```diff
@@ -28,7 +27,7 @@ type principalIDTokenFetcher func(audience string) ([][]byte, error)
 type AttestationAgent struct {
 	tpm              io.ReadWriteCloser
 	akFetcher        tpmKeyFetcher
-	client           servpb.AttestationVerifierClient
+	client           verifier.Client
 	principalFetcher principalIDTokenFetcher
 	cosCel           cel.CEL
 	logger           *log.Logger
```
josephlr, line 33: Are we still using this logger parameter?

josephlr (approved):
Would it be possible to also move all the other GRPC related stuff into `launcher/internal/verifier`, so we can move it out of the way from the core application logic?

launcher/agent.go
alexmwu: yes for one line. I've removed that line because logging can be done in the client and so removed the logger as well.

alexmwu:
> Would it be possible to also move all the other GRPC related stuff into `launcher/internal/verifier`, so we can move it out of the way from the core application logic?

Sure, I moved it into internal/verifier and also regenerated the pbs/upgraded the protobuf lib

PR #204, opened 2022-06-01
Update keys.go documentation

alexmwu:
We should clearly state that the Key object does not support concurrent access except with the `crypto.Signer` from `GetSigner`. This was found because someone asked why *only* the `tpmSigner.Sign` method allows concurrent accesses.


jkl73 (approved):
Maybe we should add a mutex to the Key struct (can be done later)?

alexmwu:
> Maybe we should add a mutex to the Key struct (can be done later)?

@josephlr can comment more but I believe the current mutex is a global is so we can concurrently sign even with different keys. If we wanted to extend synchronization across the whole library, we would do it for all TPM commands, not just per key.

PR #209, opened 2022-06-17
Update VerifyAttestation logic

jkl73:
Update VerifyAttestation to ignore pubkey if cert exists
Update hash validation on the signing hash algorithm in VerifyQuote


client/example_test.go
```diff
@@ -50,9 +50,12 @@ func ExampleKey_Quote() {
 		log.Fatalf("failed to create quote: %v", err)
 	}
 
+	// We don't put restriction on the hash algorithm in this example.
+	noopValidate := func(a tpm2.Algorithm) error { return nil }
```
josephlr, line 54: I'm wondering if it would work better to just have an array of allowed hash algorithms instead of passing a function handler.

josephlr (approved):
This is basically correct, I can clean this up a merge it.

server/verify.go
```diff
@@ -76,14 +75,30 @@ type gceInstanceInfo struct {
...
+	} else {
+		// If AK Cert is presented, ignore the AK pubkey and use the pubkey in the cert.
+		akCert, err = x509.ParseCertificate(attestation.GetAkCert())
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse certificate: %w", err)
+		}
+		var ok bool
+		if akPubKey, ok = akCert.PublicKey.(crypto.PublicKey); !ok {
+			return nil, fmt.Errorf("failed to convert AK public key from the AK cert")
+		}
```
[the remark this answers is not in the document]
jkl73: yep makes sense, you can update this and merge it

josephlr (approved):
@jkl73 @alexmwu I made some changes to how we do the hash algorithm validation.

Basically, this change just rejects all SHA-1 signatures, rather than using the `opts.AllowSHA1` option to make it configurable. We only really need to allow SHA-1 for the _PCRs_, but we shouldn't ever need to accept SHA-1 signatures. TPM2s will always have SHA256 support for signing keys, even if a firmware issue leaves us with only SHA-1 TCG eventlog events.

PTAL and let me know if it looks good to you. I split this up into two separate commits to make review easier.

jessieqliu (approved):

server/verify_test.go
```diff
@@ -571,7 +571,7 @@ func TestVerifyAttestationMissingIntermediates(t *testing.T) {
 }
 
 func TestVerifyMismatchedAKPubAndAKCert(t *testing.T) {
```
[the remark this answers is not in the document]
josephlr: Done

server/verify.go
```diff
@@ -177,8 +189,8 @@ func getInstanceInfo(extensions []pkix.Extension) (*pb.GCEInstanceInfo, error) {
 	}, nil
 }
 
-// Checks if the provided AK public key can be trusted
-func validateAK(ak crypto.PublicKey, akCertBytes []byte, opts VerifyOpts) (*pb.MachineState, error) {
+// Verify the cert or key and return a MachineState.
+func getMachineStateFromCertOrKey(ak crypto.PublicKey, akCert *x509.Certificate, opts VerifyOpts) (*pb.MachineState, error) {
```
alexmwu, line 193: pretty nit: but this function seems to be doing too much. Maybe we should pull the extraction of `MachineState` from *only the cert* into its own func? And then still call it `validateAK`

alexmwu (approved):

jkl73:
LGTM

server/verify.go
josephlr: Fixed in https://github.com/google/go-tpm-tools/pull/209/commits/352f92b26aa155950fd4b134e4091401ebe6afad to just have `validateAKPub` and `validateAKCert` functions. They are then conditionally called in `VerifyAttestation`

PR #214, opened 2022-06-28
Move verifier from internal and grpc to subpackage

alexmwu:
We now export verifier, but make no stability guarantees.
Additionally, we move grpc_client.go and other grpc-related files to the
grpc/ subpackage to better couple all verifier implementation details
with the protocol.


josephlr (approved):

josephlr (commented):
Could we rename `GRPCClient` to `Client` and the below method to `NewClient`. It should make the linter happy.

alexmwu:
> Could we rename `GRPCClient` to `Client` and the below method to `NewClient`. It should make the linter happy.

Done. Just waiting for a pass to merge.

PR #216, opened 2022-07-01
Add REST-based verifier.Client

josephlr:
This also adds a (very simple) networking test that I confirmed worked
on my local workstation. Note that this test will need to be changed in
the future, as the prod service will eventually stop having the fake
implementation.

Note that this test is also just skipped if there is no was to create a
default-authenticated client.

More tests should be added in a follow-up PR.


launcher/verifier/rest/rest.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return &verifier.Challenge{
+		Name:  output.Name,
+		Nonce: nonce,
+	}, nil
+}
+
+// VerifyAttestation implements verifier.Client
+func (c *restClient) VerifyAttestation(ctx context.Context, request verifier.VerifyAttestationRequest) (*verifier.VerifyAttestationResponse, error) {
```
alexmwu, line 65: given that the fields in `VerifyAttestationRequest` are exported, we should probably add a check that Challenge isn't nil

alexmwu: and `Attestation` since `convertRequestToREST` assumes non-nil

josephlr: I changed the code to add a check, and also added tests that our conversion methods passed.

josephlr:
@alexmwu @jkl73 I made some additional changes in this PR (which are split into separate commits).

First, I moved the `verifier/grpc` package to `verifier/grpcclient` to prevent conflicts with the existing `grpc` package. This commit (https://github.com/google/go-tpm-tools/pull/216/commits/6da98dfefa36bba1becd2aab2df9c5692f9268f4) also makes the `grpcClient` structure private, as it's fine to just return a `verifier.Client`.

Second, I added the `verifier/rest` package which uses the REST interface. Similar to the GRPC interface, it has a private `restClient` type and returns a `verifier.Client`.

Finally, I added the nil checks Alex wanted.

jkl73 (approved):

PR #219, opened 2022-07-08
Update ContainerRunner to use REST verifier

alexmwu:
We update the LauncherSpec to store the projectID and region.

The ContainerRunner will now be able to use the REST-based verifier.Client, based on the addr flag.
The flag will determine whether the endpoint is the gRPC or REST verifier. The REST verifier is used if the flag is empty or contains "http", otherwise the gRPC verifier is used.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"strings"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
@@ -23,10 +24,13 @@ import (
 	"github.com/google/go-tpm-tools/client"
 	"github.com/google/go-tpm-tools/launcher/agent"
 	"github.com/google/go-tpm-tools/launcher/spec"
+	"github.com/google/go-tpm-tools/launcher/verifier"
 	"github.com/google/go-tpm-tools/launcher/verifier/grpcclient"
+	"github.com/google/go-tpm-tools/launcher/verifier/rest"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
 	"google.golang.org/api/impersonate"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


launcher/container_runner.go
```diff
@@ -207,6 +218,43 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 	}, nil
 }
 
+// getGRPCClient returns a gRPC verifier.Client pointing to the given address.
+// It also returns a grpc.ClientConn for closing out the connection.
+func getGRPCClient(asAddr string, logger *log.Logger) (verifier.Client, *grpc.ClientConn, error) {
+	opt := grpc.WithTransportCredentials(insecure.NewCredentials())
+	conn, err := grpc.Dial(asAddr, opt)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to open connection to attestation service: %v", err)
```
josephlr, line 227: We should mention GRPC in this error message

launcher/container_runner.go
```diff
@@ -207,6 +218,43 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
...
+		return nil, fmt.Errorf("failed to create HTTP client: %v", err)
+	}
+
+	var opts []option.ClientOption
+	opts = append(opts, option.WithHTTPClient(httpClient))
+	if asAddr != "" {
+		opts = append(opts, option.WithEndpoint(asAddr))
+	}
+
+	restClient, err := rest.NewClient(context.Background(),
```
josephlr, line 247: feed in `ctx` in stead of `context.Background`

launcher/container_runner.go
```diff
@@ -207,6 +218,43 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
...
+		spec.ProjectID,
+		spec.Region,
```
josephlr, line 249: For now, should we just pass `us-central1` as the region? If we do it's fine to keep the `Region` stuff in the launcher spec.

launcher/container_runner.go
```diff
@@ -207,6 +218,43 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
...
+// It defaults to the Attestation Verifier instance at
+// https://confidentialcomputing.googleapis.com.
+func getRESTClient(ctx context.Context, asAddr string, spec spec.LauncherSpec) (verifier.Client, error) {
+	httpClient, err := google.DefaultClient(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create HTTP client: %v", err)
+	}
+
+	var opts []option.ClientOption
+	opts = append(opts, option.WithHTTPClient(httpClient))
```
josephlr, line 242: Nit, this can just be:

```go
opts := []option.ClientOption{option.WithHTTPClient(httpClient)}
```

josephlr (approved):

launcher/container_runner.go
alexmwu: oh sorry you're saying still keep Region in launcher spec because we'll eventually need it. Brainfart on my end.

josephlr: Right now the REST service is only running in us-central1. Ideally, if we were running this in every region, we would do what your code does, call the attestation verifier endpoint corresponding the Region in which the VM is running.

However, right now, we can only call the service in us-central1. To support the ability to run VM in other regions, we can just have those VMs make a cross-region API call to `us-central1`.

josephlr: Basically, for now  we just do:
```go
const defaultRegion = "us-central1"
restClient, err := rest.NewClient(ctx, spec.ProjectID, defaultRegion, opts...)
```
but we keep the code that fills out `spec.Region`.

alexmwu: Done. Thanks!

PR #220, opened 2022-07-09
Update launcher flags and launcher_spec

jkl73:
Remove useLocalImage and serverAddr flags
Add defaultAttestationServiceEndpoint in launcher_spec, and allow
the endpoint to be configured from MDS.
Update launcher service file to support prod and debug images

### DIFF
--- a/launcher/preload.sh
+++ b/launcher/preload.sh
@@ -7,7 +7,15 @@ copy_launcher() {
 setup_launcher_systemd_unit() {
   cp launcher/container-runner.service /usr/share/oem/container-runner.service
   # set attest service endpoint
-  sed -i 's/\${ATTEST_ENDPOINT}/'${ATTEST_ENDPOINT}'/g' /usr/share/oem/container-runner.service
+
+  if [ "$IMAGE_ENV" == "hardened" ]; then
+    cp launcher/hardened.conf /usr/share/oem/launcher.conf
+  elif [ "$IMAGE_ENV" == "debug" ]; then
+    cp launcher/debug.conf /usr/share/oem/launcher.conf
+  else
+    echo "Unknown IMAGE_ENV, use hardened or debug"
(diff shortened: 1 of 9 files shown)


launcher/preload.sh
```diff
@@ -7,7 +7,16 @@ copy_launcher() {
 setup_launcher_systemd_unit() {
   cp launcher/container-runner.service /usr/share/oem/container-runner.service
   # set attest service endpoint
-  sed -i 's/\${ATTEST_ENDPOINT}/'${ATTEST_ENDPOINT}'/g' /usr/share/oem/container-runner.service
+
+  if [ "$IMAGE_ENV" == "prod" ]; then
```
alexmwu, line 11: let's go with "hardened", as that is more descriptive than prod. The prod image is relative to the user but from our POV both images could be prod images.

launcher/preload.sh
```diff
@@ -7,7 +7,16 @@ copy_launcher() {
...
+    cp launcher/prod.conf /usr/share/oem/launcher.conf
+  elif [ "$IMAGE_ENV" == "debug" ]; then
+    cp launcher/debug.conf /usr/share/oem/launcher.conf
+  else
+    echo "Unknown IMAGE_ENV, use prod or debug"
+    exit 1
+  fi
+  # sed -i 's/\${ATTEST_ENDPOINT}/'${ATTEST_ENDPOINT}'/g' /usr/share/oem/container-runner.service
```
alexmwu, line 19: nit: remove

launcher/prod.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+# prod image should exit after workflow finished
+ExecStopPost=
```
alexmwu, line 3: why do we need this line?

alexmwu (approved):

PR #222, opened 2022-07-18
Print OIDC token payload

jkl73:
Print the JSON payload for the OIDC token for the better
debuggability regarding the WIP policy.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -349,6 +349,18 @@ func (r *ContainerRunner) refreshToken(ctx context.Context) (time.Duration, erro
 		return 0, fmt.Errorf("failed to write token to container mount source point: %v", err)
 	}
 
+	// Print out the claims in the jwt payload
+	mapClaims := jwt.MapClaims{}
+	_, _, err = jwt.NewParser().ParseUnverified(string(token), mapClaims)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse token: %w", err)
+	}
+	claimsString, err := json.MarshalIndent(mapClaims, "", "  ")
+	if err != nil {
+		return 0, fmt.Errorf("failed to format claims: %w", err)
+	}
+	r.logger.Println(string(claimsString))
+
 	return time.Duration(float64(time.Until(claims.ExpiresAt.Time)) * defaultRefreshMultiplier), nil
 }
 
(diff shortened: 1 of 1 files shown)


launcher/container_runner.go
```diff
@@ -334,11 +336,22 @@ func (r *ContainerRunner) refreshToken(ctx context.Context) (time.Duration, erro
 
 	// Get token expiration.
 	claims := &jwt.RegisteredClaims{}
-	_, _, err = jwt.NewParser().ParseUnverified(string(token), claims)
+	_, parts, err := jwt.NewParser().ParseUnverified(string(token), claims)
```
josephlr, line 339: It seems like we should just be passing in a different claims structure ('like [`MapClaims`](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#MapClaims)) rather than manually decoding the second segment ourselves.

launcher/container_runner.go
```diff
@@ -334,11 +336,22 @@ func (r *ContainerRunner) refreshToken(ctx context.Context) (time.Duration, erro
...
 
+	// Print out the claims in the jwt payload
+	rawDecodedPayload, err := base64.RawStdEncoding.DecodeString(string(parts[1]))
+	if err != nil {
+		return 0, err
+	}
+	var jsonBuffer bytes.Buffer
+	if err := json.Indent(&jsonBuffer, []byte(rawDecodedPayload), "", "  "); err != nil {
+		return 0, err
+	}
```
josephlr, line 352: Then we could remove this code an just do:
```
claimsString, err := json.MarshalIndent(claims, "", "  ")
r.logger.Println(claimsString)
```

josephlr (commented):
Overall I like this approch, but I think we can have the JWT library do the parsing rather than trying to manually do it ourselves

launcher/container_runner.go
jkl73: Done

launcher/container_runner.go
jkl73: Done

josephlr (approved):
Parsing the token twice isn't idea, but it should be fine. We can cleanup this later when we introduce a custom claims structure.

josephlr:
Lint failure is https://github.com/golangci/golangci-lint/issues/2997

I think we can just change our CI configuration to work around it.

josephlr:
I ran the Lints locally, no issues, we will see if the CI issues persist.

PR #225, opened 2022-07-18
Keys.go direct support

matt-tsai:
Added some attributes to the key structure for tpmDirect support along with related initialization. Fixed code style and cleaned up some comments.

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -12,6 +12,10 @@ import (
 
 	"github.com/google/go-tpm-tools/internal"
 	pb "github.com/google/go-tpm-tools/proto/tpm"
+	"github.com/google/go-tpm/direct/structures/tpm2b"
+	"github.com/google/go-tpm/direct/structures/tpmt"
+	tpm2direct "github.com/google/go-tpm/direct/tpm2"
+	"github.com/google/go-tpm/direct/transport"
 	"github.com/google/go-tpm/tpm2"
 	"github.com/google/go-tpm/tpmutil"
 )
@@ -25,6 +29,12 @@ type Key struct {
 	rw      io.ReadWriter
 	handle  tpmutil.Handle
 	pubArea tpm2.Public
+	// >>> Start Direct Implementaion <<<
+	transportTPM  transport.TPM
+	pubAreaDirect tpmt.Public
+	nameDirect    *tpm2b.Name
+	sessionDirect tpm2direct.Session
+	// >>> End Direct Implementaion <<<
 	pubKey  crypto.PublicKey
 	name    tpm2.Name
 	session session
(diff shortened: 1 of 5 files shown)


client/keys.go
```diff
@@ -25,6 +29,12 @@ type Key struct {
 	rw      io.ReadWriter
 	handle  tpmutil.Handle
 	pubArea tpm2.Public
+	// >>> Start Direct Implementaion <<<
+	transportTPM  transport.TPM
+	pubAreaDirect tpmt.Public
+	nameDirect    *tpm2b.Name
```
alexmwu, line 39: Rather than duplicating these fields, could we have funcs that do the conversions between legacy and direct?

There's no getting around the different session, but I don't think we should duplicate the data for the TPM structs or the TPM interface (transport.TPM). That could lead to data inconsistency issues.

client/keys.go
```diff
@@ -240,6 +282,12 @@ func (k *Key) PublicArea() tpm2.Public {
 	return k.pubArea
 }
 
+// PublicAreaDirect exposes the key's entire direct public area. This is useful for
+// determining additional properties of the underlying TPM key.
+func (k *Key) PublicAreaDirect() tpmt.Public {
```
alexmwu, line 287: let's unexport this. Otherwise, when we finish the migration, we would have to break uses of `Key.PublicArea` (signature change) as well as `Key.PublicAreaDirect` (removing altogether). Although the first is a required breaking change, we should generally prefer fewer breaking changes.

client/keys.go
matt-tsai: After discussing with Joe yesterday, we concluded that back compatibility would not be possible thus why I did not choose to have conversion between legacy and direct. The reason why it is duplicated so that it will be easier to deprecate old legacy code later on.

client/keys.go
```diff
@@ -207,14 +244,19 @@ func (k *Key) finish() error {
 	if k.name, err = k.pubArea.Name(); err != nil {
 		return err
 	}
+	if k.nameDirect, err = tpm2direct.ObjectName(&k.pubAreaDirect); err != nil {
+		return err
+	}
 	// We determine the right type of session based on the auth policy
 	if k.session == nil {
 		if bytes.Equal(k.pubArea.AuthPolicy, defaultEKAuthPolicy()) {
+			// Check for direct session here when sessions are implemented
```
alexmwu, line 253: nit: prefix with "TODO: "

client/keys.go
alexmwu: Chatted offline and decided to have the immutable fields (name/pubArea) duplicated but stored in `finish()`. Then we will expose the interface to the TPM with a private `Key.transport()` method so that we can migrate over to the new `transport.TPM` type

josephlr (approved):

PR #226, opened 2022-07-19
Update CI to build against 1.17 and 1.18, tpmDirect does not support 1.16.

matt-tsai:


.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
   build_old:
     strategy:
       matrix:
-        go-version: [1.16.x, 1.17.x]
```
alexmwu, line 31: Updating the golang version is done in the go.mod. This just changes which go versions we build against.

alexmwu (commented):
Per the go.mod, we are already running go 1.17. What problem are you encountering?

.github/workflows/ci.yml
alexmwu: ah I see, then maybe change this PR name to "Update CI to build against 1.17 and 1.18"

alexmwu (approved):

PR #235, opened 2022-08-10
Implemented Session Encryption

matt-tsai:
Implemented Session Encryption for Unseal.

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -385,7 +385,7 @@ func (k *Key) sealDirect(sensitive []byte, opts sealOptsDirect) (*pb.SealedBytes
 	var pcrs *pb.PCRs
 	var err error
 	var auth []byte
-
+	
 	pcrs, err = mergePCRSelAndProtoDirect(k.transportTPM(), opts.Current, opts.Target)
 	if err != nil {
 		return nil, fmt.Errorf("invalid SealOpts: %v", err)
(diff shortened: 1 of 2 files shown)


client/keys.go
```diff
@@ -688,7 +688,11 @@ func (k *Key) unsealDirect(in *pb.SealedBytes, opts unsealOptsDirect) ([]byte, e
 		return nil, fmt.Errorf("failed to create PCRSelection: %v", err)
 	}
 
-	sess, cleanup2, err := tpm2direct.PolicySession(k.transportTPM(), sessionHashAlgTpmDirect, SessionHashAlg.Size())
+	var options []tpm2direct.AuthOption
+	options = append(options, tpm2direct.Salted(tpm.Handle(k.handle), k.pubAreaDirect))
```
alexmwu, line 692: should we check the key is a decrypt key? we should probably double check it's asymmetric too, otherwise the user needs to create a sym key with a sensitive part that can be sniffed on the bus

alexmwu: I also wonder if we should check the type of the key is RSA or ECC? otherwise we get some error of the form "failed to start Policy Session: salt encryption alg '%v' not supported", which is not very descriptive. the problem is we duplicate the supported types in tools as well as direct, but I'm inclined to give better error messages a higher priority.

alexmwu (changes requested):
Please rebase so that there's only the session encryption commit/this can be merged in

alexmwu:
adding Chris to double check we used the direct API properly.

client/keys.go
matt-tsai: Resolved the decrypt check.

alexmwu (approved):

client/keys.go
```diff
@@ -385,7 +385,9 @@ func (k *Key) sealDirect(sensitive []byte, opts sealOptsDirect) (*pb.SealedBytes
 	var pcrs *pb.PCRs
 	var err error
 	var auth []byte
-
+	if !k.pubAreaDirect.ObjectAttributes.Decrypt {
```
chrisfenner, line 388: You could also check that it's a restricted key, if you want.

chrisfenner (approved):
Looks exactly right to my eyes! Only one minor comment about checks we might make.

client/keys.go
matt-tsai: Check added, resolved.

PR #240, opened 2022-08-22
Add an SEV-SNP attestation if possible

deeglaze:
When getting a vTPM quote within an SEV-SNP VM, attestation verifiers
will also want to know if the SEV-SNP technology claim can be verified
with a hardware-rooted attestation.

We add "VerifyTechnology" for the GCENonHostInfo event that carries
which memory encryption technology the guest is using. SEV and SEV-ES do
not have attestation support in GCE, but SEV-SNP will.

The addition of the go-sev-guest dependency increases the minimum Go
version to 1.19.

Signed-off-by: Dionna Glaze <dionnaglaze@google.com>


deeglaze:
Don't accept this until go-sev-guest passes hardware tests. It only passes fakes at the moment. I'm fixing currently.

deeglaze:
go-sev-guest passes hardware tests for v0.2.2, so this PR is ready for review.

server/verify.go
```diff
@@ -293,3 +302,29 @@ func makePool(certs []*x509.Certificate) *x509.CertPool {
 	}
 	return pool
 }
+
+// VerifyTechnology checks Trusted Execution Technology (TEE) attestation reports if the technology
+// supports them, and only then validates that a particular technology has proven that it is in use.
+func VerifyTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTechnology, opts *VerifyOpts) error {
```
alexmwu, line 308: Let's at least specify GCE, since it relies on the value in the GCE-written firmware event log

client/attest.go
```diff
@@ -36,6 +37,11 @@ type AttestOpts struct {
 	// Key.Attest() will construct the certificate chain by making GET requests to
 	// the contents of Key.cert.IssuingCertificateURL using this client.
 	CertChainFetcher *http.Client
+	// SevSnpDevice implements the go-sev-guest client.Device interface for
+	// collecting an SEV-SNP attestation, and is expected to be open. Attest
+	// will not close this device. If nil, then attestation will use
```
alexmwu, line 42: I wonder if this name and even interface should be so specific. We only use one CC hardware device at a time, so long term we could have a generic interface that every CC device type implements. Then each one could provide technology attestation-specific inputs e.g., technology type, different sized nonces based on requirements/best practices, different event log formats, etc.

client/attest.go
```diff
@@ -109,6 +115,30 @@ func (k *Key) getCertificateChain(client *http.Client) ([][]byte, error) {
...
+func getSevSnpAttestationReport(attestation *pb.Attestation, opts AttestOpts) {
+	var err error
+	device := opts.SevSnpDevice
+	if device == nil {
+		device, err = sg.OpenDevice()
+		if err != nil { // No access or not an SEV-SNP guest.
+			return
+		}
+	}
+	var snpNonce [64]byte
```
alexmwu, line 128: TPM nonces are limited to the max size of the TPM2B_Name. So this is ~34 bytes since all TPM2s need to support SHA256. This *can* be bigger, but not necessarily. If this size doesn't match SNP expectations/requirements, we should consider getting a nonce from another source

server/verify.go
```diff
@@ -293,3 +302,29 @@ func makePool(certs []*x509.Certificate) *x509.CertPool {
...
+	switch tech {
+	case pb.GCEConfidentialTechnology_NONE: // Nothing to verify
+		return nil
+	case pb.GCEConfidentialTechnology_AMD_SEV: // Not verifiable on GCE
+		return nil
+	case pb.GCEConfidentialTechnology_AMD_SEV_ES: // Not verifiable on GCE
+		return nil
+	case pb.GCEConfidentialTechnology_AMD_SEV_SNP:
+		switch tee := attestation.GetTeeAttestation().(type) {
+		case *pb.Attestation_SevSnpAttestation:
```
alexmwu, line 318: Will there be other possible types for `GCEConfidentialTechnology_AMD_SEV_SNP`? If not, let's just cast and check the result:

```
snpAttest, ok := attestation.GetTeeAttestation.(*pb.Attestation_SevSnpAttestation)
if !ok {
  return fmt.Errorf("TEE attestation is %v, expected an SevSnpAttestation", attestation.GetTeeAttestation())
}
```

deeglaze: There will be other technologies, so I could change this into a switch on the type.

server/verify.go
deeglaze: Done.

proto/README.md
```diff
@@ -0,0 +1,16 @@
+# Generating *.pb.go files
+
+To generate attest.pb.go, run
+
+```
+protoc -I. -I`go list -m -f "{{.Dir}}" github.com/google/go-sev-guest` --go_out=. attest.proto
```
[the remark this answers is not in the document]
deeglaze: Done.

client/attest.go
deeglaze: Done.

client/attest.go
deeglaze: Done.

deeglaze:
I don't know why the macos-latest build is failing. There are no error messages, and it builds on my macos machine.

client/attest.go
```diff
@@ -109,6 +129,76 @@ func (k *Key) getCertificateChain(client *http.Client) ([][]byte, error) {
...
+	device := opts.TEEDevice
+	if device != nil {
+		return device.AddAttestation(attestation, opts)
+	}
+
+	// Try SEV-SNP.
+	if device, err := CreateSevSnpDevice(); err == nil {
+		// Don't return errors if the attestation collection fails, since
+		// the user didn't specify a TEEDevice.
+		device.AddAttestation(attestation, opts)
```
alexmwu, line 193: I wonder if we should log this at some point, otherwise debugging is impossible. The default go log package logs to stderr, which seems like a reasonable default. I'd like to avoid having users pass in a Logger opt. Thoughts?

server/policy_constants.go
```diff
@@ -161,7 +161,7 @@ func ParseGCENonHostInfo(nonHostInfo []byte) (pb.GCEConfidentialTechnology, erro
 		return pb.GCEConfidentialTechnology_NONE, errors.New("prefix for GCE Non-Host info is missing")
 	}
 	tech := nonHostInfo[prefixLen]
-	if tech > byte(pb.GCEConfidentialTechnology_AMD_SEV_ES) {
+	if tech > byte(pb.GCEConfidentialTechnology_AMD_SEV_SNP) || tech == byte(3) {
```
alexmwu, line 164: have we already reserved 3 in the proto for another technology? If so, consider commenting that it's reserved in the `GCEConfidentialTechnology` enum in attest.proto

server/verify.go
alexmwu: I understand there will be different `pb.GCEConfidentialTechnology`. I'm wondering if there will be different Attestation types for SevSnp. If not, the above looks a bit cleaner.

Same question for whether there will be different types of Opts for just `pb.Attestation_SevSnpAttestation`. If not let's change it to:
```
snpOpts, ok := opts.TEEOpts.(*VerifySnpOpts)
if !ok {
  return fmt.Errorf("unexpected value for TEEOpts given a SEV-SNP attestation report: %v", opts.TEEOpts)
}
```

alexmwu (commented):
Can you run `go mod tidy` in your workspace? The ubuntu build is failing.

proto/gen_attest.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+protoc -I. -I`go list -m -f "{{.Dir}}" github.com/google/go-sev-guest` --go_out=. --go_opt=module=github.com/google/go-tpm-tools/proto attest.proto
```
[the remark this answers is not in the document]
deeglaze: Yes, you can't backtick in go:generate to call `go list` to get the appropriate location for go-sev-guest.

client/attest.go
deeglaze: If you'd like "log", I can add it. I just didn't want to make log noise for folks not expecting the client to look for confidential compute capabilities.

server/policy_constants.go
deeglaze: 3 is reserved internally and will be used eventually. Will add a comment.

client/attest.go
```diff
@@ -109,6 +129,76 @@ func (k *Key) getCertificateChain(client *http.Client) ([][]byte, error) {
...
+}
+
+// AddAttestation will get the SEV-SNP attestation report given opts.TEENonce with
+// associated certificates and add them to `attestation`. If opts.TEENonce is empty,
+// then uses contents of opts.Nonce.
+func (d *SevSnpDevice) AddAttestation(attestation *pb.Attestation, opts AttestOpts) error {
+	var snpNonce [64]byte
+	if len(opts.TEENonce) == 0 {
+		copy(snpNonce[:], opts.Nonce[:])
+	} else if len(opts.TEENonce) != 64 {
```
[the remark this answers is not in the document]
deeglaze: This function is associated with an open SEV-SNP device, so we will only expect SEV-SNP shaped inputs. We assume the user knows what they're doing when supplying TEENonce without supplying TEEDevice. They expect the attestation to be collected only in the environment that needs the nonce to be the size given.

If that's too weakly typed for you, would you prefer that a non-nil TEENonce with a nil TEEDevice constitutes an error?

client/attest.go
alexmwu: No, you're right. I think if we end up having issues with this, we can add a Logger opt that when nil defaults to not logging. Thanks!

server/verify_test.go
```diff
@@ -816,3 +822,136 @@ func TestGetInstanceInfoASN(t *testing.T) {
 		t.Errorf("getInstanceInfo did not return expected instance info: got %v, want %v", instanceInfo, expectedInstanceInfo)
 	}
 }
+
+func TestVerifyAttestationWithSevSnp(t *testing.T) {
```
alexmwu, line 826: Could you add a few failure mode tests? For example, using bad roots, incorrect nonce, allow debug false.

client/attest.go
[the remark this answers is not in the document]
alexmwu: Okay, so reading between the lines, for the user of `Attest` there will be some way of detecting the technology type and then requesting a nonce of the appropriate size from an attestation service?

If a user of this library should always knows what kind of TEE they are running, I think they might as well pass in the TEEDevice as well (it looks like you already made this change).

server/verify_test.go
alexmwu, line 826: Could you add a few failure mode tests? For example, using bad roots, incorrect nonce, allow debug false.

client/attest.go
```diff
@@ -36,6 +48,16 @@ type AttestOpts struct {
 	// Key.Attest() will construct the certificate chain by making GET requests to
 	// the contents of Key.cert.IssuingCertificateURL using this client.
 	CertChainFetcher *http.Client
+	// TEEDevice implements the TEEDevice interface for collecting a Trusted execution
+	// environment attestation. If nil, then Attest will try all known TEE devices,
+	// and TEENonce must be nil. If not nil, Attest will not call Close() on the device.
+	TEEDevice TEEDevice
```
alexmwu, line 54: could you add a quick test that `Attest` fails when TEEDevice and TEENonce are xor set?

server/verify_test.go
deeglaze: Added.

client/attest.go
deeglaze: Added.

deeglaze:
> Can you run `go mod tidy` in your workspace? The ubuntu build is failing.

I ran it and found no changes from what's in this PR already. I've added a commit before the main one to remove build_old from the GitHub ci.yml file since this PR makes the project require 1.19

alexmwu (approved):

server/verify_sev.go
```diff
@@ -0,0 +1,35 @@
...
+	sabi "github.com/google/go-sev-guest/abi"
+	spb "github.com/google/go-sev-guest/proto/sevsnp"
+	"github.com/google/go-sev-guest/validate"
+	sv "github.com/google/go-sev-guest/verify"
+)
+
+// VerifySnpOpts allows for customizing the functionality of VerifyAttestation's SEV-SNP verification.
+type VerifySnpOpts struct {
+	// ReportData is the expected REPORT_DATA field of the attestation report.
+	ReportData [64]byte
```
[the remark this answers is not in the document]
deeglaze: Done.

PR #241, opened 2022-08-30
Update CI Go Version to 1.19 (and other formatting nits)

josephlr:
By upgrading our default GI version to Go 1.19, we now catch more lints/nits than we did before. To make the linter happy we:
  - Remove uses of the `ioutil` library
  - Add a package comment for the `cmd/main` package
  - Fix comment formatting in `attest.proto` (which also causes the generated `.pb.go` to change whitespace).
  - Run gofmt over the entire tree (which makes some changes, but only whitespace changes).

This reduces the patches we need when importing into google3


cmd/gotpm/main.go
```diff
@@ -1,3 +1,4 @@
+// Package main is a just a binary wrapper package around cmd.
```
JoshuaKrstic, line 1: nit: Consider removing the word "just"

josephlr: FIxed

.github/workflows/ci.yml
```diff
@@ -111,8 +111,7 @@ jobs:
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v3.2.0
         with:
-          # Update when https://github.com/golangci/golangci-lint/issues/2997 is fixed.
```
alexmwu, line 114: does `version: latest` not work anymore? I'd prefer to do that if the regression is fixed than run with an older version

.github/workflows/ci.yml
```diff
@@ -125,11 +124,12 @@ jobs:
             --exclude-use-default=false
             --max-same-issues=0
             --max-issues-per-linter=0
```
alexmwu, line 126: nit: `--timeout 2m`

alexmwu (approved):

.github/workflows/ci.yml
josephlr: Fixed

.github/workflows/ci.yml
josephlr: Fixed, using version: latest seems to work.

PR #244, opened 2022-09-09
Add IsHardened in launch spec

jkl73:
~~* COS eventlog now can measure to a resettable PCR (in the debug image)~~
~~* Debug PCR will reset every time before launcher run in the debug image~~
* Added a kernel cmdline arg confidential-space.hardened=true to indicate hardened image
* Change LauncherSpec to LaunchSpec
* Add cgo flag to silence deprecate warning in openssl for ms-tpm-20-ref

### DIFF
--- a/launcher/spec/launch_spec.go
+++ b/launcher/spec/launch_spec.go
@@ -1,10 +1,11 @@
-// Package spec contains definition of some basic container launcher specs needed to
+// Package spec contains definition of some basic container launch specs needed to
 // launch a container, provided by the operator.
 package spec
 
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 
 	"cloud.google.com/go/compute/metadata"
@@ -49,9 +50,9 @@ type EnvVar struct {
 	Value string
 }
 
-// LauncherSpec contains specification set by the operator who wants to
+// LaunchSpec contains specification set by the operator who wants to
 // launch a container.
-type LauncherSpec struct {
+type LaunchSpec struct {
 	// MDS-based values.
 	ImageRef                   string
 	RestartPolicy              RestartPolicy
@@ -61,11 +62,12 @@ type LauncherSpec struct {
 	ImpersonateServiceAccounts []string
 	ProjectID                  string
 	Region                     string
+	Hardened                   bool
 }
 
 // UnmarshalJSON unmarshals an instance attributes list in JSON format from the metadata
-// server set by an operator to a LauncherSpec.
-func (s *LauncherSpec) UnmarshalJSON(b []byte) error {
+// server set by an operator to a LaunchSpec.
+func (s *LaunchSpec) UnmarshalJSON(b []byte) error {
 	var unmarshaledMap map[string]string
 	if err := json.Unmarshal(b, &unmarshaledMap); err != nil {
 		return err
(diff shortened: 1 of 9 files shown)


jkl73:
I'm not sure why git didn't recognize the move `launcher_spec_test.go -> launch_spec_test.go``, the only meaningful change in this file is the new function at the end `TestIsHardened`

josephlr:
We will need to see what we do here after #246 is merged.

My preference would probably be to just have PCR 13 used unconditionally, but have the server side be capable of understanding multiple container launches.

> I'm not sure why git didn't recognize the move `launcher_spec_test.go -> launch_spec_test.go`, the only meaningful change in this file is the new function at the end `TestIsHardened`

For future reference, you can get Git to pay attention to this stuff by having one commit that just moves the file (with `git mv`) and another that actually makes the change.

jkl73:
> We will need to see what we do here after #246 is merged.
> 
> My preference would probably be to just have PCR 13 used unconditionally, but have the server side be capable of understanding multiple container launches.
> 
> > I'm not sure why git didn't recognize the move `launcher_spec_test.go -> launch_spec_test.go`, the only meaningful change in this file is the new function at the end `TestIsHardened`
> 
> For future reference, you can get Git to pay attention to this stuff by having one commit that just moves the file (with `git mv`) and another that actually makes the change.

Updated, removed resettable PCR change. I split the commit so now I think git can recognize the file rename.

josephlr:
Maybe I'm missing something, but what is the future purpose of the IsHardened value in the launch spec? Is this so we can entirely control the machine restart behavior inside the container launcher? or something else?

launcher/spec/launch_spec.go
```diff
@@ -121,30 +123,55 @@ func getRegion(client *metadata.Client) (string, error) {
...
 }
+
+func isHardened(kernelCmd string) bool {
+	args := strings.Fields(kernelCmd)
+
+	for _, arg := range args {
+		if strings.HasPrefix(arg, "confidential-space.hardened=") {
+			return strings.HasSuffix(arg, "=true")
+		}
+	}
```
josephlr, line 167: Can this be simplified to:

```go
for _, arg := range strings.Fields(kernelCmd) {
  if arg == "confidential-space.hardened=true" {
    return true
  }
}
return false
```

simulator/internal/internal_cgo.go
```diff
@@ -19,6 +19,8 @@ package internal
 // // Silence known warnings from the reference code and CGO code.
 // #cgo CFLAGS: -Wno-missing-braces -Wno-empty-body -Wno-unused-variable -Wno-uninitialized
 // // Link against the system OpenSSL
+// // Silence openssl deprecation warnings for ms-tpm-20-ref
+// #cgo CFLAGS: -Wno-deprecated-declarations
```
josephlr, line 23: This should go above the comment about linking against openssl

josephlr (approved):

launcher/spec/launch_spec.go
jkl73: done

simulator/internal/internal_cgo.go
jkl73: done

PR #246, opened 2022-09-10
Be much stricter when processing Attested COS State

josephlr:
Currently in `getVerifiedCosState` we do two things that are bad from a security perspective:
  1. We do not require the CEL events to use PCR 13
  2. We skip events that we do not understand (`record.Content.Type != cel.CosEventType`)

(1) is an issue because after a container launchers, we assume that a breakout is possible. So a malicious user can:
  - Breakout of a container
  - Extend fake events to PCR 14
  - Create their own fake CEL eventlog matching the PCR 14 extensions
  - Verify an attestation using PCR 14 and this fake CEL eventlog
  - Get fake container state to appear in the `MachineState`.

(2) is an issue because the `record.Content.Type` hasn't (yet) been verified in the loop. This means that a malicious user can:
  - Breakout of a container
  - Modify the CEL log to change the content type
  - Our processing will then skip this valid event
  - This causes data to be omitted from the `MachineState`

This change:
  - Adds the concept of a "launch separator" event that is measured right before a container is launched
  - Only creates `AttestedCosState` from PCR 13 data
  - Fails processing if we encounter any CEL events we don't understand (COS or otherwise)
  - Will not accept any `AttestedCosState` after a separator event is seen.
  - Will not accept duplicate `ImageRefType`, `ImageDigestType`, or `ImageIDType`
    - We can't do this for other events, as having multiple events for the other types isn't an error.
    - For example: multiple `EnvVarType` events are present if multiple environment variables are overridden.

This fixes the worst of the issues described above. Note, this doesn't change the client-side launcher at all. #247 will change the launcher code to actually measure in the `LaunchSeparator` event. Until then, a container breakout can cause some of the `AttestedCosState` to be wrong, but it cannot modify `image_reference`, `image_digest`, or `image_id`.

I also had to modify our tests somewhat, including adding `test.SkipForRealTPM(t)` as we can only run some of our tests against PCR13, which cannot be reset on a Real TPM.

### DIFF
--- a/server/eventlog.go
+++ b/server/eventlog.go
@@ -123,7 +123,7 @@ func getVerifiedCosState(coscel cel.CEL, pcrs *tpmpb.PCRs) (*pb.AttestedCosState
 	for _, record := range coscel.Records {
 		// COS State only comes from the CosEventPCR
 		if record.PCR != cel.CosEventPCR {
-			continue
+			return nil, fmt.Errorf("found unexpected PCR %d in CEL log", record.PCR)
 		}
 
 		// The Content.Type is not verified at this point, so we have to fail
--- a/server/eventlog.go
+++ b/server/eventlog.go
@@ -119,12 +119,18 @@ func getVerifiedCosState(coscel cel.CEL, pcrs *tpmpb.PCRs) (*pb.AttestedCosState
 	cosState.Container.EnvVars = make(map[string]string)
 	cosState.Container.OverriddenEnvVars = make(map[string]string)
 
+	seenSeparator := false
 	for _, record := range coscel.Records {
-		// ignore non COS CEL events
-		if !record.Content.IsCosTlv() {
+		// COS State only comes from the CosEventPCR
+		if record.PCR != cel.CosEventPCR {
 			continue
 		}
 
+		// The Content.Type is not verified at this point, so we have to fail
+		// if we see any events that we do not understand. This ensures that
+		// we either verify the digest of event event in this PCR, or we fail
+		// to replay the event log.
+		// TODO: See if we can fix this to have the Content Type be verified.
 		cosTlv, err := record.Content.ParseToCosTlv()
 		if err != nil {
 			return nil, err
@@ -135,11 +141,22 @@ func getVerifiedCosState(coscel cel.CEL, pcrs *tpmpb.PCRs) (*pb.AttestedCosState
 			return nil, err
 		}
 
+		// TODO: Add support for post-separator container data
+		if seenSeparator {
+			return nil, fmt.Errorf("found COS Event Type %v after LaunchSeparator event", cosTlv.EventType)
+		}
+
 		switch cosTlv.EventType {
 		case cel.ImageRefType:
+			if cosState.Container.GetImageReference() != "" {
+				return nil, fmt.Errorf("found duplicate ImageRef event")
+			}
 			cosState.Container.ImageReference = string(cosTlv.EventContent)
 
 		case cel.ImageDigestType:
+			if cosState.Container.GetImageDigest() != "" {
+				return nil, fmt.Errorf("found duplicate ImageDigest event")
+			}
 			cosState.Container.ImageDigest = string(cosTlv.EventContent)
 
 		case cel.RestartPolicyType:
@@ -150,6 +167,9 @@ func getVerifiedCosState(coscel cel.CEL, pcrs *tpmpb.PCRs) (*pb.AttestedCosState
 			cosState.Container.RestartPolicy = pb.RestartPolicy(restartPolicy)
 
 		case cel.ImageIDType:
+			if cosState.Container.GetImageId() != "" {
+				return nil, fmt.Errorf("found duplicate ImageId event")
+			}
 			cosState.Container.ImageId = string(cosTlv.EventContent)
 
 		case cel.EnvVarType:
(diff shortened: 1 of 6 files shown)


server/eventlog.go
```diff
@@ -119,12 +119,18 @@ func getVerifiedCosState(coscel cel.CEL, pcrs *tpmpb.PCRs) (*pb.AttestedCosState
 	cosState.Container.EnvVars = make(map[string]string)
 	cosState.Container.OverriddenEnvVars = make(map[string]string)
 
+	seenSeparator := false
 	for _, record := range coscel.Records {
-		// ignore non COS CEL events
-		if !record.Content.IsCosTlv() {
+		// COS State only comes from the CosEventPCR
+		if record.PCR != cel.CosEventPCR {
 			continue
```
alexmwu, line 126: Do we want to be strict (or at least the option to be strict) here and fail? I can imagine an attestation service implementer wanting to error on extended events into PCRs 12 or 14 since those are unexpected

server/eventlog.go
```diff
@@ -119,12 +119,18 @@ func getVerifiedCosState(coscel cel.CEL, pcrs *tpmpb.PCRs) (*pb.AttestedCosState
...
 		}
 
+		// The Content.Type is not verified at this point, so we have to fail
+		// if we see any events that we do not understand. This ensures that
+		// we either verify the digest of event event in this PCR, or we fail
+		// to replay the event log.
+		// TODO: See if we can fix this to have the Content Type be verified.
```
alexmwu, line 133: I believe the Content Type is verified below in `cel.VerifyDigests`. That marshals the TLV (including type) and writes the hash. It then compares the digest to digests in the CEL digest map.

server/eventlog.go
```diff
@@ -135,11 +141,22 @@ func getVerifiedCosState(coscel cel.CEL, pcrs *tpmpb.PCRs) (*pb.AttestedCosState
...
 
+		// TODO: Add support for post-separator container data
+		if seenSeparator {
+			return nil, fmt.Errorf("found COS Event Type %v after LaunchSeparator event", cosTlv.EventType)
+		}
+
 		switch cosTlv.EventType {
 		case cel.ImageRefType:
+			if cosState.Container.GetImageReference() != "" {
+				return nil, fmt.Errorf("found duplicate ImageRef event")
```
alexmwu, line 152: nit: "found more than one ImageRef event". Found duplicate makes it sound like everything in the event is the same (including the ref string).

Same below with ImageDigest and ID

alexmwu (approved):

server/eventlog.go
alexmwu: If it's an unknown Content Type as well (but verified), we should throw an error

jkl73 (approved):

server/eventlog.go
josephlr: Good idea, I think we should be very strict for now, and we can relax this later.

server/eventlog.go
josephlr: Good catch agreed.

server/eventlog.go
josephlr: So the issue is that the current implementation can only call `cel.VerifyDigests` can only be called _after_ `record.Content.ParseToCosTlv()` which checks the `content.Type` field without verifying it.

So if we run into unknown Content Types, they will be unverified. This isn't a problem in practice, as we will fail on unknown content types, and will fail on known content types that fail to verify.

alexmwu: Ah I see. This makes sense. I think we don't even have to call `VerifyDigests`. We can just call `record.Content.MarshalBinary()` and check the hashed value against the CEL digest map. After that, the Type is verifier and the TLV can be parsed to a CosTLV

server/eventlog.go
```diff
@@ -135,11 +141,22 @@ func getVerifiedCosState(coscel cel.CEL, pcrs *tpmpb.PCRs) (*pb.AttestedCosState
...
 
+		// TODO: Add support for post-separator container data
+		if seenSeparator {
+			return nil, fmt.Errorf("found COS Event Type %v after LaunchSeparator event", cosTlv.EventType)
+		}
+
 		switch cosTlv.EventType {
 		case cel.ImageRefType:
+			if cosState.Container.GetImageReference() != "" {
+				return nil, fmt.Errorf("found more than one ImageRef event")
```
jessieqliu, line 152: This can be fmt.Error (no formatting), same w other cases below

server/eventlog_test.go
```diff
@@ -644,12 +604,45 @@ func TestParsingCELEventLog(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	coscel.Records = append(coscel.Records, event)
+	buf = bytes.Buffer{}
+	if err := coscel.EncodeCEL(&buf); err != nil {
+		t.Fatal(err)
+	}
+	// extend digests to the PCR
+	for _, hash := range implmentedHash {
```
jessieqliu, line 621: implementedHash

jessieqliu (approved):
LGTM, some minor changes

server/eventlog_test.go
josephlr: Renamed to implementedHashes

server/eventlog.go
josephlr: Unfortunely, [`fmt.Errorf`](https://pkg.go.dev/fmt#Errorf) is the only Error function in the `fmt` package. The alternative would be to use `errors.New`, but staying with `fmt.Errorf` seems fine.

See:

jessieqliu: I got it confused with t.Error(), looks good then.

PR #247, opened 2022-09-10
Measure LaunchSeparator event

josephlr:
Depends on #246 

Actually measures in the launch separator event.

We probably want to rework when in the container creation/launch process we actually do our measurements, but that can wait for a followup CL, as it isn't security critical, just a defense-in-depth.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -321,7 +321,11 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
 		}
 	}
 
-	return nil
+	separator := cel.CosTlv{
+		EventType:    cel.LaunchSeparatorType,
+		EventContent: nil, // Success
+	}
+	return r.attestAgent.MeasureEvent(separator)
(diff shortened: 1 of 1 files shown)


launcher/container_runner.go
```diff
@@ -321,7 +321,11 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
 		}
 	}
 
-	return nil
+	separator := cel.CosTlv{
+		EventType:    cel.LaunchSeparatorType,
+		EventContent: nil, // Success
```
alexmwu, line 326: is this line necessary?

alexmwu (approved):

jkl73 (approved):

launcher/container_runner.go
josephlr: Not strictly speaking, but it allows us to be explicit that we have successfully configured a container and it is about to start. The other EventConents are reserved for errors.

Issue #257, opened 2022-10-28
Invalid PCR selection during `key.Attest()`


daniel-weisse:
Assuming we are running on a system with a TPM that does not support the full range of PCR Selections:

```shell
$ sudo tpm2_getcap pcrs
selected-pcrs:
  - sha1: [ ]
  - sha256: [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23 ]
  - sha384: [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23 ]
  - sha512: [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23 ]
```

Then calling `key.Attest(AttestOpts{Nonce:nonce})` fails with the following error:

```
failed to verify quote: given PCRs and Quote do not have the same PCR selection
```

The error is fixed by adjusting [`implementedPCRs()`](https://github.com/google/go-tpm-tools/blob/9d74a1e0c39200bf24524ad3e45b14560a098841/client/pcr.go#L36-L54) to only add a PCRSelection to the implemented PCRs list if a given PCRSelection is not an empty list.

I can open a PR to fix this.

PR #258, opened 2022-10-28
Fix: Don't add empty PCR selections to list of implemented PCRs

daniel-weisse:
### Proposed change

Only consider a PCRSelection "implemented" if the list of supported PCRs for a given hash algorithm is not an empty list.

Fixes https://github.com/google/go-tpm-tools/issues/257

### DIFF
--- a/client/pcr.go
+++ b/client/pcr.go
@@ -33,7 +33,7 @@ func min(a, b int) int {
 	return b
 }
 
-// Get a list of selections corresponding to the TPM's implemented PCRs
+// implementedPCRs returns a list of selections corresponding to the TPM's implemented PCRs.
 func implementedPCRs(rw io.ReadWriter) ([]tpm2.PCRSelection, error) {
 	caps, moreData, err := tpm2.GetCapability(rw, tpm2.CapabilityPCRs, math.MaxUint32, 0)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


client/pcr.go
```diff
@@ -33,7 +33,7 @@ func min(a, b int) int {
 	return b
 }
 
-// Get a list of selections corresponding to the TPM's implemented PCRs
```
alexmwu, line 36: very nit: This function would be nice to export, so I'd like to settle on a proper name

Looks like implemented PCR means the TPM supports certain set of PCR indexes with a hash algorithm, whereas allocated means the TPM has actually memory for the use of those PCRs. It's subtle, but we change the behavior here to actually return the allocated PCRs not implemented. So, let's change the name accordingly: `allocatedPCRs`

client/pcr.go
```diff
@@ -42,13 +42,17 @@ func implementedPCRs(rw io.ReadWriter) ([]tpm2.PCRSelection, error) {
...
-	sels := make([]tpm2.PCRSelection, len(caps))
-	for i, cap := range caps {
+	var sels []tpm2.PCRSelection
+	for _, cap := range caps {
 		sel, ok := cap.(tpm2.PCRSelection)
 		if !ok {
 			return nil, fmt.Errorf("unexpected data from GetCapability")
 		}
-		sels[i] = sel
+		// skip empty PCR selections
```
alexmwu, line 51: nit: "Skip empty (unallocated) PCR selections."

alexmwu (commented):
Thanks for the contribution! Naming nits but this generally LGTM.
Also looks like you need to sign the CLA for us to submit.

alexmwu (approved):

PR #259, opened 2022-11-02
Update go-sev-guest dependency to v0.2.5

deeglaze:
This change includes the replacement of glog for klog. This change removes the memory leak that was present in glog v1.0.0.

Signed-off-by: Dionna Glaze <dionnaglaze@google.com>


go.work.sum
```diff
@@ -1,2 +1,4 @@
+github.com/go-logr/logr v1.2.2 h1:ahHml/yUpnlb96Rp8HCvtYVPY8ZYpxq3g7UYchIYwbs=
```
alexmwu, line 1: What are these two lines?

deeglaze: They were from klog, which I removed. I don't know why it didn't get removed from the tidy. Removed.

alexmwu (approved):

PR #260, opened 2022-11-12
[launcher] Change retry behavior to reboot

jkl73:
* Update/simplify scripts and service unit file
* Update launcher directory, so main.go will be in launcher/launcher
* add retryable/nonretryable/workload error, currently workload error is equivlent to retryable error


launcher/launcher/main.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if err = startLauncher(); err != nil {
+		logger.Println(err)
+	}
+
+	exitCode = getExitCode(launchSpec.Hardened, launchSpec.RestartPolicy, err)
+}
+
+func getExitCode(isHardened bool, restartPolicy spec.RestartPolicy, err error) int {
+	exitCode := 0
```
alexmwu, line 81: thought: should we retry or shutdown for the debug image? I think this code would be simpler if we, full stop, hold for debug images. We can always make it configurable, but we only do the retries/shutdown on the hardened image for now.

alexmwu: The thing on the debug image is that it's likely faster to just SSH onto the machine and rerun it. I'm imagining something like:
```
if !isHardened {
 return holdRC
}

if err != nil {
		switch err.(type) {
		default:
			// unknown error
			exitCode = failRC
		case *launcher.RetryableError, *launcher.WorkloadError:
			if restartPolicy == spec.Always || restartPolicy == spec.OnFailure {
				exitCode = rebootRC
			} else {
				exitCode = failRC
			}
		case *launcher.NonRetryableError:
				exitCode = failRC
		}
	} else {
		// if no error
		if restartPolicy == spec.Always {
			exitCode = rebootRC
		} else {
			exitCode = successRC
		}
	}

	return exitCode
}
```

alexmwu (commented):
Mostly nits and style. The main issue I see is the complexity of getExitCode (I'd like to resolve this before more rounds of review).

launcher/launcher/main.go
```diff
@@ -0,0 +1,148 @@
+// package main is a program that will start a container with attestation.
```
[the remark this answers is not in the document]
jkl73: I feel like it is okay,,, I want to separate the main file in its own directory, so the container_runner.go can be in its own package. The image folder mostly contains scripts and config, I don't think source code should be there.

launcher/launcher/main.go
jkl73: done

launcher/container_runner.go
```diff
@@ -86,7 +87,7 @@ func fetchImpersonatedToken(ctx context.Context, serviceAccount string, audience
 func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.Token, launchSpec spec.LaunchSpec, mdsClient *metadata.Client, tpm io.ReadWriteCloser, logger *log.Logger) (*ContainerRunner, error) {
 	image, err := initImage(ctx, cdClient, launchSpec, token, logger)
 	if err != nil {
-		return nil, err
```
alexmwu, line 89: Since the default is a non-retryable error, is there any reason to annotate all the errors?

it kind of adds bloat. The only reason I see for keeping NonRetryableError around is either to ensure it doesn't get wrapped as a RetryableError or future-proofing

launcher/errors.go
```diff
@@ -0,0 +1,28 @@
...
+	Err error
+}
+
+// WorkloadError represents the result of an workload/task that is non-zero.
+type WorkloadError struct {
+	ReturnCode uint32
+}
+
+func (e *RetryableError) Error() string {
+	return e.Err.Error()
```
alexmwu, line 19: maybe we should prefix in case we want to log these: `failed with retryable error:`. same below

launcher/image/exit_script.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash -
+
+if [[ $EXIT_STATUS -eq 3 ]]
+then
+	# reboot after 2 min
+	shutdown --reboot +2
+fi
+
+if [[ $EXIT_STATUS -eq 0 ]] || [[ $EXIT_STATUS -eq 1 ]] || [[ $EXIT_STATUS -eq 2 ]]
```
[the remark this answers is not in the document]
alexmwu: good to know, thanks!

launcher/launcher/main.go
```diff
@@ -0,0 +1,142 @@
...
+	// if in a debug image, will always hold
+	if !isHardened {
+		return holdRC
+	}
+
+	if err != nil {
+		switch err.(type) {
+		default:
+			// unknown error
+			exitCode = failRC
```
alexmwu, line 93: you can use `fallthrough` and then remove the logic to set `exitCode = failRC`.

launcher/launcher/main_test.go
```diff
@@ -0,0 +1,131 @@
...
+			"debug, never restart, non-retryable error",
+			false, spec.Never, &launcher.NonRetryableError{}, holdRC,
+		},
+		{
+			"debug, onfailure restart, non-retryable error",
+			false, spec.OnFailure, &launcher.NonRetryableError{}, holdRC,
+		},
+		// non-retryable error, hardened image
+		{
+			"debug, always restart, non-retryable error",
```
alexmwu, line 111: copy paste issue here and below.  name should start with "hardened":
"hardened, always restart, ..."

alexmwu (approved):

launcher/errors.go
jkl73: done

launcher/launcher/main.go
jkl73: "cannot fallthrough  in a type switch"...

launcher/launcher/main_test.go
jkl73: fixed

launcher/container_runner.go
jkl73: Make sense, I was thinking there could be different use caes, but seems like for now. I will just keep the retryable error.

PR #261, opened 2022-11-16
launcher: Retry OIDC token refresh with backoff

alexmwu:
Currently, the token refresher stops after one attempt at refreshing the attestation verifier OIDC token.

Here, we use exponential backoff to retry refreshing the attestation verifier.
After the retries, the refresher goroutine stops.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
+	"github.com/cenkalti/backoff/v4"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/content"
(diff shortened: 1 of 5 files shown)


jessieqliu (approved):

launcher/container_runner_test.go
```diff
@@ -48,6 +50,10 @@ func (f *fakeAttestationAgent) Attest(ctx context.Context) ([]byte, error) {
 }
 
 func createJWTToken(t *testing.T, ttl time.Duration) []byte {
+	return createJWTTokenWithID(t, "test token", ttl)
+}
+
+func createJWTTokenWithID(t *testing.T, id string, ttl time.Duration) []byte {
```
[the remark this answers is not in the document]
alexmwu: changed to create JWT; this was the previous name :)

launcher/container_runner_test.go
```diff
@@ -233,6 +248,92 @@ func TestTokenIsNotChangedIfRefreshFails(t *testing.T) {
...
+
+func TestTokenRefreshRetryPolicy(t *testing.T) {
+	// Test retry policy tries 3 times.
+	for numTries := 1; numTries <= 3; numTries++ {
+		t.Run("RetryPolicyWith"+strconv.Itoa(numTries)+"Tries",
+			func(t *testing.T) { testRetryPolicyWithNTries(t, numTries, true) })
+	}
+}
+
+func testRetryPolicyWithNTries(t *testing.T, numTries int, shouldRefresh bool) {
```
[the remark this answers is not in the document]
alexmwu: done. also added comments to explain like in C++ (`/*expectRefresh=*/`)

launcher/container_runner.go
```diff
@@ -389,8 +426,17 @@ func (r *ContainerRunner) fetchAndWriteToken(ctx context.Context) error {
...
+					func() error {
+						duration, err = r.refreshToken(ctx)
+						return err
+					},
+					retry,
+					func(err error, t time.Duration) {
+						r.logger.Printf("failed to refresh attestation service token at time %v: %v", t, err)
+					})
 				if err != nil {
 					r.logger.Printf("failed to refresh attestation service token: %v", err)
```
jkl73, line 441: Should indicate will stop retrying refreshing token

launcher/container_runner.go
```diff
@@ -389,8 +426,17 @@ func (r *ContainerRunner) fetchAndWriteToken(ctx context.Context) error {
 				r.logger.Printf("token refreshing stopped: %v", ctx.Err())
 				return
 			case <-timer.C:
-				// Refresh token.
-				duration, err := r.refreshToken(ctx)
+				var duration time.Duration
+				// Refresh token with default retry policy.
+				err := backoff.RetryNotify(
```
jkl73, line 431: So if retry backoff go over the MaxElapsedTime or retried about 5 times, it will never retry again? (I thought it will retry every 1 hr indefinitely, but I think this is also okay)

alexmwu: Correct. We can always add indefinite retries. Either way, there will be some time period with hourly intervals with an expired token. So, workloads will have to rearchitect against this anyway. We can always change this if we find that users want this functionality, but this is simpler for now.

launcher/container_runner.go
alexmwu: done.

Issue #263, opened 2022-11-19
Fix server.VerifyAttestation with server.VerifyOpts


alexmwu:
currently, when you call `server.VerifyAttestation()` with `server.VerifyOpts{TrustedAKs: []crypto.PublicKey{...}}`, this causes a pacnic when the proto message *also* contains an AKCert. This is because we return a nil proto in https://github.com/google/go-tpm-tools/blob/99dd67bafc27f6bd83bfc2d85c2a9fd68ec10aa2/server/verify.go#L248. Thus, when we try to merge later on https://github.com/google/go-tpm-tools/blob/99dd67bafc27f6bd83bfc2d85c2a9fd68ec10aa2/server/verify.go#L174 the proto library panics with merging into a nil message

alexmwu:
Fixed by #265

PR #265, opened 2022-11-21
Fix panic catching in launcher and AKCert

### LINKED ISSUE #264: launcher: Don't defer os.Exit
https://github.com/google/go-tpm-tools/blob/master/launcher/launcher/main.go#L38-L40
when we defer `os.Exit()`, we swallow all panics because golang will process `defer`s before terminating.

See https://stackoverflow.com/questions/18963984/exit-with-error-code-in-go.

jkl73:
Fix #264 and #263 

Catch panic using defer recovery in launcher.
Return an empty MachineState pointer instead of nil in validateAKCert.

### DIFF
--- a/launcher/launcher/main.go
+++ b/launcher/launcher/main.go
@@ -35,12 +35,22 @@ var launchSpec spec.LaunchSpec
 
 func main() {
 	var exitCode int
-	defer func() {
-		os.Exit(exitCode)
-	}()
 
 	logger = log.Default()
+	// log.Default() outputs to stderr; change to stdout.
+	log.SetOutput(os.Stdout)
 	logger.Println("TEE container launcher initiating")
+
+	defer func() {
+		// catch panic, will only output to stdout, because cloud logging closed
+		// This should rarely happen (almost impossible), the only place can panic
+		// recover here is in the deferred logClient.Close().
+		if r := recover(); r != nil {
+			logger.Println("Panic:", r)
+			exitCode = 2
+		}
+		os.Exit(exitCode)
+	}()
 
 	mdsClient = metadata.NewClient(nil)
 	projectID, err := mdsClient.ProjectID()
@@ -71,6 +81,14 @@ func main() {
 		return
 	}
 
+	defer func() {
+		// catch panic, will also output to cloud logging if possible
+		if r := recover(); r != nil {
+			logger.Println("Panic:", r)
+			exitCode = 2
+		}
+		logger.Println("TEE container launcher exiting with exit code:", exitCode)
+	}()
 	if err = startLauncher(); err != nil {
 		logger.Println(err)
 	}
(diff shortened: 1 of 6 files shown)


launcher/launcher/main.go
```diff
@@ -35,13 +35,22 @@ var launchSpec spec.LaunchSpec
 
 func main() {
 	var exitCode int
-	defer func() {
-		os.Exit(exitCode)
-	}()
 
 	logger = log.Default()
+	// default logger output to stderr
```
alexmwu, line 40: stdout or stderr?

server/verify.go
```diff
@@ -245,7 +245,7 @@ func validateAKPub(ak crypto.PublicKey, opts VerifyOpts) error {
 
 func validateAKCert(akCert *x509.Certificate, opts VerifyOpts) (*pb.MachineState, error) {
 	if len(opts.TrustedRootCerts) == 0 {
-		return nil, validateAKPub(akCert.PublicKey.(crypto.PublicKey), opts)
+		return &pb.MachineState{}, validateAKPub(akCert.PublicKey.(crypto.PublicKey), opts)
```
alexmwu, line 248: I think `validateAKPub` should return `(*pb.MachineState, error)` instead of just `error`. That way we always get a non-nil MachineState xor a non-nil error. 

If for some reason we don't do this, we should explain why this is needed, since someone might try to "simplify" this.

alexmwu, line 248: Please add a test case for this and ideally check that it fails and then passes after changing the code.

launcher/launcher/main.go
```diff
@@ -35,13 +35,22 @@ var launchSpec spec.LaunchSpec
...
+	log.SetOutput(os.Stdout)
 	logger.Println("TEE container launcher initiating")
 
+	defer func() {
```
alexmwu, line 44: These defers seem convoluted. Why not just put the os.Exit at the bottom of the code? We don't need the `recover` if we do that.

launcher/launcher/main.go
jkl73: log.Default() will print to stderr, I want it to output to stdout. I will make the comment more clear.

server/verify.go
jkl73: added

jkl73: added

launcher/launcher/main.go
jkl73: This way we can output panic to cloud logging for better debug.

launcher/launcher/main.go
```diff
@@ -35,13 +35,22 @@ var launchSpec spec.LaunchSpec
 
 func main() {
 	var exitCode int
-	defer func() {
-		os.Exit(exitCode)
-	}()
 
 	logger = log.Default()
+	// default logger output to stderr, change it to stdout
```
alexmwu, line 40: nit: log.Default() outputs to stderr; change to stdout.

launcher/launcher/main.go
alexmwu: any reason we want stdout instead

alexmwu (approved):

launcher/launcher/main.go
jkl73: no particular, most of the log we output info, so stdout seems like a better fit.

PR #266, opened 2022-11-23
launcher: Change retry policy to hourly and add jitter to refresh time

alexmwu:
We want the retry to occur ~hourly (with jitter), as this will match the normal refreshed requests. This allows the launcher to provide a token even with long service outages.

Also, jitter on the refresh duration allows smearing of requests in the event many clients come online at similar times after an outage. Without this, load on the service side would be less balanced over time.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -57,35 +57,7 @@ const (
 	snapshotID  = "tee-snapshot"
 )
 
-/*
-	Values for token refresh and retries.
-
-With a 60m token expiration, the refresher goroutine will refresh beginning at .9*60=54m.
-
-Given the following default arguments, the retry sequence will be,
-assuming we go over the MaxElapsedTime:
-
-	 RetryInterval = 30s
-	 RandomizationFactor = 0.5
-	 Multiplier = 1.5
-	 MaxInterval = 180s
-	 MaxElapsedTime = 600s
-
-	 Request #  RetryInterval (seconds)  Randomized Interval (seconds)
-										 RetryInterval*[1-RandFactor, 1+RandFactor]
-	  1          30                      [15,   45]
-	  2          60                      [30,   90]
-	  3          120                     [60,   180]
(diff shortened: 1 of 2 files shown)


jessieqliu (approved):

launcher/container_runner.go
```diff
@@ -427,15 +409,45 @@ func (r *ContainerRunner) fetchAndWriteTokenWithRetry(ctx context.Context,
...
-// defaultRetryPolicy retries with:
-// initial interval of 30s, multiplication factor of 1.5
-// randomization factor of 0.5, max interval of 3m, and
-// max elapsed time of 10m.
+func getNextRefreshFromExpiration(expiration time.Duration) time.Duration {
+	diff := defaultRefreshJitter * float64(expiration)
+	center := defaultRefreshMultiplier * float64(expiration)
+	minRange := center - diff
+	return time.Duration(minRange + rand.Float64()*2*diff)
+
```
jkl73, line 417: extra space

launcher/container_runner.go
```diff
@@ -427,15 +409,45 @@ func (r *ContainerRunner) fetchAndWriteTokenWithRetry(ctx context.Context,
 	return nil
 }
 
-// defaultRetryPolicy retries with:
-// initial interval of 30s, multiplication factor of 1.5
-// randomization factor of 0.5, max interval of 3m, and
-// max elapsed time of 10m.
+func getNextRefreshFromExpiration(expiration time.Duration) time.Duration {
```
jkl73, line 412: Ideally, should have a unit test to test the returned time is always smaller than the expiration time.

jkl73 (approved):

launcher/container_runner.go
alexmwu: done. lmk if the test looks good to you.

michael-pregman (approved):

launcher/container_runner.go
jkl73: LGTM

PR #275, opened 2023-01-05
Add/update tests for images/launcher

jkl73:
Add a basic/smoke test which test some basic functionality of an image.
Update some scripts so they can take in args like vm name and zone.
Add a test to check user-data won't be executed in the hardened image.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,41 +1,89 @@
 substitutions:
+  # using this base image for now, because there is an issue causing the newest COS dev
+  # image not booting with cs.
   '_BASE_IMAGE': 'cos-dev-105-17234-0-0'
   '_OUTPUT_IMAGE_PREFIX': 'confidential-space'
   '_OUTPUT_IMAGE_SUFFIX': ''
 
 steps:
+# determine the base image
 - name: 'gcr.io/cloud-builders/gcloud'
+  id: BaseImageIdent
   env:
   - 'BASE_IMAGE=$_BASE_IMAGE'
+  script: |
+    #!/usr/bin/env bash
+
+    # if BASE_IMAGE is not specified in the substitutions, use the latest COS dev image
+    base_image=${BASE_IMAGE}
+    if [ -z ${base_image} ]
+    then
+      echo "getting the latest COS image"
+      base_image=$(gcloud compute images describe-from-family cos-dev --project cos-cloud | grep name | cut -d ' ' -f 2)
+    fi
+
+    echo ${base_image} > /workspace/base_image.txt
+
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: DebugImageBuild
+  waitFor: ['BaseImageIdent']
+  env:
+  - 'OUTPUT_IMAGE_PREFIX=$_OUTPUT_IMAGE_PREFIX'
+  - 'OUTPUT_IMAGE_SUFFIX=$_OUTPUT_IMAGE_SUFFIX'
+  script: |
+    #!/usr/bin/env bash
+    set -exuo pipefail
+
+    base_image=$(cat /workspace/base_image.txt)
+    echo "building the debug image: ${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX} with the base image: ${base_image}"
+    gcloud builds submit --config=launcher/image/cloudbuild.yaml \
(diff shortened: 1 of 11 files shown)


michael-pregman (approved):

launcher/image/test/util/read_serial.sh
```diff
@@ -1,23 +1,36 @@
...
     local next_cmd="${base_cmd} ${next}"
+    
+    # sleeping 5s for the next serial console read"
+    sleep 5
+
     local tmp=$(eval ${next_cmd})
     serial_out="$serial_out $tmp"
 
+    # break the loop if the workload is finished
+    if echo ${serial_out} | grep -qi "TEE container launcher exiting"; then
```
alexmwu, line 30: suggestion for future lifecycle tests: we can parse the exit code from here with the prefix "TEE container launcher exiting with exit code:"

launcher/image/test/test_basic.sh
```diff
@@ -0,0 +1,48 @@
+ #!/bin/bash
```
alexmwu, line 1: we should call this something like `test_workload.sh` or `test_launcher_workload.sh`

launcher/image/test/test_cloud_init_userdata_disabled.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
alexmwu, line 1: suggestion (can be in a follow up): this test and other "negative tests" should be written so that they can run on both debug and hardened. that way, we can see these tests pass in one environment and fail in the other.

alexmwu: speaking of, were you able to see this script fails on a non-hardened image?

alexmwu (commented):
Great work! Thanks for parallelizing the image builds too

launcher/image/test/test_basic.sh
```diff
@@ -0,0 +1,48 @@
...
+if echo $SERIAL_OUTPUT | grep -q 'workload args: \[/main arg_foo\]'
+then
+    echo "- arguments verified"
+else
+    echo "FAILED: arguments not verified"
+    echo 'TEST FAILED.' > /workspace/status.txt
+    print_serial=true
+fi
+
+if echo $SERIAL_OUTPUT | grep -q 'env_bar=val_bar'
```
alexmwu, line 28: where are arg_foo and val_bar passed in?

cloudbuild.yaml
```diff
@@ -1,41 +1,89 @@
 substitutions:
-  '_BASE_IMAGE': ''
+  # using this base image for now, becaues there cloud be an issue with the newest COS image.
```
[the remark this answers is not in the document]
jkl73: Done

cloudbuild.yaml
```diff
@@ -1,41 +1,89 @@
...
-    if [ -z ${base_image}]
+    if [ -z ${base_image} ]
     then
       echo "getting the latest COS image"
       base_image=$(gcloud compute images describe-from-family cos-dev --project cos-cloud | grep name | cut -d ' ' -f 2)
     fi
 
-    debug_image_name=${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX}
-    hardened_image_name=${OUTPUT_IMAGE_PREFIX}-hardened-${OUTPUT_IMAGE_SUFFIX}
+    echo $base_image > /workspace/base_image.txt
```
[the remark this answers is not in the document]
jkl73: Done

cloudbuild.yaml
```diff
@@ -1,41 +1,89 @@
...
+    exit
```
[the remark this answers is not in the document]
jkl73: deleted

launcher/image/test/util/read_serial.sh
jkl73: Yes, it is on my todo list

launcher/image/test/test_basic.sh
jkl73: changed to test_launcher_workload.sh

launcher/image/test/test_cloud_init_userdata_disabled.sh
jkl73: I think this test can run on debug image.
But this specific test should also pass on debug image, because user-data cloudinit is also disabled on debug image.

We can try to run this test on a regular cos image, and then it should fail.

launcher/image/test/test_basic.sh
jkl73: It is part of the image (when building with the Dockerfile), we can test cmd/env override in a separate test.

alexmwu (approved):

PR #277, opened 2023-01-13
Parse EFI App state from the TCG event log 

alexmwu:
The TCG PCClient firmware event log contains digests of all EFI
applications. These are measured by UEFI whenever launching an
application prior to handing off control to the boot loader or OS after
ExitBootServices().

This change introduces MachineState.EfiState. EfiState will contain all the PE/COFF digests of EFI applications measured by UEFI. These are typically bootloaders but may be the Linux kernel in cases the EFI stub mode (https://lwn.net/Articles/632528/). UEFI measures these values in PCR4.

This introduces a change to pull those EFI measurements out of the event
log so that VerifyAttestation users can validate the PCR4 values are
exactly what they expect. This may be useful in cases where the Secure
Boot hierarchy does not provide enough trust (e.g., a vulnerability in a
        signed component that requires a Secure Boot update).


alexmwu:
@brandonweeks mind giving this a review for security issues?

server/eventlog.go
```diff
@@ -422,6 +458,107 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
...
+			if seenSeparator4 {
+				return nil, errors.New("found duplicate Separator event in PCR4")
+			}
+			seenSeparator4 = true
+		case 5:
+			// Process ExitBootServices event.
+			if bytes.Equal(exitBootSvcDigest, event.GetDigest()) {
+				if evtType != EFIAction {
+					return nil, fmt.Errorf("PCR%d contains ExitBootServices event but non EFIAction type: %d",
+						index, event.GetUntrustedType())
```
jkl73, line 528: same as above, can use `evtType`

server/eventlog.go
```diff
@@ -422,6 +458,107 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
...
+				if seenCallingEfiApp {
+					return nil, fmt.Errorf("found duplicate CallingEFIApp event in PCR%d", index)
+				}
+				if seenSeparator4 {
+					return nil, fmt.Errorf("found CallingEFIApp event in PCR%d after separator event", index)
+				}
+				seenCallingEfiApp = true
+			}
+
+			if evtType == EFIBootServicesApplication {
```
jkl73, line 508: Before appending the efi hash, seems like need to check `seenCallingEfiApp` equals true here

brandonweeks (approved):

proto/attest.proto
```diff
@@ -190,6 +190,21 @@ message AttestedCosState {
   SemanticVersion launcher_version = 3;
 }
 
+message EfiApp {
+  // The PE/COFF digest of the EFI application (pulled from the raw event digest).
+  bytes digest = 1;
```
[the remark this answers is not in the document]
alexmwu: As discussed offline, it should be the same HashAlgo as in the top level MachineState.hash field

server/eventlog.go
alexmwu: Nice catch, thanks!

jkl73 (approved):

server/eventlog.go
```diff
@@ -422,6 +458,109 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
...
+	var seenCallingEfiApp bool
+	var seenExitBootServices bool
+	for _, event := range events {
+		index := event.GetPcrIndex()
+		evtType := event.GetUntrustedType()
+
+		switch index {
+		default:
+			// Only process PCRs 4 and 5.
+			continue
```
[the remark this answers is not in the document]
alexmwu: I think it's better to be safe and explicitly early exit on security-critical code. Now that I think of it, I prefer early exiting with the if statement.

PR #283, opened 2023-02-06
Upgrade to go-sev-guest v0.5.0

deeglaze:
The trust API has changed a little to allow the fake KDS implementation to work appropriately. The "bad root" tests that swap around certs were broken since the default root didn't have any x509 certs to swap. That has been fixed in v0.5.0.

### DIFF
--- a/server/verify_sev.go
+++ b/server/verify_sev.go
@@ -5,22 +5,44 @@ import (
 	spb "github.com/google/go-sev-guest/proto/sevsnp"
 	"github.com/google/go-sev-guest/validate"
 	sv "github.com/google/go-sev-guest/verify"
-	"github.com/google/go-sev-guest/verify/trust"
 )
+
+// The policy on GCE is to allow SMT, and eventually MigrateMA, but no debug bit.
+var defaultSevSnpGuestPolicy = sabi.SnpPolicy{
+	SMT:       true,
+	MigrateMA: true,
+}
 
 // VerifySnpOpts allows for customizing the functionality of VerifyAttestation's SEV-SNP verification.
 type VerifySnpOpts struct {
-	// ReportData is the expected REPORT_DATA field of the attestation report.
-	ReportData [sabi.ReportDataSize]byte
-	// TrustedRoots defines which AMD root (ARK) and intermediate (ASK) keys to trust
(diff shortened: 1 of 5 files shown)


server/verify_sev.go
```diff
@@ -5,38 +5,54 @@ import (
...
-	// Allow the debug bit to be set (should only be used for testing).
-	AllowDebugTestOnly bool
-	// Getter is the object that will fetch files from URLs.
-	Getter trust.HTTPSGetter
+	Validation   *validate.Options
+	Verification *sv.Options
+}
+
+// SevSnpDefaultPolicy returns a default validation policy for SEV-SNP attestation reports on GCE.
+func SevSnpDefaultPolicy(tpmNonce []byte) *validate.Options {
```
alexmwu, line 23: nit: should this be something like `SevSnpDefaultValidateOpts`?

server/verify_test.go
```diff
@@ -1022,27 +1024,34 @@ func TestVerifyAttestationWithSevSnp(t *testing.T) {
...
+						Getter:       kdsGetter,
+						TrustedRoots: badSnpRoot,
+					},
 				},
 			},
 			wantErr: "error verifying VCEK certificate",
 		},
-		{
+	}
+	// Production SEV VMs are not run with debug enabled, so this test must be skipped when the
```
alexmwu, line 1037: This comment makes it sound like the following test will fail in prod since they don't have debug enabled. However, we use SevSnpDefaultPolicy, not SevSnpDefaultPolicyForTest, which actually sets `policy.GuestPolicy.Debug = true`.

deeglaze: The PolicyForTest sets the *allowance* for a debug bit. By just using the default policy, we're not allowing the debug bit. Production doesn't set the debug bit, so the report will not fail this check that the debug bit is not allowed.

server/verify_sev.go
deeglaze: Done.

server/verify_test.go
alexmwu: Got it, thanks for clarifying!

alexmwu (approved):

PR #291, opened 2023-03-08
[launcher] Bump version and fix a kernel cmd issue

jkl73:
Bump to the latest COS beta image for the launcher image.
Add a tmp fix to the kernel cmd caused by the customizer.
Bump containerd to v1.6.18 to match the version COS uses.
Print struct field name when printing launch space in launcher.


cloudbuild.yaml
```diff
@@ -1,7 +1,7 @@
 substitutions:
   # using this base image for now, because there is an issue causing the newest COS dev
   # image not booting with cs.
-  '_BASE_IMAGE': 'cos-dev-105-17234-0-0'
+  '_BASE_IMAGE': '' # will use the latest cos dev image
```
michael-pregman, line 4: cos-dev is now at the 109 milestone. Do we have changes from cos-dev that we will want to incorporate eventually? If not, then it may be useful to switch to cos-beta for more stability between releases.

launcher/image/fixup_oem.sh
```diff
@@ -6,6 +6,16 @@ main() {
   fi
   mount /dev/sda12 /mnt/disks/efi
   sed -i -e 's|systemd.mask=usr-share-oem.mount||g' /mnt/disks/efi/efi/boot/grub.cfg
+
+  # TODO: Remove this fix once the upstream customizer fixed the bug.
```
michael-pregman, line 10: Is there an available reference to this?

cloudbuild.yaml
jkl73: I am also considering this, on the cos-dev track we may have a (slightly) newer kernel and other dependencies.

launcher/image/fixup_oem.sh
jkl73: I don't think so, I can create one internally to track.

alexmwu (approved):

michael-pregman (approved):

cloudbuild.yaml
jkl73: Switched to cos-beta 105

PR #292, opened 2023-03-16
test: Add OS Configuration OS Policy tests.

michael-pregman:
Add OS Policy assignment tests for both debug and hardened. Currently, OS Policy assignment rollout does not finish in a timely manner. These tests are part of "unstable" cloudbuilds until this is resolved.


launcher/image/test/test_os_config_os_policy.sh
```diff
@@ -0,0 +1,65 @@
...
+
+gcloud compute instances add-labels $2 --labels=shutdown-label=$2 --zone=$3 || true
+GCLOUD_OUTPUT=$(gcloud compute os-config os-policy-assignments create shutdown-policy --location=$3 --file=shutdown-ospolicy.yaml | tail -1 || true)
+
+if echo $GCLOUD_OUTPUT | grep -q 'Created OS policy assignment [shutdown-policy]'
+then
+    GCLOUD_OUTPUT=$(gcloud compute instances describe $2 --zone=$3 --format="value(status)" || true)
+else
+    echo 'TEST FAILED: OS policy assignment could not be created'
+    echo 'TEST FAILED.' > /workspace/status.txt
```
jkl73, line 43: the script should return here (it is already failed)

launcher/image/test/test_hardened_unstable_cloudbuild.yaml
```diff
@@ -0,0 +1,41 @@
+substitutions:
+  # Expects hardened image (not debug) and should have startup-script service
+  # disabled. google-startup-scripts.service is only enabled with multi-user.target.
+  '_IMAGE_NAME': ''
+  '_IMAGE_PROJECT': ''
+  # Add user-data in the metadata to test if it is disabled.
+  '_METADATA_FILE': 'startup-script=data/echo_startupscript.sh,user-data=data/cloud-init-config.yaml'
```
jkl73, line 7: Should we have startupscript enabled here? because seem like it is only testing os policy here

jkl73 (commented):
Also could you rebase the branch, it should fix those lint issues.

PR #293, opened 2023-03-23
Add attest and debugVerify command to gotpm

Pranjali-2501:
The attest command generates an attestation report containing the
quote, attestation key and TCG event log. Users can specify the type
of key used (GCE or go-tpm-tools generated attestation key) , random
value for nonce, name of files that contains text format and binary 
format of attestation report, TEEDevice and TEENonce for hardware 
attestation report, GCE instance and project Metadata (if  attestation
key is of GCE type) .

The debugVerify command uses a binary file that stores the attestation
report in binary format, nonce and output file to store verified machine
state of the attestation report.


cmd/attest_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+		})
+	}
+}
+
+func TestGceFlagFail(t *testing.T) {
+	rwc := test.GetTPM(t)
+	defer client.CheckedClose(t, rwc)
+	ExternalTPM = rwc
+	//@TODO This case works when we are able to create gce key with ExternalTPM.
```
deeglaze, line 95: Best to remove TODOs in a PR or open an issue to reference if you really can't get to it in the PR.

cmd/attest_test.go
```diff
@@ -0,0 +1,199 @@
...
+		key           string
+		algo          string
+		nonce         string
+		teenonce      string
+		outputText    string
+	}{
+		{"DefaultKey","","rsa","2222","",""},
+		{"Withnonce", "AK", "rsa", "2222", "", ""},
+		{"WithTextOutput", "AK", "rsa", "2222", "", "text1111"},
+		//@TODO These cases works when we are able to create gce key with ExternalTPM.
```
deeglaze, line 56: Remove TODO or open an Issue and reference it.

cmd/attest_test.go
```diff
@@ -0,0 +1,199 @@
...
+	ExternalTPM = rwc
+
+	//without nonce
+	RootCmd.SetArgs([]string{"attest","--key","AK"})
+	if err := RootCmd.Execute(); err == nil {
+		t.Error(err)
+	}
+
+	//with odd length nonce
+	RootCmd.SetArgs([]string{"attest","--nonce","12345","--key","AK"})
```
deeglaze, line 33: gofmt should be adding in the spaces between your commas. Please lint your code.

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+	Use:   "attest",
+	Short: "Create a remote attestation report",
+	Long: `Gather information for remote attestation.
+
+The Attestation report contains a quote on all available PCR banks ,a way to validate 
+the quote, and a TCG Event Log (Linux only).
+
+Use --key to specify the type of attestation key. It can be gceAK for GCE attestation
+key or AK for a custom attestation key.By default it uses gceAK.
+--algo flag override the public key algorithm for attestation key. If not provided then
```
alexmwu, line 51: ECC is more performant and our template uses an algorithm that is required to be implemented by the spec: https://trustedcomputinggroup.org/resource/tcg-algorithm-registry/.

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+			return fmt.Errorf("failed to create attestation key: %v", err)
+		}
+		defer attestationKey.Close()
+
+
+		attestopts := client.AttestOpts{}
+		attestopts.Nonce = nonce
+		if(len(teenonce) != 0){
+			attestopts.TEENonce = teenonce
+			attestopts.TEEDevice,err = client.CreateSevSnpDevice()
```
alexmwu, line 86: @deeglaze can you take a closer look at this section? We are tying it pretty closely with SEV-SNP for now, and I would prefer that our CLI commands work compatibly with future additions like TDX.

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+			if(err != nil){
+				return fmt.Errorf("collecting TEE attestation report: %v",err)
+			}
+		}		
+
+
+		attestation ,err := attestationKey.Attest(attestopts)
+		if(err != nil){
+			return fmt.Errorf("while collecting attestation report : %v",err)
```
alexmwu, line 95: nit: "failed to collect attestation report: %v"

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		defer attestationKey.Close()
+
+
+		attestopts := client.AttestOpts{}
+		attestopts.Nonce = nonce
+		if(len(teenonce) != 0){
+			attestopts.TEENonce = teenonce
+			attestopts.TEEDevice,err = client.CreateSevSnpDevice()
+			if(err != nil){
```
alexmwu, line 87: nit: "failed to open SEV-SNP device: %v"

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+Use --key to specify the type of attestation key. It can be gceAK for GCE attestation
+key or AK for a custom attestation key.By default it uses gceAK.
+--algo flag override the public key algorithm for attestation key. If not provided then
+by default rsa is used.
+--nonce attaches even length extra data to the report and guarantees a fresh quote.
+--teenonce attaches a 64 bytes extra data to the attestation report of hardware and 
+guarantees a fresh quote.
+--output flag provides file name and location to store attestation report in the binary
+format. If filename is not provided then output will be printed on stdout.
+--outputText flag can be used to add text format of attestation report in a file.
```
alexmwu, line 58: I will again register that you can download protoc and use it instead of building it into this CLI :).

Regardless, I think we should keep flags orthogonal as much as possible. Here, we could have a global flag that is useful elsewhere like `--format=prototext` and `--format=binarypb`. So instead of specifying `--outputText` and/or `--output` to print prototext, you just specify the format you want it in and `--output`.

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+	cmd.PersistentFlags().StringVar(&outputText,"outputText","","output file where attestation report stored in text format")
+}
+func addKeyFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&key, "key", "gceAK", "indicates type of attestation key to use <gceAK|AK>")
+}
+func addNonceflag(cmd *cobra.Command) {
+	cmd.PersistentFlags().BytesHexVar(&nonce, "nonce", []byte{}, "Hex encoded nonce for vTPM attestation , cannot be empty")
+}
+func addTeeNonceflag(cmd *cobra.Command) {
+	cmd.PersistentFlags().BytesHexVar(&teenonce, "teenonce", []byte{}, "Hex encoded teenonce for hardware attestation, can be empty")
```
alexmwu, line 194: nit: stylistically, all comments should be lower-case. Also, remove unnecessary spaces: "hex encoded nonce for vTPM attestation, cannot be empty"

cmd/attest_test.go
```diff
@@ -0,0 +1,199 @@
...
+	defer client.CheckedClose(t, rwc)
+	ExternalTPM = rwc
+
+	//without nonce
+	RootCmd.SetArgs([]string{"attest","--key","AK"})
+	if err := RootCmd.Execute(); err == nil {
+		t.Error(err)
+	}
+
+	//with odd length nonce
```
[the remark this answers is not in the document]
alexmwu: Also should end with a period.

cmd/open.go
```diff
@@ -20,9 +21,18 @@ func (ic ignoreClose) Close() error {
 	return nil
 }
 
+type ignoreCloseWithEventLogGetter struct {
```
alexmwu, line 24: This was a hacky workaround in the original PR #128. Perhaps we instead add an option to pass a `TcgEventLog` field in `client.AttestOpts` and then fill that in directly with GetEventLog if it doesn't exist.

alexmwu (commented):
Make sure to either sign the CLA or go through go/github.

cmd/attest_test.go
Pranjali-2501: Agreed , this PR is mostly to get the first review done with Alex. After concluding the discussion to clarify the functionality , I'll remove those TODOs and update you on this thread .

cmd/attest.go
Pranjali-2501: Noted.
I'll change all the comments to lower-case.
Formatting issues can be resolved using gofmt.

cmd/attest.go
Pranjali-2501: --algo flag is already defined in go-tpm-tools/cmd/flags.go and it takes rsa as a default value for --algo flag.

cmd/attest.go
Pranjali-2501: Noted.

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+					return fmt.Errorf("failed to write attestation proto: %v", err)
+			}
+		}
+		return nil
+	},
+	
+}
+
+func getInstanceMetadata() error{
+	if(len(gceZone) == 0){
```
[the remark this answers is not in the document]
Pranjali-2501: Sure

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+		out ,err := exec.Command("curl", "http://metadata.google.internal/computeMetadata/v1/instance/zone").Output()
```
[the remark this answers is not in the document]
Pranjali-2501: I have used some string slicing to get the correct value for Zone.
But as you suggested , we can get this using https://pkg.go.dev/cloud.google.com/go/compute/metadata 
package.
So  I'll change the implementation and use [Zone()](https://pkg.go.dev/cloud.google.com/go/compute/metadata#Client.Zone) instead of using curl command.

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+		if(err != nil){
+			return err
+		}
+		zone := string(out)
+		gceZone = zone[strings.Index(zone,"zones/")+6:]
+	}
+
+	if(gceInstanceID == 0){
```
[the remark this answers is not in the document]
Pranjali-2501: Noted.

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+		if(err != nil){
+			return err
+		}
+		gceInstanceID ,err = strconv.ParseUint(string(id),10,64)
+		if(err != nil){
+			return err
+		}
+	}
+
+	if(len(gceInstanceName)==0){
```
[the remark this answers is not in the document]
Pranjali-2501: Noted

cmd/attest_test.go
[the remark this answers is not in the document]
Pranjali-2501: Noted.

cmd/attest_test.go
Pranjali-2501: Thankyou , I will do gofmt before updating the pull request.

cmd/attest.go
Pranjali-2501: Noted. 
Instead of using --outputText , we will use --format to specify the type of output file.
But if format is set to prototext , then debugVerify will fail because it needs a binary file to read .

cmd/attest.go
alexmwu: you can just `return err` in that case.

cmd/attest.go
alexmwu: then arguably, `--format` should be used for input as well in `debugVerify`

cmd/attest.go
alexmwu: I'm sorry: I meant help text comments, not code comments. Comments should be capitalized and end with a period. Help text on flags should be lower case.

You still need to update the help text string here. gofmt does not format within strings to my knowledge.

cmd/attest.go
```diff
@@ -0,0 +1,193 @@
...
+		attestopts.Nonce = nonce
+		if len(teenonce) != 0 {
+			attestopts.TEENonce = teenonce
+			attestopts.TEEDevice, err = client.CreateSevSnpDevice()
+			if err != nil {
+				return fmt.Errorf("collecting TEE attestation report: %v", err)
+			}
+		}
+
+		if ExternalTPM != nil {
```
alexmwu, line 87: Please also add a test case for running this against the actual TPM on the host (skipping if unavailable) and showing it skips this logic.

cmd/attest.go
```diff
@@ -0,0 +1,193 @@
...
+			if elg, ok := ExternalTPM.(client.EventLogGetter); ok {
```
alexmwu, line 88: What if `!ok`? It's fine to attest without an event log, but ExternalTPM typically indicates using the simulator.

cmd/attest.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+			if _, err := dataOutput().Write(out); err != nil {
+				return fmt.Errorf("failed to write attestation proto: %v", err)
+			}
+		}
+		return nil
+	},
+}
+
+func getMetadata() error {
```
alexmwu, line 124: This should probably be called `getInstanceInfoFromMetadata` return a `*attest.GCEInstanceInfo`

cmd/attest.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		attestation, err := attestationKey.Attest(attestopts)
+		if err != nil {
+			return fmt.Errorf("failed to collect attestation report: %v", err)
+		}
+
+		if key == "gceAK" {
+			getMetadata()
+			addGceInstanceInfo(attestation)
```
alexmwu, line 103: We don't need this function. Just 
```
instanceInfo, err := getInstanceInfoFromMetadata()
if err != nil ...
attestation.InstanceInfo = instanceInfo
```

cmd/attest.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		if len(output) != 0 {
+			var out []byte
+			if format == "binarypb" {
+				out, err = proto.Marshal(attestation)
+				if err != nil {
+					return fmt.Errorf("failed to marshal attestation proto: %v", attestation)
+				}
+			} else {
```
alexmwu, line 113: Explicitly check the string for "textproto". Then return an error for incorrect flags. This checker util func should go in `flags.go`.

cmd/attest.go
```diff
@@ -0,0 +1,193 @@
...
+	attestation.InstanceInfo = &attest.GCEInstanceInfo{
+		Zone:          gceZone,
+		ProjectId:     gceProjectID,
+		ProjectNumber: gceProjectNumber,
+		InstanceName:  gceInstanceName,
+		InstanceId:    gceInstanceID,
+	}
+}
+
+func addFormatFlag(cmd *cobra.Command) {
```
alexmwu, line 172: Since it's shared, this should go in `flags.go`

cmd/attest.go
```diff
@@ -0,0 +1,193 @@
...
+	cmd.PersistentFlags().StringVar(&format, "format", "binarypb", "type of output file where attestation report stored <binarypb|prototext>")
+}
+func addKeyFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&key, "key", "AK", "indicates type of attestation key to use <gceAK|AK>")
+}
+func addNonceflag(cmd *cobra.Command) {
```
alexmwu, line 178: Since it's shared, this should go in `flags.go`

cmd/attest.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+		}
+		return nil
+	},
+}
+
+func getMetadata() error {
+
+	var err error
+	gceProjectID, err = metadata.ProjectID()
```
alexmwu, line 127: Please add tests against a fake MDS for every item we pull, and double check we get the correct information.

cmd/attest.go
Pranjali-2501: Sure , I'll change it that way.

cmd/attest.go
Pranjali-2501: Agree , it should return *attest.GCEInstanceInfo or error if unable to fetch the information from MDS server.

cmd/attest.go
Pranjali-2501: Yes , it is a shared flag and it should be in flags.go .
I'll move it in flags.go.

cmd/attest.go
Pranjali-2501: If !ok , it means that ExternalTPM is unable to retrieve the Eventlog that came through the simulator.
So , it should return an error if !ok from there only.
I'll change the implementation in the same way.

cmd/attest.go
Pranjali-2501: Agree , I'll move it in flags.go .

cmd/attest.go
Pranjali-2501: Marshaling is implemented in attest.go and Unmarshaling is implemented in debugVerify.go .
We have to check the value of the format flag just before doing marshal and unmarshal .
And implementation of `Marshal` and `Unmarshal` are not common. 


Also I have added  `if len(output) != 0 ...` this check to ensure that it will not print an attestation report on stdout if file name is not provided.
And planning to add the same check in dubugVerify.go while taking the file from `--input` .

cmd/attest.go
Pranjali-2501: As of now , we are unable to make a GCE key using ExternalTPM.
And without GCE key , we can't retrieve the GCEInstanceInfo as per the current implementation.

client/attest.go
```diff
@@ -236,8 +239,10 @@ func (k *Key) Attest(opts AttestOpts) (*pb.Attestation, error) {
 		}
 		attestation.Quotes = append(attestation.Quotes, quote)
 	}
-	if attestation.EventLog, err = GetEventLog(k.rw); err != nil {
-		return nil, fmt.Errorf("failed to retrieve TCG Event Log: %w", err)
+	if opts.TCGEventLog == nil {
```
deeglaze, line 242: This option appears only ever to be written and never read. Wouldn't the opts.TCGEventLog != nil case need to have attestation.EventLog = opts.TCGEventLog?

cmd/attest.go
Pranjali-2501: How to run a test with an actual TPM ?
IMO , test can only run against ExternalTPM.

client/attest.go
Pranjali-2501: Yes , this check should be there otherwise `attestation.EventLog` is always nil while testing.
Thank you .

cmd/attest_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+		// {"ProjectMetadataWithAK","1111","gceAK","sev-team-project","1071581149866","","",""},
+		// {"ProjectMetadataWithAK","1111","gceAK","sev-team-project","1071581149866","us-west1-b","",""},
+	}
+
+	for _, op := range tests {
+		t.Run(op.name, func(t *testing.T) {
+
+			attestFlagsArgs := []string{"attest", "--nonce", op.nonce, "--key", op.key}
+			if op.gceProjectId != "" {
```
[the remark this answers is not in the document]
Pranjali-2501: I want to understand how to use the reflect package appropriately so that I can use the learnings for go-tdx-guest work as well.  I will raise another request once reflect changes are done. I hope that works since they are more from a readability perspective and functionality will remain the same.

cmd/attest_test.go
Pranjali-2501: I have removed all those @TODOs. 
I'll be resolving this comment.

cmd/attest_test.go
Pranjali-2501: Yes , I have removed all TODO comments from the test file.

cmd/attest_test.go
Pranjali-2501: I have used gofmt in the new commits . Please review it once.
For now I am resolving this comment.

cmd/attest.go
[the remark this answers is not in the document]
Pranjali-2501: I have implemented it in the suggested way and removed GCE flags.
Now I am resolving this comment.

cmd/attest.go
```diff
@@ -0,0 +1,218 @@
...
+			if(len(gceProjectID) != 0 || gceProjectNumber != 0){
+				return fmt.Errorf("attestation key type AK cannot use Gce Flags")
+			}
+		}
+		
+
+		if(len(outputText) != 0){
+			protoOut := []byte(prototext.Format(attestation))
+			if _,err := outputProto().Write(protoOut); err != nil{
+				return fmt.Errorf("failed to write attesattion report in textProto format: %v",err)
```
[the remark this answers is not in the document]
Pranjali-2501: I have fixed that ,  please review it if required.

cmd/attest.go
Pranjali-2501: I have implemented it in the same way as discussed. Please review it if required.

cmd/attest.go
Pranjali-2501: I have moved the `--format` flag in flags.go and now resolve this comment.

client/attest.go
Pranjali-2501: I have implemented that .

cmd/attest.go
Pranjali-2501: I have implemented it in the same way as discussed above. Please review it.
For now I have resolved this comment.

cmd/attest.go
Pranjali-2501: I have moved `--nonce` flags in flags.go.

cmd/attest.go
Pranjali-2501: I have implemented it in the same way.
Also used gofmt with each modified file .
It is resolved from my side , please review it if required

cmd/attest_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func TestAttestTeeNonceFail(t *testing.T) {
+	rwc := test.GetTPM(t)
+	defer client.CheckedClose(t, rwc)
+	ExternalTPM = rwc
+
+	// non-nil TEENonce when TEEDevice is nil
+	RootCmd.SetArgs([]string{"attest", "--nonce", "1234", "--key", "AK", "--teenonce", "12345678"})
+	if err := RootCmd.Execute(); err == nil {
+		t.Error(err)
```
jkl73, line 90: same here, error is nil in the error

cmd/attest_test.go
```diff
@@ -0,0 +1,92 @@
...
+	// Without nonce.
+	RootCmd.SetArgs([]string{"attest", "--key", "AK"})
+	if err := RootCmd.Execute(); err == nil {
+		t.Error(err)
+	}
+
+	// With odd length nonce.
+	RootCmd.SetArgs([]string{"attest", "--nonce", "12345", "--key", "AK"})
+	if err := RootCmd.Execute(); err == nil {
+		t.Error(err)
```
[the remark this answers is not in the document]
Pranjali-2501: Thank you for pointing that out . 
I'll change that.

cmd/attest_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func TestNonce(t *testing.T) {
+	rwc := test.GetTPM(t)
+	defer client.CheckedClose(t, rwc)
+	ExternalTPM = rwc
+
+	// Without nonce.
+	RootCmd.SetArgs([]string{"attest", "--key", "AK"})
+	if err := RootCmd.Execute(); err == nil {
+		t.Error(err)
```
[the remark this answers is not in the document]
Pranjali-2501: I have added a message instead of returning nil.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+
+		createFunc := attestationKeys[key][keyAlgo]
+
+		attestationKey, err := createFunc(rwc)
+		if err != nil {
+			return fmt.Errorf("failed to create attestation key: %v", err)
+		}
+		defer attestationKey.Close()
+
+		attestopts := client.AttestOpts{}
```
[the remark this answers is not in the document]
Pranjali-2501: Sure.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+		tpm2.AlgECC: client.GceAttestationKeyECC,
+	},
+}
+
+var attestCmd = &cobra.Command{
+	Use:   "attest",
+	Short: "Create a remote attestation report",
+	Long: `Gather information for remote attestation.
+
+The Attestation report contains a quote on all available PCR banks ,a way to validate
```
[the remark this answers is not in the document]
Pranjali-2501: OK.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/google/go-tpm-tools/client"
+	"github.com/google/go-tpm-tools/proto/attest"
+	"github.com/google/go-tpm/tpm2"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/proto"
+)
+
+var (
+	key      string
+	teenonce []byte
```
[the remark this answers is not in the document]
Pranjali-2501: Sure.

cmd/attest_test.go
Pranjali-2501: I have updated it with a message instead of returning the nil error.

cmd/attest_test.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+			if op.teenonce != "" {
+				attestArgs = append(attestArgs, "--teenonce", op.teenonce)
+			}
+
+			RootCmd.SetArgs(attestArgs)
+			if err := RootCmd.Execute(); err != nil {
+				t.Error(err)
+			}
+
```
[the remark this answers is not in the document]
Pranjali-2501: Sorry for that.
I have removed the extra line.

cmd/debugVerify.go
```diff
@@ -0,0 +1,67 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		attestationBytes, err := io.ReadAll(dataInput())
+		if err != nil {
+			return err
+		}
+		attestation := &pb.Attestation{}
+
+		if format == "binarypb" {
+			err = proto.Unmarshal(attestationBytes, attestation)
+		} else {
```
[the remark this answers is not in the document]
Pranjali-2501: Sure , I'll change it to `else if` and add `else` which returns an error stating "format did not match"` .

cmd/debugVerify.go
```diff
@@ -0,0 +1,67 @@
...
+		out, err := marshalOptions.Marshal(ms)
+		if err != nil {
+			return nil
+		}
+		_, err = dataOutput().Write(out)
+		if err != nil {
+			return nil
+		}
+
+		return nil
```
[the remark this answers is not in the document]
Pranjali-2501: No , it should return an error. I 'll change it to 
```
if _, err := dataOutput().Write(out); err != nil {
			return fmt.Errorf("failed to write a verified attestation report: %v", err)
}
```

cmd/attest_test.go
```diff
@@ -0,0 +1,92 @@
...
+		nonce    string
+		teenonce string
+	}{
+		{"DefaultKey", "", "rsa", "2222", ""},
+		{"Withnonce", "AK", "ecc", "2222", ""},
+	}
+
+	for _, op := range tests {
+		t.Run(op.name, func(t *testing.T) {
+			secretFile1 := makeOutputFile(t, "attest.binarypb")
```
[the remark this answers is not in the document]
Pranjali-2501: Agree , we can do what you have suggested. 
But right now we are taking reference from seal.go and to make the cmd folder consistent, we implement the `makeOutputFile()` function to handle creation and deletion of temp files.

Also alex has suggested to not randomly create files on the user's workspace.

@alexmwu any suggestion on that?

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+	Args: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		rwc, err := openTpm()
+		if err != nil {
+			return err
+		}
+		defer rwc.Close()
+
+		createFunc := attestationKeys[key][keyAlgo]
```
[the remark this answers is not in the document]
Pranjali-2501: Sure , I'll add a check to verify that the value of the `--key` flag should be either "AK" or "gceAK".
Otherwise it will return a panic error.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+		if len(teenonce) != 0 {
+			attestopts.TEENonce = teenonce
+			attestopts.TEEDevice, err = client.CreateSevSnpDevice()
+			if err != nil {
+				return fmt.Errorf("collecting TEE attestation report: %v", err)
+			}
+		}
+
+		if ExternalTPM != nil {
+			attestopts.TCGEventLog, err = ExternalTPM.(client.EventLogGetter).EventLog()
```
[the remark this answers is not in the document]
Pranjali-2501: Yes , client.Attest has handled that logic , but when we run tests , it was not able to get the simulated eventlog.
So alex has suggested to add a TCGEventLog field in the client.AttestOpts and fill it if `ExternalTPM != nil`.
Also add a check in `client.Attest()` that is if `TCGEventLog == nil` then call `GetEventLog` .

cmd/attest.go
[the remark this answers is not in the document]
Pranjali-2501: I have added a check to verify the key. Kindly review the changes.

cmd/attest.go
```diff
@@ -0,0 +1,179 @@
...
+
+var attestCmd = &cobra.Command{
+	Use:   "attest",
+	Short: "Create a remote attestation report",
+	Long: `Gather information for remote attestation.
+
+The Attestation report contains a quote on all available PCR banks, a way to validate 
+the quote, and a TCG Event Log (Linux only).
+
+Use --key to specify the type of attestation key. It can be gceAK for GCE attestation
```
alexmwu, line 40: "By default it uses AK."

cmd/attest.go
```diff
@@ -0,0 +1,179 @@
...
+	Args: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		rwc, err := openTpm()
+		if err != nil {
+			return err
+		}
+		defer rwc.Close()
+
+		var attestationKey *client.Key
```
alexmwu, line 61: We can simplify this as follows:
```
algoToAKCreate, ok := attestationKeys[key]
if !ok {
  return fmt.Errorf("key should be one of the following: %v", maps.Keys(attestationKeys))
}
createFunc := algoToAKCreate[keyAlgo]
attestationKey, err = createFunc(rwc)
...
```

cmd/attest.go
```diff
@@ -0,0 +1,179 @@
...
+			return fmt.Errorf("key should be either AK or gceAK")
+		}
+
+		attestOpts := client.AttestOpts{}
+		attestOpts.Nonce = nonce
+		if len(teeNonce) != 0 {
+			attestOpts.TEENonce = teeNonce
+			attestOpts.TEEDevice, err = client.CreateSevSnpDevice()
+			if err != nil {
+				return fmt.Errorf("collecting TEE attestation report: %v", err)
```
alexmwu, line 79: "failed to collect TEE attestation report: %v"

cmd/attest.go
alexmwu: Please rebase on #302 and remove this logic. The tests should continue to pass.

cmd/attest.go
alexmwu: Please remove the `if len(output) != 0` check. We should allow users to pipe from stdout into a file or into another command.

cmd/attest.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+		}
+
+		if ExternalTPM != nil {
```
jkl73, line 83: We have implemented the EventLog() function for externalTPM in this #302, you can rebase the change, and then the logic here should be simpler, you can just call:

```
attestOpts.TCGEventLog, err := client.GetEventLog(rwc)
if err != nil {
    ....
```

cmd/flags.go
```diff
@@ -142,6 +144,15 @@ func addHashAlgoFlag(cmd *cobra.Command, hashAlgo *tpm2.Algorithm) {
 	cmd.PersistentFlags().Var(&f, "hash-algo", "hash algorithm: "+f.Allowed())
 }
 
+func addNonceflag(cmd *cobra.Command) {
```
alexmwu, line 147: nit: addNonceFlag

cmd/debugVerify.go
```diff
@@ -0,0 +1,67 @@
...
+
+	pb "github.com/google/go-tpm-tools/proto/attest"
+	"github.com/google/go-tpm-tools/server"
+	"github.com/google/go-tpm/tpm2"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/proto"
+)
+
+var debugVerifyCmd = &cobra.Command{
+	Use:   "debugVerify",
```
alexmwu, line 16: @jkl73 and I discussed this yesterday. Can we make this "verify" and add debug as a suboption of verify? That way we can reuse the top-level verify command for talking to the GCE attestation verifier in the future. We can also make it a flag, but it seems better suited as a subcommand since debug would be orthogonal to actually verifying.

cmd/attest_test.go
```diff
@@ -0,0 +1,339 @@
...
+			}
+			defer k.Close()
+			attestOpts := client.AttestOpts{}
+			attestOpts.Nonce = op.nonce
+			attestation, err := k.Attest(attestOpts)
+			if err != nil {
+				t.Error(err)
+			}
+			// Validate AK Pub.
+			KeyPub, err := gceAttestationKey(t, rwc, op.keyAlgo, data)
```
alexmwu, line 270: nit: `KeyPub` -> `ak`

cmd/attest_test.go
```diff
@@ -0,0 +1,339 @@
...
+		Type:       tpm2.AlgRSA,
+		NameAlg:    tpm2.AlgSHA256,
+		Attributes: tpm2.FlagSignerDefault,
+		RSAParameters: &tpm2.RSAParams{
+			Sign: &tpm2.SigScheme{
+				Alg:  tpm2.AlgRSASSA,
+				Hash: tpm2.AlgSHA256,
+			},
+			KeyBits: 2048,
+			ModulusRaw: []byte{240, 53, 205, 125, 1, 223, 48, 193, 34, 51, 180, 74, 245, 203, 239, 248, 235, 135, 119, 178, 102, 200, 62, 85, 151, 199,
```
alexmwu, line 58: Why set the points and modulus?

cmd/metadata.go
```diff
@@ -0,0 +1,72 @@
+package cmd
```
alexmwu, line 1: rename this to `fake_metadata.go` to make it clear it's a test fake

cmd/attest_test.go
```diff
@@ -0,0 +1,339 @@
...
+				t.Fatalf("failed to encode GCEAKTemplateRSA: %v", err)
+			}
+			k, err := gceAttestationKey(t, rwc, op.keyAlgo, data)
+			if err != nil {
+				t.Error(err)
+			}
+			defer k.Close()
+			attestOpts := client.AttestOpts{}
+			attestOpts.Nonce = op.nonce
+			attestation, err := k.Attest(attestOpts)
```
alexmwu, line 265: We're testing `client` package functionality here. We should instead be doing calling the CLI:
```
if err := RootCmd.Execute(); err != nil {
```

This will also require setting up the fake MDS in this test, which the test should also verify.

cmd/attest.go
Pranjali-2501: OK , I'll remove this check.

cmd/attest.go
Pranjali-2501: I have rebase it with #302 and removed this check `if ExternalTPM != nil`.
Thankyou.

cmd/attest.go
Pranjali-2501: Thank you for adding the EventLog() function .. I have rebase it and removed `ExternalTPM != nil` check.

cmd/metadata.go
Pranjali-2501: Sure , I'll rename it.

cmd/flags.go
Pranjali-2501: Done.

cmd/attest_test.go
Pranjali-2501: Sure.

cmd/attest_test.go
Pranjali-2501: If we use CLI to test this ,then it will try to create a GCEAK from the client/attest.go and give the same error .
```
--- FAIL: TestAttestWithGCEAK/gceAK:RSA (0.08s)
        attest_test.go:284: failed to create attestation key: read error at index 29425667: decoding NV_ReadPublic response: handle 1, error code 0xb : the handle is not correct for the use
```

cmd/attest_test.go
Pranjali-2501: When I tried to fetch the GCE key template from NV data , I found that it has a points field for ECC and a modulus field for RSA algo based GCE key.
That is why I have added it in the template. 

Also if I remove this then GCEAKTemplate would be the same as AKTemplate .

cmd/attest.go
Pranjali-2501: I have changed this.

cmd/attest.go
Pranjali-2501: I have imported the maps package  and implemented it in the suggested way.

cmd/attest.go
Pranjali-2501: Yes by default it uses AK.

cmd/attest_test.go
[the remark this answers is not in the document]
Pranjali-2501: Thank you @jkl73 for identifying that.
I have used `os.RemoveAll(filename)` to delete temp files after the test finish.

cmd/debugVerify.go
Pranjali-2501: sure @alexmwu . I have changed it to "**debugVerify** -> **verify**" and add a **debug** subcommand .

cmd/debugVerify.go
```diff
@@ -0,0 +1,67 @@
...
+	"io"
+
+	pb "github.com/google/go-tpm-tools/proto/attest"
+	"github.com/google/go-tpm-tools/server"
+	"github.com/google/go-tpm/tpm2"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/proto"
+)
+
+var debugVerifyCmd = &cobra.Command{
```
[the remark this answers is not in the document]
Pranjali-2501: Sure @jkl73 , I have changed "debugVerify -> verify" and added a debug subcommand to verify.

cmd/attest_test.go
alexmwu: That shouldn't matter since the GCEAKTemplate you have here and AKTemplate are created under different parents. Please remove, since this makes it harder to read. The points should be set when you create the key. Setting it on input is useful when loading a key generated externally from the TPM.

cmd/verify.go
```diff
@@ -0,0 +1,73 @@
+package cmd
```
alexmwu, line 1: Can you make sure there is a test for this command line option?

cmd/attest_test.go
Pranjali-2501: Okay , I have removed points and modulus fields from GCEKeyTemplate.

cmd/attest.go
Pranjali-2501: The maps package is supported by go version 1.18 or more , so for now I'll add a hard coded error message saying `"key should be either gceAK or AK"`

cmd/attest.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		defer rwc.Close()
+
+		var attestationKey *client.Key
+		algoToCreateAK, ok := attestationKeys[key]
+		if !ok {
+			return fmt.Errorf("key should be either AK or gceAK")
+		}
+		createFunc := algoToCreateAK[keyAlgo]
+		attestationKey, err = createFunc(rwc)
```
alexmwu, line 67: defer attestationKey.Close() here

jkl73 (commented):
LGTM. Seems like you picked the newer commits after your initial commits, so this PR is showing some commits that's already in the master/main.

Make sure you rebase it correctly so only new commits are showing up in this PR, and there is no merge conflict.

https://stackoverflow.com/questions/9790448/how-to-update-a-pull-request-from-forked-repo
https://git-scm.com/book/en/v2/Git-Branching-Rebasing

cmd/attest.go
Pranjali-2501: Yes , I have added that.

Pranjali-2501:
Reopen this PR to merge attest-cli in go-tpm-tools.

cmd/verify.go
```diff
@@ -0,0 +1,80 @@
...
+       } else {
+           return fmt.Errorf("format should be either binarypb or textproto")
+       }
+       if err != nil {
+           return fmt.Errorf("fail to unmarshal attestation report: %v", err)
+       }
+
+
+       pub, err := tpm2.DecodePublic(attestation.GetAkPub())
+
```
jkl73, line 47: nit: extra lines

cmd/verify.go
```diff
@@ -0,0 +1,80 @@
...
+           err = proto.Unmarshal(attestationBytes, attestation)
+       } else if format == "textproto" {
+           err = unmarshalOptions.Unmarshal(attestationBytes, attestation)
+       } else {
+           return fmt.Errorf("format should be either binarypb or textproto")
+       }
+       if err != nil {
+           return fmt.Errorf("fail to unmarshal attestation report: %v", err)
+       }
+
```
jkl73, line 44: nit: extra lines

jkl73 (commented):
Thanks for the update! Mostly nits. Not sure what IDE you're using, but try to run "gofmt -s -w <file>.go" to fix those lint issues.

cmd/verify.go
Pranjali-2501: As discussed , I have created a verify_test.go for testing the verify command line option.

cmd/attest_test.go
Pranjali-2501: I have changed the implementation of `TestAttestWithGCEAK()` and use attest CLI to test it.

cmd/verify.go
Pranjali-2501: I have run go fmt.
That extra line is removed now.

cmd/verify.go
Pranjali-2501: I have removed that by running go fmt.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+	Short: "Create a remote attestation report",
+	Long: `Gather information for remote attestation.
+The Attestation report contains a quote on all available PCR banks, a way to validate 
+the quote, and a TCG Event Log (Linux only).
+Use --key to specify the type of attestation key. It can be gceAK for GCE attestation
+key or AK for a custom attestation key. By default it uses AK.
+--algo flag override the public key algorithm for attestation key. If not provided then
+by default rsa is used.
+--nonce attaches even length extra data to the report and guarantees a fresh quote.
+--teenonce attaches a 64 bytes extra data to the attestation report of hardware and
```
deeglaze, line 43: 64 bytes is just a coincidence that both TDX and SEV-SNP have the same length for their report_data field. A different TEE technology could have a different length.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+	Use:   "attest",
+	Short: "Create a remote attestation report",
+	Long: `Gather information for remote attestation.
+The Attestation report contains a quote on all available PCR banks, a way to validate 
+the quote, and a TCG Event Log (Linux only).
+Use --key to specify the type of attestation key. It can be gceAK for GCE attestation
+key or AK for a custom attestation key. By default it uses AK.
+--algo flag override the public key algorithm for attestation key. If not provided then
+by default rsa is used.
+--nonce attaches even length extra data to the report and guarantees a fresh quote.
```
deeglaze, line 42: even length? I thought TPM 2.0 used 48 bytes. The "fresh quote" claim is only a guarantee if the user is providing a true nonce. It's not a claim that this library can make on its own.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+--teenonce attaches a 64 bytes extra data to the attestation report of hardware and 
+guarantees a fresh quote.
+--format flag specify the type of output file. It can be "binarypb" or "textproto".
+By default it uses binarypb.
+--output flag provides file name to store attestation report in the format specified
```
deeglaze, line 47: nit: the file name

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+
+var attestCmd = &cobra.Command{
+	Use:   "attest",
+	Short: "Create a remote attestation report",
+	Long: `Gather information for remote attestation.
+The Attestation report contains a quote on all available PCR banks, a way to validate 
+the quote, and a TCG Event Log (Linux only).
+Use --key to specify the type of attestation key. It can be gceAK for GCE attestation
+key or AK for a custom attestation key. By default it uses AK.
+--algo flag override the public key algorithm for attestation key. If not provided then
```
deeglaze, line 40: nit: overrides

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+		var out []byte
+		if format == "binarypb" {
+			out, err = proto.Marshal(attestation)
+			if err != nil {
+				return fmt.Errorf("failed to marshal attestation proto: %v", attestation)
+			}
+		} else if format == "textproto" {
+			out = []byte(marshalOptions.Format(attestation))
+		} else {
+			return fmt.Errorf("format should be either binarypb or textproto")
```
deeglaze, line 108: flag validation should happen before all the TPM interactions.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to create attestation key: %v", err)
+		}
+		defer attestationKey.Close()
+
+		attestOpts := client.AttestOpts{}
+		attestOpts.Nonce = nonce
+		if len(teeNonce) != 0 {
+			attestOpts.TEENonce = teeNonce
+			attestOpts.TEEDevice, err = client.CreateSevSnpDevice()
```
deeglaze, line 75: getTeeAttestationReport will choose the device if TEEDevice is not set, so I don't think you need to make that default decision here.

cmd/attest_test.go
```diff
@@ -0,0 +1,279 @@
...
+		t.Error("expected not-nil error")
+	}
+	// With odd length nonce.
+	RootCmd.SetArgs([]string{"attest", "--nonce", "12345", "--key", "AK"})
+	if err := RootCmd.Execute(); err == nil {
+		t.Error("expected not-nil error")
+	}
+}
+
+func TestAttestPass(t *testing.T) {
```
deeglaze, line 91: Any test for SEV-SNP simulation? Note there is a testclient device from go-sev-guest that will select hardware or fake based on the device path.

cmd/verify_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+	"testing"
+
+	"github.com/google/go-tpm-tools/client"
+	"github.com/google/go-tpm-tools/internal/test"
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
+)
+
+func TestVerifyNoncePass(t *testing.T) {
```
deeglaze, line 13: Please also test a SEV-SNP verification.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, err
+	}
+
+	return instanceInfo, err
+}
+
+func addKeyFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&key, "key", "AK", "indicates type of attestation key to use <gceAK|AK>")
+}
+func addTeeNonceflag(cmd *cobra.Command) {
```
alexmwu, line 161: I notice this isn't added in verify.go. How does that test pass?

cmd/attest_test.go
```diff
@@ -0,0 +1,279 @@
...
+			if err != nil {
+				t.Fatalf("failed to encode GCEAKTemplateRSA: %v", err)
+			}
+			err = setGCETemplate(t, rwc, op.keyAlgo, data)
+			if err != nil {
+				t.Error(err)
+			}
+			RootCmd.SetArgs([]string{"attest", "--nonce", "1234", "--key", "gceAK", "--algo", op.keyAlgo, "--output", secretFile1, "--format", "binarypb"})
+			if err := RootCmd.Execute(); err != nil {
+				t.Error(err)
```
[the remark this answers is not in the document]
jkl73: Seems like this test should only run on a GCE VM. I think we can SKIP the test if it's not running on a GCE VM environment.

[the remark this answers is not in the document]
Pranjali-2501: For me it is working fine.
```
Error: failed to collect attestation report : provided nonce must not be empty
Error: invalid argument "12345" for "--nonce" flag: encoding/hex: odd length hex string
Error: fail to unmarshal attestation report: proto: syntax error (line 2:1): invalid field name:
Error: fail to unmarshal attestation report: proto: cannot parse invalid wire-format data
Error: format should be either binarypb or textproto
Error: failed to collect TEE attestation report: could not open AMD SEV guest device at /dev/sev-guest: no such file or directory
Error: unsealing data: session 1, error code 0x1d : a policy check failed
Error: unsealing data: failed to certify PCRs: PCR 23 mismatch: expected 0000000000000000000000000000000000000000000000000000000000000000, got 9ef814b42fa0be12d197c44d3e8e03441a4b1118237658368ba1351090e556ed
Error: unsealing data: failed to certify PCRs: PCR 23 mismatch: expected 9ef814b42fa0be12d197c44d3e8e03441a4b1118237658368ba1351090e556ed, got 12a0883f16abf44dcc4cac1dec108eb99c652fed124c6989eddc811fe8effb64
Error: verifying attestation: failed to verify quote: quote extraData [18 52] did not match expected extraData [67 33]
PASS
ok      github.com/google/go-tpm-tools/cmd      2.297s
```

cmd/flags.go
```diff
@@ -142,6 +144,15 @@ func addHashAlgoFlag(cmd *cobra.Command, hashAlgo *tpm2.Algorithm) {
 	cmd.PersistentFlags().Var(&f, "hash-algo", "hash algorithm: "+f.Allowed())
 }
 
+func addNonceFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().BytesHexVar(&nonce, "nonce", []byte{}, "hex encoded nonce for vTPM attestation , cannot be empty")
```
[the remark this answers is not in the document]
Pranjali-2501: sure

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+The Attestation report contains a quote on all available PCR banks, a way to validate 
+the quote, and a TCG Event Log (Linux only).
+Use --key to specify the type of attestation key. It can be gceAK for GCE attestation
+key or AK for a custom attestation key. By default it uses AK.
+--algo flag override the public key algorithm for attestation key. If not provided then
+by default rsa is used.
+--nonce attaches even length extra data to the report and guarantees a fresh quote.
+--teenonce attaches a 64 bytes extra data to the attestation report of hardware and 
+guarantees a fresh quote.
+--format flag specify the type of output file. It can be "binarypb" or "textproto".
```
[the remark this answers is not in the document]
Pranjali-2501: Done

cmd/verify.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+var verifyCmd = &cobra.Command{
+	Use:   "verify",
+	Short: "Output attestation related data",
+	Args:  cobra.NoArgs,
+}
+var debugCmd = &cobra.Command{
+	Use:   "debug",
+	Short: "VerifyAttestation and Output attestation related data",
```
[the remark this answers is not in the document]
Pranjali-2501: Sure , I will change that.

cmd/verify.go
```diff
@@ -0,0 +1,71 @@
...
+	pb "github.com/google/go-tpm-tools/proto/attest"
+	"github.com/google/go-tpm-tools/server"
+	"github.com/google/go-tpm/tpm2"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/proto"
+)
+
+var verifyCmd = &cobra.Command{
+	Use:   "verify",
+	Short: "Output attestation related data",
```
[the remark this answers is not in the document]
Pranjali-2501: I have changed that.

cmd/attest_test.go
```diff
@@ -0,0 +1,279 @@
...
+			Sign: &tpm2.SigScheme{
+				Alg:  tpm2.AlgRSASSA,
+				Hash: tpm2.AlgSHA256,
+			},
+			KeyBits: 2048,
+		},
+	}
+}
+
+func setGCETemplate(tb testing.TB, rwc io.ReadWriteCloser, algo string, data []byte) error {
```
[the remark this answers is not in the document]
Pranjali-2501: Sure , I'll change the function name.
Also add a comment stating "Need to call tpm2.NVUndefinespace on the handle with authHandle tpm2.HandlePlatform."

cmd/attest_test.go
```diff
@@ -0,0 +1,279 @@
...
+	rwc := test.GetTPM(t)
+	defer client.CheckedClose(t, rwc)
+	ExternalTPM = rwc
+	tests := []struct {
+		name  string
+		key   string
+		algo  string
+		nonce string
+	}{
+		{"DefaultKey", "", "rsa", "2222"},
```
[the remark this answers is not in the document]
Pranjali-2501: AK is used as default, so instead of doing this  I will first make `AK` with `rsa` and then with `ecc`.

cmd/attest_test.go
```diff
@@ -0,0 +1,279 @@
...
+			err = setGCETemplate(t, rwc, op.keyAlgo, data)
+			if err != nil {
+				t.Error(err)
+			}
+			RootCmd.SetArgs([]string{"attest", "--nonce", "1234", "--key", "gceAK", "--algo", op.keyAlgo, "--output", secretFile1, "--format", "binarypb"})
+			if err := RootCmd.Execute(); err != nil {
+				t.Error(err)
+			}
+
+			defer tpm2.NVUndefineSpace(rwc, "", tpm2.HandlePlatform, tpmutil.Handle(getIndex[op.keyAlgo]))
```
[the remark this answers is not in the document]
Pranjali-2501: If we `NVUndefineSpace` in `setGCETemplate` , then it will close it in the scope .
You have suggested to take it out from the scope.
https://github.com/google/go-tpm-tools/pull/293#discussion_r1165782256

cmd/verify_test.go
```diff
@@ -0,0 +1,100 @@
...
+			if err := RootCmd.Execute(); err != nil {
+				t.Error(err)
+			}
+
+			RootCmd.SetArgs([]string{"verify", "debug", "--nonce", op.nonce, "--input", file1, "--output", file2})
+			if err := RootCmd.Execute(); err != nil {
+				t.Error(err)
+			}
+
+			defer tpm2.NVUndefineSpace(rwc, "", tpm2.HandlePlatform, tpmutil.Handle(getIndex[op.keyAlgo]))
```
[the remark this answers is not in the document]
Pranjali-2501: https://github.com/google/go-tpm-tools/pull/293#discussion_r1165782256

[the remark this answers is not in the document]
Pranjali-2501: We have discussed taking it out from the `setGCEtemplate`.

cmd/attest.go
```diff
@@ -0,0 +1,173 @@
...
+	},
+}
+
+var attestCmd = &cobra.Command{
+	Use:   "attest",
+	Short: "Create a remote attestation report",
+	Long: `Gather information for remote attestation.
+The Attestation report contains a quote on all available PCR banks, a way to validate 
+the quote, and a TCG Event Log (Linux only).
+Use --key to specify the type of attestation key. It can be gceAK for GCE attestation
```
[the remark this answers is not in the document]
Pranjali-2501: Sure , I'll remove that part from the description.

cmd/attest.go
Pranjali-2501: Changed.

cmd/attest.go
Pranjali-2501: Added.

cmd/attest.go
Pranjali-2501: Yes, but if we give an odd length nonce , then it gives an error stating `"Error: invalid argument "12345" for "--nonce" flag: encoding/hex: odd length hex string"` .
Because it try to do hex-encoding while taking input in --nonce flag.
```
func addNonceFlag(cmd *cobra.Command) {
	cmd.PersistentFlags().BytesHexVar(&nonce, "nonce", []byte{}, "hex encoded nonce for vTPM attestation, cannot be empty")
}
```

cmd/attest.go
Pranjali-2501: For now TDX and SEV-SNP have 64 byte report data feild. 
So we will add comments in cmd/attest.go stating 
" If hardware technology needs a variable length teenonce then please modify the flags description"

cmd/attest_test.go
Pranjali-2501: Thank you @deeglaze .
I have added tests for SEV-SNP in attest_test.go .

cmd/attest.go
Pranjali-2501: Agree , I have changed it in that way.

cmd/attest.go
Pranjali-2501: In `getTeeAttestationReport` , there is a check 
```
// TEEDevice can't be nil while TEENonce is non-nil
	if opts.TEENonce != nil {
		return fmt.Errorf("got non-nil TEENonce when TEEDevice is nil: %v", opts.TEENonce)
	}
```
And if the user provides a TEENonce flag , then it always fails that check because TEEDevice is nil.

cmd/attest.go
alexmwu: Thanks. Please move this to flags.go

cmd/attest_test.go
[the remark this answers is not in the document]
alexmwu: I mean to keep it in the same function `TestAttestWithGCEAK`, but closer to the call of `setGCETemplate` (e.g., move to line 267). Typically in Golang, deferred cleanups happen soon after the creation

cmd/verify_test.go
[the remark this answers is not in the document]
alexmwu: See my response above. defer should be on line 87

cmd/attest.go
alexmwu: Perhaps we should just default to using the nonce as the teenonce (and then remove the option). That's what client.Attest does if it successfully creates an SNP device but doesn't get a TEENonce. 

Or, we can add a new flag requesting the hardware type (`hwattestation`?) as part of requesting hardware attestation. And teenonce would fail without that flag.

Either way, we need to be forwards compatible with new technologies like TDX.

cmd/attest_test.go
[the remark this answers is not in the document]
alexmwu: Oh sorry, I meant we should still have a test case `{"DefaultKey", "", "rsa", "2222"},` that is different from `	{"AKWithRSA", "AK", "rsa", "2222"},`. This allows us to test that AK is indeed the default (and not gceAK).

cmd/attest_test.go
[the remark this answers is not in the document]
alexmwu: @jkl73 We create a fake metadata server so that we can still test on non-GCE environments

cmd/attest.go
deeglaze: It was my intention for teenonce to take the value of nonce if teenonce is not populated.

cmd/attest.go
Pranjali-2501: Sure

cmd/verify_test.go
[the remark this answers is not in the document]
Pranjali-2501: I have moved "defer..." closer to setGCEAKtemplate()

cmd/attest_test.go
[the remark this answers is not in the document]
Pranjali-2501: I have moved "defer..." closer to setGcCEAKtemplate()

cmd/attest_test.go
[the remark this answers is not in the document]
Pranjali-2501: I have modified the test by calling a fake_Metadata Server instead of using original MDS in `TestAttestWithGCEAK`

cmd/attest_test.go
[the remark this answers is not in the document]
Pranjali-2501: I have added a test for defaultKey as well.

cmd/attest.go
Pranjali-2501: Nonce does not need to be 48 bytes long.
Do you want me to implement a zero filling behavior if nonce is of odd length?

alexmwu: The answer here is "it depends". Specifically on the size of TPM2B_DATA, which is TPM2 implementation specific, since it also depends on the max name size. The lower bound IIRC is a 2byte hash identifier and the size of the digest for the hash function. Since TPM2s are required to support SHA256, we get a lower bound of 2+(256/8)=34. However, many TPMs support larger hash algorithms, like SHA512 which would yield 66 bytes. You can get the size by calling TPM2_GetCapability and checking TPM2_PT_MAX_DIGEST

This discussion should be moot though since we removed the nonce help text in attest in favor of the default flag help text.

alexmwu (approved):
Thanks so much! Be sure to fix the linter issues

cmd/attest_test.go
```diff
@@ -0,0 +1,374 @@
...
+		name     string
+		nonce    []byte
+		teenonce [64]byte
+	}{
+		{"gceAK:RSA", []byte{1, 2, 3, 4}, [64]byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4}},
+		{"gceAK:ECC", []byte{1, 2, 3, 4}, [64]byte{}},
+	}
+	for _, op := range tests {
+		t.Run(op.name, func(t *testing.T) {
+			if len(op.teenonce) == 0 {
```
deeglaze, line 300: If the length is zero, copying anything into it will still result in a length 0 array.
You'd need to make a nonce of the expected length, say make([]byte, abi.ReportDataSize) from go-sev-guest's abi library before copying into it.

deeglaze: But given that the test case type is [64]byte, len(op.teenonce) will never equal zero. You're just not populating the teenonce.

cmd/attest_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+			RootCmd.SetArgs([]string{"attest", "--nonce", op.nonce, "--key", "gceAK", "--algo", op.keyAlgo, "--output", secretFile1, "--format", "binarypb"})
+			if err := RootCmd.Execute(); err != nil {
+				t.Error(err)
+			}
+		})
+	}
+}
+
+func TestAttestSevSnpPass(t *testing.T) {
```
deeglaze, line 285: I'd like to see a test that the verify command works as well when the attestation includes a SevSnp report.

cmd/attest_test.go
Pranjali-2501: Yes , I have put a wrong check . It will never pass. I will remove this check.

jkl73 (approved):

PR #302, opened 2023-04-12
Add a wrapper for ExternalTPM

jkl73:
Add extTPMWrapper so the CMD package can call EventLog() in a test.

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"fmt"
 	"io"
+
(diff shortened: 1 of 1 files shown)


alexmwu (commented):
Since this is meant for #293, were you able to test `openTpm` with `client.Attest`?

jkl73:
> Since this is meant for #293, were you able to test `openTpm` with `client.Attest`?

You mean test calling `EventLog()` in a CMD command after `opentTpm()`? then yes I tested

alexmwu (approved):

PR #307, opened 2023-05-15
add tee_technology flag and test for tee_technology flag

Pranjali-2501:
Added --tee_technology flag in attest CLI that specifies the type of hardware the user wants to use for attestation.

### DIFF
--- a/cmd/attest.go
+++ b/cmd/attest.go
@@ -175,7 +175,7 @@ func addKeyFlag(cmd *cobra.Command) {
 }
 
 func addTeeTechnology(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVar(&teeTechnology, "tee_technology", "", "indicates the type of TEE hardware <sev-snp>")
+	cmd.PersistentFlags().StringVar(&teeTechnology, "tee_technology", "", "indicates the type of TEE hardware. Should be empty or sev-snp")
 }
 
 func init() {
--- a/cmd/attest.go
+++ b/cmd/attest.go
@@ -18,6 +18,11 @@ var (
 	teeTechnology string
 )
 
+const (
+	// Add constants for other devices when required
+	SevSnp = "sev-snp"
+)
+
 var attestationKeys = map[string]map[tpm2.Algorithm]func(rw io.ReadWriter) (*client.Key, error){
 	"AK": {
 		tpm2.AlgRSA: client.AttestationKeyRSA,
@@ -73,22 +78,20 @@ hardware and guarantees a fresh quote.
 		if len(teeTechnology) != 0 {
 			// Add logic to open other hardware devices when required.
 			switch teeTechnology {
-			case "sev-snp":
+			case SevSnp:
 				attestOpts.TEEDevice, err = client.CreateSevSnpDevice()
 				if err != nil {
-					return fmt.Errorf("failed to open sev-snp device: %v", err)
+					return fmt.Errorf("failed to open %s device: %v", SevSnp, err)
 				}
 			default:
 				// Change the return statement when more devices are added
-				return fmt.Errorf("tee_technology should be sev-snp")
+				return fmt.Errorf("tee_technology should be empty or %s", SevSnp)
 			}
 			if len(teeNonce) != 0 {
 				attestOpts.TEENonce = teeNonce
 			}
-		} else {
-			if len(teeNonce) != 0 {
-				return fmt.Errorf("use of --teenonce requires specifying TEE hardware type with --tee_technology")
-			}
+		} else if len(teeNonce) != 0 {
+			return fmt.Errorf("use of --teenonce requires specifying TEE hardware type with --tee_technology")
 		}
 
 		attestOpts.TCGEventLog, err = client.GetEventLog(rwc)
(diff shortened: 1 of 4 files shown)


cmd/attest.go
```diff
@@ -69,11 +70,24 @@ hardware and guarantees a fresh quote.
...
+			// Add logic to open other hardware devices when required.
+			switch teeTechnology {
+			case "sev-snp":
+				attestOpts.TEEDevice, err = client.CreateSevSnpDevice()
+				if err != nil {
+					return fmt.Errorf("failed to open sev-snp device: %v", err)
+				}
+			default:
+				// Change the return statement when more devices are added
+				return fmt.Errorf("tee_technology should be sev-snp")
```
deeglaze, line 83: should be one of ["", "sev-snp"]

cmd/attest.go
```diff
@@ -69,11 +70,24 @@ hardware and guarantees a fresh quote.
...
+			}
+			if len(teeNonce) != 0 {
+				attestOpts.TEENonce = teeNonce
+			}
+		} else {
+			if len(teeNonce) != 0 {
```
deeglaze, line 89: else if to about rightward shift

cmd/attest.go
```diff
@@ -156,6 +170,10 @@ func addKeyFlag(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&key, "key", "AK", "indicates type of attestation key to use <gceAK|AK>")
 }
 
+func addTeeTechnology(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&teeTechnology, "tee_technology", "", "indicates the type of TEE hardware <sev-snp>")
```
deeglaze, line 174: make <sev-snp> text the same as the above error message about valid values.

cmd/attest.go
Pranjali-2501: I have added a constant for the device name and changed the return statement to "tee_technology should be empty or sev-snp"

cmd/attest.go
Pranjali-2501: Instead of else and a if , I'll change it to `else if`

cmd/attest.go
Pranjali-2501: By default it is set to nil, if used then the possible value for now is `"sev-snp"` only.

Pranjali-2501: If user left the flag empty then it will say ,
`Error: flag needs an argument: --tee_technology`

Pranjali-2501: Should I change the description to `"indicates the type of TEE hardware, if set to sev-snp"`

deeglaze (approved):

cmd/attest.go
```diff
@@ -156,6 +174,10 @@ func addKeyFlag(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&key, "key", "AK", "indicates type of attestation key to use <gceAK|AK>")
 }
 
+func addTeeTechnology(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&teeTechnology, "tee_technology", "", "indicates the type of TEE hardware. Should be empty or sev-snp")
```
jkl73, line 178: Based on the current naming convention for this CLI, it should be "tee-technology"

cmd/attest.go
```diff
@@ -69,12 +76,23 @@ hardware and guarantees a fresh quote.
...
+				}
+			default:
+				// Change the return statement when more devices are added
+				return fmt.Errorf("tee_technology should be empty or %s", SevSnp)
 			}
+			if len(teeNonce) != 0 {
+				attestOpts.TEENonce = teeNonce
+			}
+		} else if len(teeNonce) != 0 {
+			return fmt.Errorf("use of --teenonce requires specifying TEE hardware type with --tee_technology")
```
jkl73, line 95: Could we change "--teenonce" to be "--tee-nonce" for the naming convention? (I understand this flag is created in this PR)

cmd/attest.go
```diff
@@ -69,12 +76,23 @@ hardware and guarantees a fresh quote.
...
 		attestOpts := client.AttestOpts{}
 		attestOpts.Nonce = nonce
-		if len(teeNonce) != 0 {
-			attestOpts.TEENonce = teeNonce
-			attestOpts.TEEDevice, err = client.CreateSevSnpDevice()
-			if err != nil {
-				return fmt.Errorf("failed to collect TEE attestation report: %v", err)
+		if len(teeTechnology) != 0 {
+			// Add logic to open other hardware devices when required.
+			switch teeTechnology {
```
jkl73, line 81: I feel like it is a bit confusing with a switch block inside an if block.
Can you just use the switch block and add the logic under `else if len(teeNonce) != 0 {` as part of the `case ""`?

cmd/attest.go
Pranjali-2501: I have changed the flag name to `--tee-technology` to make it consistent throughout the file.

cmd/attest.go
Pranjali-2501: I have changed it to `--tee-nonce`

cmd/attest.go
Pranjali-2501: As we discussed , I have used a `switch case` instead of `if else` and it is working fine with all possible cases.

jkl73 (approved):

Issue #237, opened 2022-08-11
Support loading cached key


daniel-weisse:
Is it possible to simply load a cached key without having to know extra parameters, e.g. a template matching the key?

Basically, I want a TPM client function like the following:
```Go
// LoadKey loads a key from the provided handle.
func LoadKey(rw io.ReadWriter, handle tpmutil.Handle) (*Key, error) {
        cachedPub, _, _, err := tpm2.ReadPublic(rw, handle)
        if err != nil {
                return nil, err
        }
        k := &Key{
                rw: rw,
                handle: handle,
                pubArea: cachedPub,
        }
        return k, k.finish()
}
```

The only function that does a similar thing is `NewCachedKey`, but comes with the risk of overwriting a key at the given handle if the provided template does not match the existing key.

alexmwu:
That would be useful. We probably won't get to it for some time, but I'd be happy to review this PR.

PR #313, opened 2023-05-25
Allow loading of cached keys

3u13r:
Implementation for https://github.com/google/go-tpm-tools/issues/237.

As discussed in the issue, I exposed the different sessions to be able to set it in the loaded key.
If you don't want to expose all possible sessions, exposing just the `NullSession` is sufficient for our use-case.

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -116,7 +116,7 @@ func GceAttestationKeyECC(rw io.ReadWriter) (*Key, error) {
 
 // LoadCachedKey loads a key from cachedHandle.
 // If the key is not found, an error is returned.
-// This function will no overwrite an existing key, unlike NewCachedKey.
+// This function will not overwrite an existing key, unlike NewCachedKey.
 func LoadCachedKey(rw io.ReadWriter, cachedHandle tpmutil.Handle, keySession Session) (k *Key, err error) {
 	cachedPub, _, _, err := tpm2.ReadPublic(rw, cachedHandle)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


client/keys.go
```diff
@@ -27,7 +27,7 @@ type Key struct {
 	pubArea tpm2.Public
 	pubKey  crypto.PublicKey
 	name    tpm2.Name
-	session session
+	session Session
```
alexmwu, line 30: nit: could you make this a getter?

alexmwu (commented):
Could you add tests for LoadCachedKey, e.g., loading successfully and failing the load with a bad handle?

client/keys.go
3u13r: Could you elaborate on the problem here? I exported the type / interface of `Session` in order to use it for `LoadCachedKey` as a parameter. The actual field in the struct is still unexported.

alexmwu: Not a big deal: it would just be nice if someone wanted to call LoadCachedKey with an unknown Session type. This shouldn't be the case now

client/keys.go
```diff
@@ -114,6 +114,19 @@ func GceAttestationKeyECC(rw io.ReadWriter) (*Key, error) {
 	return akEcc, nil
 }
 
+// LoadCachedKey loads a key from cachedHandle.
+// If the key is not found, an error is returned.
+// This function will no overwrite an existing key, unlike NewCachedKey.
```
alexmwu, line 119: nit: This function will *not* overwrite...

alexmwu (approved):
Please fix the linter issues

PR #314, opened 2023-05-25
Add SSH test for image.

michael-pregman:


launcher/image/test/test_ssh_manual.sh
```diff
@@ -0,0 +1,76 @@
...
+
+    # Add an SSH public key to an OS Login profile
+    gcloud compute os-login ssh-keys add --key-file=$HOME_DIR/.ssh/google_compute_engine.pub || true
+
+    echo "Sleeping so settings have time to propagate."
+    sleep 30
+
+    # SSH into VM with script
+    if [[ $IMAGE_NAME == *"debug"* ]]; then
+        if ssh -i ~/.ssh/google_compute_engine -o StrictHostKeyChecking=no $ACCOUNT_NAME@nic0.$VM_NAME.$ZONE.c.$PROJECT_NAME.internal.gcpnode.com "echo 'SSHABLE'; exit" ; then
```
jkl73, line 32: Is it possible to not use the internal address?

launcher/image/test/test_ssh_manual.sh
```diff
@@ -0,0 +1,76 @@
...
+    HOME_DIR=$(echo ~)
+    VM_NAME="cs-ssh-test-$BUILD_ID"
+    WORKLOAD_IMAGE='us-west1-docker.pkg.dev/confidential-space-images-dev/cs-integ-test-images/basic-test:latest'
+    ZONE="us-central1-a"
+
+    ACCOUNT_NAME=$(gcloud config list account --format "value(core.account)" | tr @. _)
+    PROJECT_NAME=$(gcloud config get-value project)
+
+    # Create a new VM
+    source create_vm.sh -n $VM_NAME -i $IMAGE_NAME -p $IMAGE_PROJECT -m tee-image-reference=$WORKLOAD_IMAGE,tee-container-log-redirect=true,tee-cmd=["newCmd"],tee-env-ALLOWED_OVERRIDE=overridden,enable-osconfig=TRUE -z $ZONE
```
jkl73, line 22: I think `tee-cmd=["newCmd"],tee-env-ALLOWED_OVERRIDE=overridden,enable-osconfig=TRUE` is not needed in this test?

launcher/image/test/test_ssh_manual.sh
```diff
@@ -0,0 +1,76 @@
...
+
+    # Add an SSH public key to an OS Login profile
+    gcloud compute os-login ssh-keys add --key-file=$HOME_DIR/.ssh/google_compute_engine.pub || true
```
jkl73, line 25: How is the key generated?

michael-pregman: This key is one of the defaults that is created when connecting to GCE. Since this is a manual test, I figured it is likely this key exists for those who will be running it. The alternative here is to generate and delete a key specifically for this test, which I can implement if preferred?

michael-pregman: This key is one of the defaults that is created when connecting to GCE. Since this is a manual test, I figured it is likely this key exists for those who will be running it. The alternative here is to generate and delete a key specifically for this test, which I can implement if preferred?

launcher/image/test/test_ssh_manual.sh
michael-pregman: I believe the first two options can be removed (and can do so following this) but OS Config is needed for OS Login as part of the BeyondCorp-style SSH requirements (see below comment).

launcher/image/test/test_ssh_manual.sh
michael-pregman: I was following the current BeyondCorp-style SSH guidance. Are you suggesting using direct IP access or some other path? Thanks!

jkl73: Is it possible to use "gcloud compute ssh \<instance name\>" ?

launcher/image/test/test_ssh_manual.sh
jkl73: Yes we can remove the first two.
Also I think the osconfig is disabled on the hardened image, so is this this only works on debug image?

launcher/image/test/test_ssh_manual.sh
michael-pregman: Technically yes, but only for old GCP projects. This syntax adheres to the now recommended BeyondCorp-style SSH flow, which new GCP projects require.

launcher/image/test/test_ssh_manual.sh
michael-pregman: That's correct. Though we want to keep the reference to OS Config here so we can ensure it is disabled for the hardened image even when set to be enabled. And it should work as regular for debug.

launcher/image/test/test_ssh_manual.sh
jkl73: SG

PR #315, opened 2023-06-06
Update error handling for VerifyAttestation and CreateChallenge methods

ahdavies:


launcher/verifier/rest/rest.go
```diff
@@ -83,7 +83,7 @@ func (c *restClient) CreateChallenge(_ context.Context) (*verifier.Challenge, er
 		&v1alpha1.Challenge{},
 	).Do()
 	if err != nil {
-		return nil, fmt.Errorf("calling v1alpha1.CreateChallenge: %w", err)
+		return nil, handleError(err, "Error calling v1alpha1.CreateChallenge")
```
alexmwu, line 86: these error messages will get further wrapped (in https://github.com/google/go-tpm-tools/blob/e0f2e6b3305071ba6e6a006e9687b127092f8ec1/launcher/container_runner.go#L344). They should not be capitalized: https://google.github.io/styleguide/go/decisions#error-strings

launcher/verifier/rest/rest.go
```diff
@@ -157,6 +157,14 @@ func convertRequestToREST(request verifier.VerifyAttestationRequest) *v1alpha1.V
 	}
 }
 
+func handleError(err error, message string) error {
+	const invalidArgument = "invalid argument"
+	if strings.Contains(err.Error(), invalidArgument) {
```
alexmwu, line 162: Can we instead parse the error as a *googleapi.Error and use the appropriate status code? For example, see the ProjectsLocationsGetCall.Do documentation for our v1alpha1 client https://github.com/josephlr/google-api-go-client/blob/main/confidentialcomputing/v1alpha1/confidentialcomputing-gen.go#L589-L595

and an example in our v1 client: https://github.com/googleapis/google-cloud-go/blob/main/confidentialcomputing/apiv1/confidential_computing_client.go#L613

https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

launcher/verifier/rest/rest.go
```diff
@@ -157,6 +157,14 @@ func convertRequestToREST(request verifier.VerifyAttestationRequest) *v1alpha1.V
...
+		return fmt.Errorf(message+": %w", err)
+	}
+	return fmt.Errorf(message)
```
alexmwu, line 165: I'm not sure we should completely drop non-InvalidArgument errors. An UNAVAILABLE error gives some information (we're down) although it is not actionable. Even an internal error will return an ID that we can use to help debug. This is probably where we rely mostly on the server to filter out implementation details and we can augment here based on error code

PR #318, opened 2023-06-15
use legacy tpm2 at its new path

chrisfenner:
As part of releasing go-tpm 0.9.0, we need to update go-tpm-tools to use go-tpm/tpm2 at its new legacy path (go-tpm/legacy/tpm2).

This change updates go-tpm-tools to depend on go-tpm @ 511507721cb172faf1149ab9e38ff7881eb1b776 (which adds tpmdirect but doesn't yet delete the Startup adapter), and use legacy/tpm2 instead of tpm2.


alexmwu (approved):

anatol:
What are the plans to migrate away from legacy tpm2 interface into the default one?

PR #319, opened 2023-06-16
Add GoReleaser release action for gotpm CLI

deeglaze:
The CLI binary is a useful artifact to have prebuilt and ready to use. This action ensures that all new releases that get cut will have the binary artifacts ready for download for the tableau of supported OSes and architectures.

### DIFF
--- a/.github/workflows/releaser.yaml
+++ b/.github/workflows/releaser.yaml
@@ -1,0 +1,52 @@
+name: release
+
+on:
+  push:
+    branches:
+    tags:
+      - 'v*'
+  pull_request:
+
+jobs:
+  release:
+    strategy:
+      matrix:
+        go-version: [1.20.x]
+        os: [macos-latest, ubuntu-latest]
+
+    name: Release (${{ matrix.os}}, Go ${{ matrix.go-version }})
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/releaser.yaml
```diff
@@ -0,0 +1,52 @@
...
+    branches:
+    tags:
+      - 'v*'
+  pull_request:
+
+jobs:
+  release:
+    strategy:
+      matrix:
+        go-version: [1.19.x]
```
alexmwu, line 14: can you bump to 1.20?

alexmwu (approved):

.github/workflows/releaser.yaml
deeglaze: Done.

.github/workflows/releaser.yaml
```diff
@@ -0,0 +1,52 @@
...
+    tags:
+      - 'v*'
+  pull_request:
+
+jobs:
+  release:
+    strategy:
+      matrix:
+        go-version: [1.20.x]
+        os: [macos-latest, ubuntu-latest]
```
jkl73, line 15: Can we add Windows here?

Also since mac doesn't have TPM, the usage will be limited.

deeglaze: Given the ci.yaml has "todo: get this working on windows-latest", I'll leave that for a future PR from Alex.

jkl73: sgtm

PR #328, opened 2023-07-05
[Signature discovery] Add a signature interface and a library to parse signature from image manifest

yawangwang:
Adding a signature interface and a cosign library that implements the interface to parse cosign-generated signature from OCI image manifest.


launcher/internal/oci/interface.go
```diff
@@ -0,0 +1,59 @@
...
+// SigningAlgorithm is a specific type for string constants used for sigature signing and verification.
+type SigningAlgorithm string
+
+const (
+	// RsassaPssSha256 is RSASSA-PSS with a SHA256 digest supported for cosign sign.
+	RsassaPssSha256 SigningAlgorithm = "RSASSA_PSS_SHA256"
+	// RsassaPkcs1v15Sha256 is RSASSA-PKCS1 v1.5 with a SHA256 digest supported for cosign sign.
+	RsassaPkcs1v15Sha256 SigningAlgorithm = "RSASSA_PKCS1V15_SHA256"
+	// EcdsaP256Sha256 is ECDSA on the P-256 Curve with a SHA256 digest supported for cosign sign.
+	EcdsaP256Sha256 SigningAlgorithm = "ECDSA_P256_SHA256"
```
haydentherapper, line 18: ED25519 is a possibility too.

launcher/internal/oci/interface.go
```diff
@@ -0,0 +1,59 @@
+// Package oci contains functionalities to interact with OCI image signatures.
+// https://github.com/opencontainers/image-spec/tree/main#readme.
+package oci
+
+import (
+	"go.uber.org/multierr"
+)
+
+// SigningAlgorithm is a specific type for string constants used for sigature signing and verification.
+type SigningAlgorithm string
```
haydentherapper, line 10: Do you want to support only this set of signing algorithms? Users may provide KMS keys with another algorithm. For example, here are all the supported GCP KMS algorithms - https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/gcp/client.go#L51-L62

launcher/internal/oci/cosign/signature.go
```diff
@@ -0,0 +1,64 @@
...
+		return "", errors.New("cosign signature not found in the layer annotations")
+	}
+	if _, err := encoding.DecodeString(sig); err != nil {
+		return "", fmt.Errorf("invalid base64 encoded signature: %w", err)
+	}
+	return sig, nil
+}
+
+// PublicKey implements oci.Signature interface.
+// Since public key is attached to the `optional` field of payload, we don't actually implement this method.
```
haydentherapper, line 53: To confirm, given this is under `internal`, this is not meant to be used by users, correct? Just wondering if these should be implemented.

launcher/internal/signaturediscovery/client.go
```diff
@@ -37,6 +41,36 @@ func (c *Client) FetchSignedImageManifest(ctx context.Context, targetRepository
 	return getManifest(ctx, image)
 }
 
+// FetchImageSignatures returns a list of valid image signatures associated with the target OCI image.
+func (c *Client) FetchImageSignatures(ctx context.Context, targetRepository string) ([]oci.Signature, error) {
```
haydentherapper, line 45: This looks good, but I don't have a lot of experience with the OCI implementation in Cosign. To confirm this is working as expected, I would either manually compare to what Cosign generates, or use this library to verify what Cosign generates.

launcher/internal/oci/interface.go
yawangwang: `ED25519` is not supported for the first launch of this feature as designed, see go/signed-image-impl-design

launcher/internal/oci/interface.go
yawangwang: Yes, only RSA_SSA, RSA_PKCS1, and ECDSA_P256 will be supported, see `signing algorithm support` section on the design go/signed-image-impl-design

launcher/internal/oci/interface.go
```diff
@@ -0,0 +1,59 @@
...
+
+const (
+	// RsassaPssSha256 is RSASSA-PSS with a SHA256 digest supported for cosign sign.
+	RsassaPssSha256 SigningAlgorithm = "RSASSA_PSS_SHA256"
```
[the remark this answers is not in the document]
yawangwang: The current implementations use Tink to verify signatures. Since Tink only accepts at least 2048 bit key, the key size here should be at least 2048 bit long. The `SigningAlgorithm` will be used by Tink to determine which signing algorithm to use for verification.

launcher/internal/oci/cosign/signature.go
yawangwang: Correct. The confidential VM will be the only caller of this method, so we want to make it usable by other packages such as `signaturediscovery` in the library, but NOT export it to users of the library. For the current implementations since the container launcher will send trusted payload with public key attached directly to the service for verification, this method is intended as not implemented.

launcher/internal/signaturediscovery/client.go
yawangwang: Done, see tests here https://github.com/google/go-tpm-tools/pull/328/files#diff-fc42aca0ab2213e764bddeab995045792c9d118a151ea446c7ad03fe21b1444fR61.

launcher/internal/oci/interface.go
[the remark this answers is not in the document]
haydentherapper: It might be worth documenting this, since I would normally expect that key strength is specified as part of the signature algorithm.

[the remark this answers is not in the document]
haydentherapper: Whoops, nvm, saw that was added!

launcher/internal/oci/interface.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"go.uber.org/multierr"
+)
+
+// SigningAlgorithm is a specific type for string constants used for sigature signing and verification.
+type SigningAlgorithm string
+
+const (
+	// RsassaPss2048Sha256 is RSASSA-PSS 2048 bit key with a SHA256 digest supported for cosign sign.
+	RsassaPss2048Sha256 SigningAlgorithm = "RSASSA_PSS_2048_SHA256"
```
haydentherapper, line 14: nit: I believe golang style is to capitalize acronyms, so RSASSAPSS2048SHA256 (as painfully loud as that looks haha)

launcher/internal/signaturediscovery/client.go
haydentherapper: Looks great, thanks!

launcher/internal/oci/interface.go
yawangwang: Done.

jkl73: I think in golang we do prefer the CamelCase. Like in the standard library https://cs.opensource.google/go/go/+/master:src/io/io.go;l=22 and in this repo as well.

jkl73: Ah I see they are acronyms, then all caps should be good.

launcher/internal/oci/cosign/signature.go
```diff
@@ -0,0 +1,64 @@
...
+
+var (
+	// Verify that our Sig struct implements the expected public interface.
+	_        oci.Signature = Sig{}
+	encoding               = base64.StdEncoding
+)
+
+// Payload implements oci.Signature interface.
+func (s Sig) Payload() ([]byte, error) {
+	if digest.FromBytes(s.Blob) != s.Layer.Digest {
```
jkl73, line 34: Comparing digests, probably a good idea to use:  https://pkg.go.dev/crypto/subtle#ConstantTimeCompare

launcher/internal/oci/cosign/signature.go
```diff
@@ -0,0 +1,64 @@
...
+		return "", fmt.Errorf("invalid base64 encoded signature: %w", err)
+	}
+	return sig, nil
+}
+
+// PublicKey implements oci.Signature interface.
+// Since public key is attached to the `optional` field of payload, we don't actually implement this method.
+// Instead we send payload directly to the Attestation service and let the service parse the payload.
+func (s Sig) PublicKey() ([]byte, error) {
+	return nil, nil
```
jkl73, line 56: here and below, can return a fmt.Error("not implemented")

launcher/internal/oci/interface.go
```diff
@@ -0,0 +1,71 @@
...
+	// RSASSAPKCS1V153072SHA256 is RSASSA-PKCS1 v1.5 3072 bit key with a SHA256 digest supported for cosign sign.
+	RSASSAPKCS1V153072SHA256 SigningAlgorithm = "RSASSA_PKCS1V15_3072_SHA256"
+	// RSASSAPKCS1V154096SHA256 is RSASSA-PKCS1 v1.5 4096 bit key with a SHA256 digest supported for cosign sign.
+	RSASSAPKCS1V154096SHA256 SigningAlgorithm = "RSASSA_PKCS1V15_4096_SHA256"
+	// ECDSAP256SHA256 is ECDSA on the P-256 Curve with a SHA256 digest supported for cosign sign.
+	ECDSAP256SHA256 SigningAlgorithm = "ECDSA_P256_SHA256"
+)
+
+// Signature represents a single OCI image signature.
+type Signature interface {
```
jkl73, line 30: Why do we need this interface (like where is this defined externally)?
Are we going to have other implementation other than cosign?

jkl73 (commented):
lgtm

launcher/internal/oci/interface.go
yawangwang: There is no other external interface. The intention to define this interface was to align with the API protos we defined for ContainerImageSignature (see https://source.corp.google.com/piper///depot/google3/google/cloud/confidentialcomputing/v1main/service.proto;rcl=551011596;l=315). This interface is more generic and compatible for non-cosign generated signatures if we decide to have implementations in the future.

launcher/internal/signaturediscovery/client_test.go
```diff
@@ -48,17 +48,46 @@ func TestFormatSigTag(t *testing.T) {
...
+		}
+	}
+}
+
+func createTestClient(t *testing.T, originalImageDesc v1.Descriptor) *Client {
+	t.Helper()
+
+	containerdClient, err := containerd.New(defaults.DefaultAddress)
+	if err != nil {
+		t.Skipf("test needs containerd daemon: %v", err)
```
alexmwu, line 89: (this can be a followup) Can we test this in CI https://github.com/google/go-tpm-tools/blob/master/.github/workflows/ci.yml

launcher/internal/signaturediscovery/client_test.go
```diff
@@ -48,17 +48,46 @@ func TestFormatSigTag(t *testing.T) {
...
+	if len(signatures) == 0 {
+		t.Errorf("no image signatures found for the original image %v", originalImageDesc)
+	}
+	for _, sig := range signatures {
+		if _, err := sig.Payload(); err != nil {
+			t.Errorf("Payload() failed: %v", err)
+		}
+		if _, err := sig.Base64Encoded(); err != nil {
+			t.Errorf("Base64Encoded() failed: %v", err)
+		}
```
alexmwu, line 80: Can you also validate the signatures?

alexmwu (changes requested):

launcher/internal/oci/interface.go
```diff
@@ -0,0 +1,45 @@
+// Package oci contains functionalities to interact with OCI image signatures.
+// https://github.com/opencontainers/image-spec/tree/main#readme.
+package oci
+
+// SigningAlgorithm is a specific type for string constants used for sigature signing and verification.
+type SigningAlgorithm string
+
+const (
+	// RSASSAPSS2048SHA256 is RSASSA-PSS 2048 bit key with a SHA256 digest supported for cosign sign.
+	RSASSAPSS2048SHA256 SigningAlgorithm = "RSASSA_PSS_2048_SHA256"
```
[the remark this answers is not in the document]
yawangwang: No, our API enum does not specify bit length for a RSA public key. Since the current implementations use Tink to verify signatures and Tink only accepts at least 2048 bit key, I think it would be good to imply this length restriction from the naming. But to align with our API enum, we can keep the naming but change the value to `RSASSA_PSS_SHA256`. Done.

launcher/internal/oci/cosign/signature.go
```diff
@@ -0,0 +1,64 @@
...
+	"encoding/base64"
+	"errors"
+	"fmt"
+
+	"github.com/google/go-tpm-tools/launcher/internal/oci"
+	"github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Sig implements oci.Signature interface for cosign-gernated signatures.
```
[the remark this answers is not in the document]
yawangwang: Done.

launcher/internal/oci/cosign/signature.go
```diff
@@ -0,0 +1,64 @@
...
+type Sig struct {
+	// Layer represents a layer descriptor for OCI image manifest.
+	Layer v1.Descriptor
+	// Blob represents the opaque data uploaded to OCI registory associated with the layer.
```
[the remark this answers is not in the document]
yawangwang: Done.

launcher/internal/signaturediscovery/client_test.go
yawangwang: For this specific test, it has to be run under root permission. Will add a new entry to CI to run this test under root only in my followup PR.

launcher/internal/oci/cosign/signature.go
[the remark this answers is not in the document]
yawangwang: Done.

launcher/internal/signaturediscovery/client_test.go
yawangwang: Done.

launcher/internal/oci/cosign/signature.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+
+	"github.com/google/go-tpm-tools/launcher/internal/oci"
+	"github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// Sig implements oci.Signature interface for cosign-gernated signatures.
+type Sig struct {
+	// Layer represents a layer descriptor for OCI image manifest.
```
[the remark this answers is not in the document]
yawangwang: Done.

launcher/internal/oci/cosign/signature.go
yawangwang: Thanks for pointing me to the materials related to timing attack. I agree with your point because the layer digest is public visible from OCI image manifest so directly comparing these two values should be ok. Added more comments here.

launcher/internal/signaturediscovery/client_test.go
```diff
@@ -48,17 +49,55 @@ func TestFormatSigTag(t *testing.T) {
...
+	client := createTestClient(t, originalImageDesc)
 	// testing image manifest fetching using a public docker repo url
 	if _, err := client.FetchSignedImageManifest(ctx, targetRepository); err != nil {
 		t.Errorf("failed to fetch signed image manifest from targetRepository [%s]: %v", targetRepository, err)
 	}
 }
+
+func TestFetchImageSignaturesDockerPublic(t *testing.T) {
+	ctx := namespaces.WithNamespace(context.Background(), "test")
+	originalImageDesc := v1.Descriptor{Digest: "sha256:905a0f3b3d6d0fb37bfa448b9e78f833b73f0b19fc97fed821a09cf49e255df1"}
```
alexmwu, line 64: Does the codelab project make this repo public? Otherwise, the test will fail for someone without permissions.

launcher/internal/oci/interface.go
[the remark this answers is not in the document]
alexmwu: I actually was wondering about the other direction: should we update/change our API enum? It's confusing to put the key size here but not in the API

alexmwu (approved):

launcher/internal/signaturediscovery/client_test.go
yawangwang: Yes, this repo only grants roles/artifactregistry.reader permission to all users, so it is a public accessible repo.

PR #336, opened 2023-07-17
 Integrate go-tdx-guest with go-tpm-tools.

jrjatin:
When getting a vTPM quote within a TDX VM, the attestation service will utilize go-tdx-guest to generate and verify Intel's TDX hardware based attestation quote.

> Add TDX attestation protobuf in the go-tpm-tools attestation message.
> Update client changes for TDX device-related attestation quote generation.
> Update server changes for TDX quote verification utilizing go-tdx-guest.
> Enhance CLI tool for TDX functionality updates.


server/verify.go
```diff
@@ -347,6 +348,25 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
...
 			return fmt.Errorf("TEE attestation is %v, expected an SevSnpAttestation", attestation.GetTeeAttestation())
 		}
+	case pb.GCEConfidentialTechnology_Intel_TDX:
+		switch tee := attestation.GetTeeAttestation().(type) {
+		case *pb.Attestation_TdxAttestation:
+			var tdxOpts *VerifyTdxOpts
+			if opts.TEEOpts == nil {
+				tdxOpts = TdxDefaultOptions()
+			} else {
+				switch teeopts := opts.TEEOpts.(type) {
```
alexmwu, line 358:
```
tdxOpts, ok := opts.TEEOpts.(*VerifyTdxOpts)
if !ok {
  return fmt.Errorf("unexpected value for TEEOpts given a TDX attestation quote: %v", opts.TEEOpts)
}
```

server/verify.go
```diff
@@ -347,6 +348,25 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
...
+				case *VerifyTdxOpts:
+					tdxOpts = teeopts
+				default:
+					return fmt.Errorf("unexpected value for TEEOpts given a TDX attestation quote: %v",
+						opts.TEEOpts)
+				}
+			}
+			return VerifyTdxAttestation(tee.TdxAttestation, tdxOpts)
+		default:
+			return fmt.Errorf("TEE attestation is %v, expected an TdxAttestation", attestation.GetTeeAttestation())
```
alexmwu, line 368: "... expected __a__ TdxAttestation..."

server/verify_test.go
```diff
@@ -1065,3 +1117,139 @@ func TestVerifyAttestationWithSevSnp(t *testing.T) {
 		})
 	}
 }
+
+func TestVerifyAttestationWithTdx(t *testing.T) {
+
+	pcr0 := uint32(0)
+	algorithms := []struct {
+		ID         uint16
+		DigestSize uint16
```
alexmwu, line 1126: you don't need DigestSize here. You can just call hash.Size(): https://pkg.go.dev/hash#Hash

server/verify_test.go
```diff
@@ -1065,3 +1117,139 @@ func TestVerifyAttestationWithSevSnp(t *testing.T) {
...
+		Make       func() hash.Hash
+	}{
+		{ID: 0x04, DigestSize: 0x14, Make: crypto.SHA1.New},
```
alexmwu, line 1129: please use the go-tpm constants: https://github.com/google/go-tpm/blob/e9722e4450de281f46cb7f4241f59695de4ed998/tpm2/constants.go#L27-L28

alexmwu: You should only need to pass in the go-tpm algo. You can convert to a `crypto.Hash` using https://github.com/google/go-tpm/blob/e9722e4450de281f46cb7f4241f59695de4ed998/tpm2/structures.go#L442

server/verify_test.go
```diff
@@ -1065,3 +1117,139 @@ func TestVerifyAttestationWithSevSnp(t *testing.T) {
 		})
 	}
 }
+
+func TestVerifyAttestationWithTdx(t *testing.T) {
```
alexmwu, line 1121: AIUI, you are trying to write a TPM2 event log out below. Consider separating these out into helper functions, as that really helps the readability.

alexmwu (changes requested):

client/attest_test.go
```diff
@@ -291,3 +294,81 @@ func TestSevSnpDevice(t *testing.T) {
...
+		name           string
+		opts           AttestOpts
+		wantReportData [64]byte
+		wantErr        string
+	}{
+		{
+			name: "Happy case no nonce",
+			opts: AttestOpts{
+				Nonce:            someNonce,
+				CertChainFetcher: localClient,
```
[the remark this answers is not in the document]
jrjatin: Done

cmd/verify.go
```diff
@@ -47,16 +48,24 @@ var debugCmd = &cobra.Command{
...
+		switch attestation.GetTeeAttestation().(type) {
+		case *pb.Attestation_TdxAttestation:
+			validateOpts = &server.VerifyTdxOpts{
+				Verification: tv.DefaultOptions(),
 			}
-		} else {
-			validateOpts = &server.VerifySnpOpts{
-				Validation:   server.SevSnpDefaultValidateOpts(nonce),
-				Verification: &sv.Options{},
+		default:
```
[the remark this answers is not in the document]
jrjatin: Added a case for sev-snp attestation.
Since verifyGceTechnology does not return an error when attestation contains GCEConfidentialTechnology_NONE, I assume that the default case should not return an error. Please correct me if it's incorrect.

proto/attest.proto
```diff
@@ -51,6 +53,7 @@ enum GCEConfidentialTechnology {
   AMD_SEV_ES = 2;
   // Enum value 3 is reserved.
   AMD_SEV_SNP = 4;
+  Intel_TDX = 5;
```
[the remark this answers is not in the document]
jrjatin: Changed to INTEL_TDX = 3

cmd/attest.go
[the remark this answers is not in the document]
jrjatin: Added a testcase similar to the one written for sev-snp device.

server/verify.go
jrjatin: Done

server/verify.go
```diff
@@ -347,6 +348,25 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
 		default:
 			return fmt.Errorf("TEE attestation is %v, expected an SevSnpAttestation", attestation.GetTeeAttestation())
 		}
+	case pb.GCEConfidentialTechnology_Intel_TDX:
+		switch tee := attestation.GetTeeAttestation().(type) {
+		case *pb.Attestation_TdxAttestation:
```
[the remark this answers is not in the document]
jrjatin: Removed switch statements for TDX related attestation.

server/verify_test.go
jrjatin: Done

server/verify.go
jrjatin: Done

cmd/verify.go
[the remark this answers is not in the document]
jrjatin: Need some help related to testing the changes in verify CLI.
Since the verify command needs an attestation file as an input, which contains tdx hardware attestation, I need guidance on how to create a test file for this purpose. Also, I am unable to find a reference test written for sev-snp hardware attestation testing.

client/attest.go
```diff
@@ -140,6 +142,12 @@ type SevSnpDevice struct {
 	Device sg.Device
 }
 
+// TdxDevice encapsulates the TDX attestation device to add its attestation quote
```
deeglaze, line 145: I don't think you need a wrapper type if you're not adding extra information. Your functions can just take a Device argument. Then you don't need to have Open/Close wrapper functions.

server/verify_test.go
```diff
@@ -270,6 +348,54 @@ func TestVerifyBasicAttestation(t *testing.T) {
...
+	defer client.CheckedClose(t, rwc)
+
+	ak, err := client.AttestationKeyRSA(rwc)
+	if err != nil {
+		t.Fatalf("failed to generate AK: %v", err)
+	}
+	defer ak.Close()
+
+	nonce := []byte("super secret nonce")
+	attestation, err := ak.Attest(client.AttestOpts{
```
deeglaze, line 362: The test says it uses Tdx, but you're not providing a test TDX device.

cmd/verify.go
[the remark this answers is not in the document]
alexmwu: Sorry you're right. validateOpts should be nil by default.

I don't see the case handling the SNP attestation. As the code is currently written, we always try to validate with a VerifySnpOpts, even if there is no TeeAttestation in the proto.

server/verify_test.go
```diff
@@ -270,6 +348,54 @@ func TestVerifyBasicAttestation(t *testing.T) {
...
+	if _, err := VerifyAttestation(attestation, VerifyOpts{
+		Nonce:      nonce,
+		TrustedAKs: []crypto.PublicKey{ak.PublicKey()},
+	}); err != nil {
+		t.Errorf("failed to verify: %v", err)
+	}
+
+	if _, err := VerifyAttestation(attestation, VerifyOpts{
+		Nonce:   nonce,
+		TEEOpts: teeopts,
```
deeglaze, line 381: The fact that this doesn't error makes me think your test attestation doesn't include a NonHostInfo event that claims to use TDX. To avoid this kind of miss, I'd recommend that VerifyAttestation fails in VerifyTechnology when TEEOpts is non-null and the technology doesn't match the type / there is no technology event.

server/verify_test.go
```diff
@@ -36,6 +41,79 @@ import (
 
 var measuredHashes = []crypto.Hash{crypto.SHA1, crypto.SHA256}
 
+func createTpm2EventLog(deviceEnum byte) []byte {
```
alexmwu, line 44: this should take the GCEConfidentialTechnology enum type.

server/verify_test.go
```diff
@@ -1065,3 +1191,62 @@ func TestVerifyAttestationWithSevSnp(t *testing.T) {
...
+		t.Fatalf("failed to attest: %v", err)
+	}
+
+	type testCase struct {
+		name    string
+		opts    VerifyOpts
+		wantErr string
+	}
+	tcs := []testCase{
+		{
```
alexmwu, line 1233: there should be failure tests (e.g., bad key, bad cert, bad nonce, etc.)

server/verify_test.go
alexmwu: Also, please refactor `TestVerifyAttestationWithSevSnp` so that it uses this function.

jkl73, line 44: This function is very similar to the beginning of `TestVerifyAttestationWithSevSnp` , try to reuse the code there as well.

server/verify_test.go
jkl73, line 1: This file is very long. If possible try to break down the test file (maybe to verify_tdx_test.go , verify_sev_test.go and verify_test.go?)

server/verify.go
```diff
@@ -345,8 +346,24 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
...
 		default:
-			return fmt.Errorf("TEE attestation is %v, expected an SevSnpAttestation", attestation.GetTeeAttestation())
+			return fmt.Errorf("TEE attestation is %v, expected a SevSnpAttestation", attestation.GetTeeAttestation())
 		}
+	case pb.GCEConfidentialTechnology_INTEL_TDX:
+		var tdxOpts *VerifyTdxOpts
+		var ok bool
+		tee, ok := attestation.TeeAttestation.(*pb.Attestation_TdxAttestation)
+		if !ok {
+			return fmt.Errorf("TEE attestation is %v, expected a TdxAttestation", attestation.GetTeeAttestation())
```
jkl73, line 356: I assume you want to log the type here? Try to use %T in the Errorf string

client/attest.go
```diff
@@ -183,6 +191,49 @@ func (d *SevSnpDevice) Close() error {
...
+
+// AddAttestation will get the TDX attestation quote given opts.TEENonce
+// and add them to `attestation`. If opts.TEENonce is empty, then uses
+// contents of opts.Nonce.
+func (d *TdxDevice) AddAttestation(attestation *pb.Attestation, opts AttestOpts) error {
+	var tdxNonce [tabi.TdReportDataSize]byte
+	if len(opts.TEENonce) == 0 {
+		copy(tdxNonce[:], opts.Nonce[:])
+	} else if len(opts.TEENonce) != tabi.TdReportDataSize {
+		return fmt.Errorf("the TEENonce size is %d. Intel TDX device requires 64", len(opts.TEENonce))
```
jkl73, line 212: Nit: don't hardcode 64, use tabi.TdReportDataSize in the message

cmd/attest_test.go
```diff
@@ -326,3 +328,38 @@ func TestAttestTeeNonceFail(t *testing.T) {
...
+	attestopts := client.AttestOpts{
+		Nonce:     []byte{1, 2, 3, 4},
+		TEENonce:  []byte{1, 2, 3, 4},
+		TEEDevice: &client.TdxDevice{Device: tdxTestDevice},
+	}
+	_, err = ak.Attest(attestopts)
+	if err == nil {
+		t.Error("expected non-nil error")
+	}
+
```
jkl73, line 364: nit: extra line

jkl73 (commented):
lgtm (go-tpm-tools side)

client/attest.go
jrjatin: I tried to keep the code uniform in the file, similar to SevSnpDevice struct, I have created TdxDevice struct. Should I remove it?

cmd/attest_test.go
jrjatin: Done

client/attest.go
jrjatin: Done

server/verify.go
jrjatin: Done

server/verify_test.go
jrjatin: Done

server/verify_test.go
jrjatin: Added a failure testcase for wrong TDX Attestation. Since go-tdx-guest has pending validate changes, I will add a testcase related to badNonce in the later PR.

server/verify_test.go
jrjatin: Done

server/verify_test.go
jrjatin: Added TDX device.

server/verify_test.go
jrjatin: I will raise a bug for this change and this will be handled in another PR.

cmd/verify.go
```diff
@@ -47,18 +48,29 @@ var debugCmd = &cobra.Command{
...
-		var validateOpts *server.VerifySnpOpts
-		if len(teeNonce) != 0 {
-			validateOpts = &server.VerifySnpOpts{
-				Validation:   server.SevSnpDefaultValidateOpts(teeNonce),
-				Verification: &sv.Options{},
+
+		var validateOpts interface{}
+		switch attestation.GetTeeAttestation().(type) {
+		case *pb.Attestation_TdxAttestation:
+			validateOpts = &server.VerifyTdxOpts{
```
alexmwu, line 55: I may be missing something about TDX attestation here or go-tdx-guest, but where and how does the nonce get verified?

server/verify_test.go
deeglaze: I think each of NONE, SEV, and SEV_ES should not simply return nil, but also check that opts.TEEOpts is nil. That would match my expectation.

client/attest.go
deeglaze: Nah, leave it in I suppose.

Pranjali-2501: Okay.

PR #338, opened 2023-07-19
Fix logging blocking issue

alexmwu:
Add additional options to "tee-container-log-redirect" to allow redirection to Cloud Logging and serial console (and configure them).

true maps to both serial and Cloud Logging
serial and cloud_logging map to the respective destinations
false does not emit workload logs
   
This also removes serial logging for the workload by-default, since performance issues occur when a high throughput workload emits a large number of logs. Users need to specify "tee-enable-serial-log=true" or "tee-enable-serial-log=serial" 
See https://github.com/systemd/systemd/issues/2815 as an example.


launcher/launcher/main.go
```diff
@@ -102,18 +75,18 @@ func main() {
 
 	defer func() {
 		// catch panic, will also output to cloud logging if possible
-		if r := recover(); r != nil {
-			logger.Println("Panic:", r)
-			exitCode = 2
-		}
+		// if r := recover(); r != nil {
```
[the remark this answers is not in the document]
alexmwu: I think it was an accident. Done.

launcher/spec/launch_spec.go
```diff
@@ -30,6 +29,29 @@ const (
...
 
+// LogRedirectLocation specifies the workload logging redirect location.
+type LogRedirectLocation string
+
+func (l LogRedirectLocation) isValid() error {
+	switch l {
+	case Everywhere, CloudLogging, Serial, Nowhere:
+		return nil
+	}
+	return fmt.Errorf("invalid logging redirect location: %s", l)
```
jkl73, line 40: Can add all valid redirect option strings in the error message

launcher/launcher/main.go
```diff
@@ -152,7 +125,7 @@ func getExitCode(isHardened bool, restartPolicy spec.RestartPolicy, err error) i
 	return exitCode
 }
 
-func startLauncher() error {
+func startLauncher(launchSpec spec.LaunchSpec) error {
```
jkl73, line 128: Can serialConsole pass in here like the launchSpce here instead of a global variable?

jkl73 (approved):

PR #343, opened 2023-08-02
[Signature discovery] Integrate signature discovery client into attestation agent.

yawangwang:
This PR includes the following changes:
1. Added `containerImageSignaturesFetcher` to the attestation agent such that the agent can fetch any new container image signatures on refreshToken. 
2. Integrated the signature discovery client into signature fetcher and made corresponding changes to `container_runner`.
3. Add specific tests that need root permission to run to ci.yaml.
4. Integrate client experimental flag to gate the signed container feature.


launcher/container_runner.go
```diff
@@ -236,14 +266,38 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 		return nil, fmt.Errorf("failed to create REST verifier client: %v", err)
 	}
 
+	// Fetch container image signatures by using signaturediscovery client.
+	signaturesFetcher := func(ctx context.Context) []internal.Signature {
+		authToken, err := RetrieveAuthToken(mdsClient)
+		if err != nil {
+			logger.Printf("failed to retrieve auth token: %v, using empty auth for image pulling\n", err)
+		}
+		updatedSpec, err := spec.GetLaunchSpec(mdsClient)
```
alexmwu, line 275: Why do you need to update the spec? The launcher should only ever get the spec at startup. Refetching it can lead to weird states and potential security issues.

launcher/container_runner.go
```diff
@@ -236,14 +266,38 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 		return nil, fmt.Errorf("failed to create REST verifier client: %v", err)
 	}
 
+	// Fetch container image signatures by using signaturediscovery client.
```
alexmwu, line 269: If you don't update the spec, this func should be unnecessary. Which is good because this is tough to follow.

launcher/container_runner.go
```diff
@@ -79,6 +82,33 @@ const (
 	defaultRefreshJitter = 0.1
 )
 
+func fetchContainerImageSignatures(ctx context.Context, sdClient *signaturediscovery.Client, targetRepos []string, logger *log.Logger) []internal.Signature {
```
alexmwu, line 85: not required, but it would be nice to be able to cache signatures so we don't have to fetch every time. This complicates things, so maybe add a TODO with an issue.

alexmwu (commented):
IMO, this is the crucial bit of this feature and should go behind an experiment. Please do not submit until we have that functionality.

launcher/container_runner.go
yawangwang: The rational behind updating the spec is to enable operators don't have to restart VM when new docker repos added. Even though ADMC team is ok with restarting VM when new repos added, I feel it might be nice to have this feature. But if refetching it will lead to any potential security issues, we should definitely avoid that.

yawangwang:
> IMO, this is the crucial bit of this feature and should go behind an experiment. Please do not submit until we have that functionality.

Agree, this PR should also go behind this PR for updating v1 SDK https://github.com/google/go-tpm-tools/pull/305

yawangwang:
> > IMO, this is the crucial bit of this feature and should go behind an experiment. Please do not submit until we have that functionality.
> 
> Agree, this PR should also go behind this PR for updating v1 SDK #305

Now since V1 SDK has been merged into master, this PR is now ready for review again with client experimental flag integration.

launcher/container_runner_test.go
```diff
@@ -414,6 +417,71 @@ type idTokenResp struct {
 	Token string `json:"token"`
 }
 
+func TestFetchContainerImageSignatures(t *testing.T) {
+	ctx := namespaces.WithNamespace(context.Background(), "test")
```
alexmwu, line 421: We should consider using something like https://pkg.go.dev/github.com/docker/docker/testutil/registry#V2 or the mock https://pkg.go.dev/github.com/docker/docker/testutil/registry#Mock instead of actually making a network call in our unit tests

launcher/container_runner_test.go
```diff
@@ -414,6 +417,71 @@ type idTokenResp struct {
...
+			targetRepos: []string{"us-docker.pkg.dev/vegas-codelab-5/cosign-test/base", "us-docker.pkg.dev/vegas-codelab-5/cosign-test/base"},
+			wantLen:     2,
+			wantBase64Sigs: []string{
+				// Check signatures from the OCI image manifest at https://pantheon.corp.google.com/artifacts/docker/vegas-codelab-5/us/cosign-test/base/sha256:1febaa6ac3a5c095435d5276755fb8efcb7f029fefe85cd9bf3ec7de91685b9f;tab=manifest?project=vegas-codelab-5.
+				"MEUCIQDgoiwMiVl1SAI1iePhH6Oeqztms3IwNtN+w0P92HTqQgIgKjJNcHEy0Ep4g4MH1Vd0gAHvbwH9ahD+jlnMP/rXSGE=",
+				"MEUCIQDgoiwMiVl1SAI1iePhH6Oeqztms3IwNtN+w0P92HTqQgIgKjJNcHEy0Ep4g4MH1Vd0gAHvbwH9ahD+jlnMP/rXSGE=",
+			},
+		},
+		{
+			name:        "fetchContainerImageSignatures success with nil target repos",
```
alexmwu, line 440: Can we add a case for a nonexistent repo and one that exists but doesn't have signatures?

launcher/container_runner.go
```diff
@@ -230,15 +259,29 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 		return nil, fmt.Errorf("failed to create REST verifier client: %v", err)
 	}
 
+	// Fetch container image signatures with specific repositories.
+	containerSignaturesFetcher := func(targetRepos []string) []internal.Signature {
+		sdClient := getSignatureDiscoveryClient(cdClient, token, image.Target())
```
alexmwu, line 264: are we getting a new signaturediscovery Client every time we call agent.Attest? Can we make this a closure to avoid this? Alternatively, we can pass the signature discovery client in the agent and use a fake for testing.

launcher/agent/agent.go
```diff
@@ -79,14 +91,27 @@ func (a *agent) Attest(ctx context.Context) ([]byte, error) {
...
+
+	if a.launchSpec.Experiments.EnableSignedContainerImage {
+		signatures := a.sigsFetcher(a.launchSpec.SignedImageRepos)
+		if len(signatures) > 0 {
+			req.ContainerImageSignatures = signatures
+			a.logger.Printf("Found container image signatures: %v\n", signatures)
+		}
+	}
+
+	resp, err := a.client.VerifyAttestation(ctx, req)
```
alexmwu, line 108: Can you add unit test coverage here? The fake verifier client should expect to see signatures.

alexmwu (approved):

PR #347, opened 2023-08-18
Integration of go-tpm-tools with go-tdx-guest

Pranjali-2501:
When getting a vTPM quote within a TDX VM, the attestation service will utilize go-tdx-guest to generate and verify Intel's TDX hardware based attestation quote.

Add TDX attestation protobuf in the go-tpm-tools attestation message.
Update client changes for TDX device-related attestation quote generation.
Update server changes for TDX quote verification using go-tdx-guest.
Enhance CLI tool for TDX functionality updates.


proto/attest.proto
```diff
@@ -4,6 +4,7 @@ package attest;
 
 import "proto/sevsnp.proto";
 import "tpm.proto";
+import "proto/tdx.proto";
```
deeglaze, line 7: nit: keep imports sorted.

cmd/verify.go
```diff
@@ -47,17 +48,27 @@ var debugCmd = &cobra.Command{
...
-		if len(teeNonce) != 0 {
-			validateOpts = &server.VerifySnpOpts{
-				Validation:   server.SevSnpDefaultValidateOpts(teeNonce),
-				Verification: &sv.Options{},
+
+		var validateOpts interface{}
+		switch attestation.GetTeeAttestation().(type) {
+		case *pb.Attestation_TdxAttestation:
+			validateOpts = &server.VerifyTdxOpts{
+				Verification: tv.DefaultOptions(),
```
deeglaze, line 56: Do we not have a teeNonce check for TDX?

go.mod
```diff
@@ -8,19 +8,20 @@ require (
 	github.com/google/go-sev-guest v0.7.0
 	github.com/google/go-tpm v0.9.0
 	github.com/google/logger v1.1.1
-	google.golang.org/protobuf v1.28.0
+	google.golang.org/protobuf v1.31.0
```
deeglaze, line 11: You'll need to either update the protoc-gen-go version in .github/workflows/ci.yaml (I'm not sure if there's a github action available at this version) or re-generate these with the existing version. To do that, you'll need to have a local install at the old version and prefer its /bin location in your PATH environment variable.

server/verify.go
```diff
@@ -329,24 +330,36 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
 	case pb.GCEConfidentialTechnology_AMD_SEV_ES: // Not verifiable on GCE
 		return nil
```
deeglaze, line 331: return an error if opts is non-nil for all NONE, SEV, and SEV_ES cases. Add a test case for that expecting to verify SEV-SNP on SEV is an error.

deeglaze (changes requested):

cmd/verify.go
Pranjali-2501: go-tdx-guest/validate is not in MVP, so we didn't integrate validate in this PR.
And validate package have the teeNonce check .

go.mod
Pranjali-2501: Only maintainers can modify the ci.yaml file.
I have changed the version of protoc-gen-go to 1.28.0 .

server/verify.go
```diff
@@ -323,30 +324,51 @@ func makePool(certs []*x509.Certificate) *x509.CertPool {
 func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTechnology, opts *VerifyOpts) error {
 	switch tech {
 	case pb.GCEConfidentialTechnology_NONE: // Nothing to verify
+		if opts.TEEOpts != nil {
+			return fmt.Errorf("TEEopts is non-nil")
```
deeglaze, line 328: You're not adding any formatting here, but I think you should. The tech is in an event that should be surfaced as the reason this is an error, e.g., fmt.Errorf("memory encryption technology %v does not support TEEOpts", tech)

Pranjali-2501: Done.

client/attest.go
```diff
@@ -183,6 +191,49 @@ func (d *SevSnpDevice) Close() error {
...
+	return &TdxDevice{Device: d}, nil
+}
+
+// AddAttestation will get the TDX attestation quote given opts.TEENonce
+// and add them to `attestation`. If opts.TEENonce is empty, then uses
+// contents of opts.Nonce.
+func (d *TdxDevice) AddAttestation(attestation *pb.Attestation, opts AttestOpts) error {
+	var tdxNonce [tabi.TdReportDataSize]byte
+	if len(opts.TEENonce) == 0 {
+		copy(tdxNonce[:], opts.Nonce[:])
```
jkl73, line 210: opts.Nonce is already a slice, so I think [:] is not needed

Also do we care about the size of opts.Nonce (what if it's zero)?

proto/gen_attest.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-protoc -I. -I`go list -m -f "{{.Dir}}" github.com/google/go-sev-guest` --go_out=. --go_opt=module=github.com/google/go-tpm-tools/proto attest.proto
+protoc -I. -I`go list -m -f "{{.Dir}}" github.com/google/go-sev-guest` -I`go list -m -f "{{.Dir}}" github.com/google/go-tdx-guest` --go_out=. --go_opt=module=github.com/google/go-tpm-tools/proto attest.proto
```
jkl73, line 3: nit: newline EOF

go.mod
jkl73: It's still showing v1.31.0 as the version, have you changed it back to v1.28.0? the header in proto/attest/attest.pb.go is showing v1.28.0

You can either update both to 1.31, or just remain 1.28

client/attest.go
Pranjali-2501: Nonce can not be zero. Check for this is already there 
```
if len(opts.Nonce) == 0 {
	return nil, fmt.Errorf("provided nonce must not be empty")
}
```
nonce can be of any size. For teenonce the size is fixed.

Pranjali-2501: Yes, slicing is not required. I have changed it.

proto/gen_attest.sh
Pranjali-2501: There is no new line at the EOF.
Do you want me to add a blank line?

go.mod
Pranjali-2501: Done.

proto/gen_attest.sh
jkl73: yes, because I think the original file has a new line at the end

client/attest.go
jkl73: Okay, looks a little bit weird to me because the tdxNonce can be only the first part of the opt.Nonce (may result not enough entropy for the nonce?). I'm not sure maybe @deeglaze and @alexmwu can confirm if it's okay.

jkl73 (commented):
lgtm

client/attest.go
Pranjali-2501: I have tested the implementation and it is working fine.

And for `copy(tdxNonce[:], opts.Nonce[:])` , if length of nonce is less that 64 then it will concatenate zeros at the end.

proto/gen_attest.sh
Pranjali-2501: Done

server/verify.go
Pranjali-2501: IIUC, it will detect the hardware technology enabled on the cvm by itself only.

client/attest.go
Pranjali-2501: If the user does not provide a teenonce, then the nonce will be used as the teenonce, either concatenated with zeroes or truncated to make it 64 bytes long.

But If teenonce is present, then it must be 64 bytes in length.

jkl73 (approved):

PR #348, opened 2023-08-24
Add experiment support

JoshuaKrstic:
Add experiment support.
This change:
  - pulls the experiment binary from a GCS bucket (not open-source-able)
  - gets metadata from the MDS for the experiment binary
  - calls experiment binary with metadata
  - uses new experiment data in container-runner.go


launcher/image/sync_experiments.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
```
alexmwu, line 2: you should `set -euo pipefail` here

launcher/container_runner.go
```diff
@@ -49,13 +50,15 @@ type ContainerRunner struct {
 }
 
 const (
-	// hostTokenPath defined the directory in the host that will store attestation tokens
-	hostTokenPath = "/tmp/container_launcher/"
+	// tmpContainerLauncherPath defined the directory in the host that will store attestation tokens and experiment data.
```
alexmwu, line 53: please do this rename in a separate PR, ideally putting it in a const package. @jkl73 was planning on doing this.

launcher/container_runner.go
```diff
@@ -505,6 +508,16 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to fetch and write OIDC token: %v", err)
 	}
 
+	e := experiments.NewExperiments(r.logger)
```
alexmwu, line 511: this callpoint should probably go somewhere in `main.go` and you should put the experiment flag settings in the LaunchSpec.

launcher/experiments/experiments.go
```diff
@@ -0,0 +1,78 @@
+// Package experiments contains functionalities to retrieve syned experiments
```
alexmwu, line 1: "Package experiments contains functionalities to retrieve synced experiments"

launcher/experiments/experiments.go
```diff
@@ -0,0 +1,78 @@
...
+	"io"
+	"log"
+	"os"
+)
+
+// Experiments represents a collection of experiments that were synced.
+// All failures during the reading of experiment data are logged and returned.
+// This struct intends to be as safe as possible. Errors default to experiments being off, or not
+// existing. Failures should not cascade from this struct.
+// When using this class, if an experiment is not found, it should be assumed it is off.
```
alexmwu, line 16: I don't think this line is needed, since we return an error from GetExperiment.

launcher/experiments/experiments.go
```diff
@@ -0,0 +1,78 @@
...
+type Experiments struct {
+	experiments map[string]any
+	logger      *log.Logger
+}
+
+// NewExperiments returns an experiments.
+func NewExperiments(logger *log.Logger) Experiments {
```
alexmwu, line 23: Name this "New" go/go-style/decisions#repetitive-with-package

launcher/experiments/experiments.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// ReadJSONInput  takes a reader and unmarshalls the contents into the experiments map.
+// If the unmarsahlling fails, the experiments map is set to an empty map.
+func (e *Experiments) ReadJSONInput(read io.Reader) error {
```
alexmwu, line 62: This doesn't need to be exported.

launcher/experiments/experiments.go
```diff
@@ -0,0 +1,78 @@
...
+		return nil, fmt.Errorf("experiment not found for key %s", key)
+	}
+
+	return val, nil
+}
+
+// ReadExperimentsFile takes a filepath, opens the file, and calls ReadJsonInput with the contents
+// of the file.
+// If the file cannot be opened, the experiments map is set to an empty map.
+func (e *Experiments) ReadExperimentsFile(fpath string) error {
```
alexmwu, line 43: Why doesn't the `New` function take the reader and do the initialization? This function seems unnecessarily exported.

launcher/experiments/experiments.go
alexmwu, line 1: This can probably go in the internal directory.

launcher/image/preload.sh
```diff
@@ -2,11 +2,19 @@
...
+readonly EXPERIMENTS_SCRIPT="experiments_sync"
+readonly GCS_BUCKET="gs://confidential-space-images_third-party_test"
 
 copy_launcher() {
   cp launcher "${CS_PATH}/cs_container_launcher"
 }
 
+copy_exeriment_client() {
+  gcloud storage cp ${GCS_BUCKET} "${CS_PATH}/${EXPERIMENTS_SCRIPT}"
+  chmod +x "${CS_PATH}/${EXPERIMENTS_SCRIPT}"
```
alexmwu, line 14: I think this may be why the script isn't running: it should go after the cp.

alexmwu (changes requested):

launcher/image/sync_experiments.sh
```diff
@@ -0,0 +1,33 @@
...
+readonly OEM_PATH='/usr/share/oem'
+readonly CS_PATH="${OEM_PATH}/confidential_space"
+readonly EXPERIMENTS_SCRIPT="experiments_sync"
+readonly SYNC_RESULT_FILE="/tmp/container_launcher/experiment_data"
+
+getImageDetails() {
+    # image will look like `projects/confidential-space-images/global/images/confidential-space-hardened-230600`
+    #                   or `projects/confidential-space-images/global/images/confidential-space-debug-230600`
```
alexmwu, line 10: Using the image name won't work for images created by our CI. It's not a big issue since we can copy an image over, but it is more work. Is there a reason we didn't go with something like build time?

launcher/image/preload.sh
```diff
@@ -2,11 +2,19 @@
...
 readonly OEM_PATH='/usr/share/oem'
 readonly CS_PATH="${OEM_PATH}/confidential_space"
+readonly EXPERIMENTS_SCRIPT="experiments_sync"
+readonly GCS_BUCKET="gs://confidential-space-images_third-party_test"
 
 copy_launcher() {
   cp launcher "${CS_PATH}/cs_container_launcher"
 }
 
+copy_exeriment_client() {
```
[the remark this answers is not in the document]
JoshuaKrstic: Thanks for catching, fixed.

launcher/experiments/experiments.go
JoshuaKrstic: Done

launcher/experiments/experiments.go
JoshuaKrstic: I wanted this class to contain the logic for opening the experiment file to be able to easily suppress errors.
To be able to test this code, I needed to pass in a reader.

This was my solution to accomplish both: I can test the reader method and consumers don't have to convert the file input into a reader to use this class.

launcher/image/preload.sh
JoshuaKrstic: Fixed this previously, thanks for catching this.

launcher/image/sync_experiments.sh
JoshuaKrstic: Yes I did notice this in my testing.
We should talk about this offline because there's no perfect solution here.

launcher/experiments/experiments.go
alexmwu: You can still test in the same package (`experiments_test.go`) without exporting. `New` can take a filepath and then call `readExperimentsFile`.

JoshuaKrstic: Gotcha, made the change.

launcher/image/sync_experiments.sh
JoshuaKrstic: Will change how our presubmit images are generated as we talked about offline.

PR #352, opened 2023-09-06
Add experiment support

JoshuaKrstic:
Add experiment support.
This change:

- pulls the experiment binary from a GCS bucket (not open-sourceable code)
- gets metadata from the MDS for the experiment binary (image family, image version, git commit hash for presubmit images)
- calls experiment binary with metadata fetched
  - the binary fetches the project id + vm instance ID from the metadata server
- provides experiment data as a map in the launch spec
- provides helper functions to make the launch spec map type-safe

### DIFF
--- a/launcher/image/preload.sh
+++ b/launcher/image/preload.sh
@@ -2,9 +2,16 @@
 
 readonly OEM_PATH='/usr/share/oem'
 readonly CS_PATH="${OEM_PATH}/confidential_space"
+readonly EXPERIMENTS_BINARY="confidential_space_experiments"
 
 copy_launcher() {
   cp launcher "${CS_PATH}/cs_container_launcher"
+}
+
+copy_experiment_client() {
+  # DownloadExpBinary creates the file at EXPERIMENTS_BINARY.
+  cp $EXPERIMENTS_BINARY "${CS_PATH}/${EXPERIMENTS_BINARY}"
+  chmod +x "${CS_PATH}/${EXPERIMENTS_BINARY}"
 }
 
 setup_launcher_systemd_unit() {
@@ -90,6 +97,8 @@ main() {
 
   # Install container launcher entrypoint.
   configure_entrypoint "entrypoint.sh"
+  # Install experiment client.
+  copy_experiment_client
   # Install container launcher.
   copy_launcher
   setup_launcher_systemd_unit
--- a/launcher/internal/experiments/experiments.go
+++ b/launcher/internal/experiments/experiments.go
@@ -1,0 +1,42 @@
+// Package experiments contains functionalities to retrieve synced experiments
+package experiments
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+// Experiments contains the experiments flags this version of the launcher expects to receive.
+// Failure to unmarshal the experiment JSON data will result in an empty object being returned
+// to treat experiment flags as their default value. The error should still be checked.
+type Experiments struct {
+	EnableTestFeatureForImage  bool
+	EnableSignedContainerImage bool
+}
+
+// New takes a filepath, opens the file, and calls ReadJsonInput with the contents
+// of the file.
+// If the file cannot be opened, the experiments map is set to an empty map.
+func New(fpath string) (Experiments, error) {
+	f, err := os.ReadFile(fpath)
+	if err != nil {
+		// Return default values on failure.
+		return Experiments{}, err
+	}
+
+	r, err := readJSONInput(f)
+
+	return r, err
+}
+
+// ReadJSONInput  takes a reader and unmarshals the contents into the experiments map.
+// If the unmarsahlling fails, the experiments map is set to an empty map.
+func readJSONInput(b []byte) (Experiments, error) {
+	var experiments Experiments
+	if err := json.Unmarshal(b, &experiments); err != nil {
+		// Return default values on failure.
+		return Experiments{}, fmt.Errorf("failed to unmarshal json: %w", err)
+	}
+	return experiments, nil
+}
--- a/launcher/internal/experiments/experiments_test.go
+++ b/launcher/internal/experiments/experiments_test.go
@@ -1,0 +1,52 @@
+package experiments
+
+import (
+	"testing"
+)
+
+func TestExperiments(t *testing.T) {
+	tests := []struct {
+		input string
+	}{
+		{input: "{\"EnableTestFeatureForImage\":true,\"EnableSignedContainerImage\":true}"},
+		{input: "{\"EnableTestFeatureForImage\":true,\"EnableSignedContainerImage\":true,\"FloatFeature\":-5.6,\"OtherTestFeatureForImage\":false}"},
+	}
+
+	for i, test := range tests {
+		e, err := readJSONInput([]byte(test.input))
+
+		if err != nil {
+			t.Errorf("testcase %d: failed to create experiments object: %v", i, err)
+		}
+
+		if e.EnableTestFeatureForImage == false {
+			t.Errorf("testcase %d: expected EnableTestFeatureForImage to be true, got false", i)
+		}
+
+		if e.EnableSignedContainerImage == false {
+			t.Errorf("testcase %d: expected EnableSignedContainerImage to be true, got false", i)
+		}
+	}
+}
+
+func TestExperimentsBadJson(t *testing.T) {
+	tests := []struct {
+		input string
+	}{
+		{input: "{\"EnableTestFeatureForImage\":true,\"EnableSignedContainerImage\":true"},
+		{input: "{}"},
+		{input: ""},
+	}
+
+	for i, test := range tests {
+		e, _ := readJSONInput([]byte(test.input))
+
+		if e.EnableTestFeatureForImage == true {
+			t.Errorf("testcase %d: expected EnableTestFeatureForImage to be false, got true", i)
+		}
+
+		if e.EnableSignedContainerImage == true {
+			t.Errorf("testcase %d: expected EnableSignedContainerImage to be false, got true", i)
+		}
+	}
+}
--- a/launcher/launcher/main.go
+++ b/launcher/launcher/main.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path"
 	"regexp"
 	"strings"
 
@@ -18,6 +19,8 @@ import (
 	"github.com/containerd/containerd/namespaces"
 	"github.com/google/go-tpm-tools/client"
 	"github.com/google/go-tpm-tools/launcher"
+	"github.com/google/go-tpm-tools/launcher/internal/experiments"
+	"github.com/google/go-tpm-tools/launcher/launcherfile"
 	"github.com/google/go-tpm-tools/launcher/spec"
 	"github.com/google/go-tpm/legacy/tpm2"
 )
@@ -28,6 +31,10 @@ const (
 	// panic() returns 2
 	rebootRC = 3 // reboot
 	holdRC   = 4 // hold
+	// experimentDataFile defines where the experiment sync output data is expected to be.
+	experimentDataFile = "experiment_data"
+	// binaryPath contains the path to the experiments binary.
+	binaryPath = "/usr/share/oem/confidential_space/confidential_space_experiments"
 )
 
 var rcMessage = map[int]string{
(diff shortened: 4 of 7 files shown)


launcher/image/sync_experiments.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+set -euo pipefail
```
JoshuaKrstic, line 2: We should consider failing silently given that the experiments server has a lower slo (99%) than us.
We shouldn't block the launcher if experiment sync fails. The experiment class is resilient to the lack of a file + lack of expected flags.

alexmwu: This is generally good practice. You otherwise don't have error handling here  for exceptional cases.

If you want to avoid failure, you can wrap it in an if statement and log the error or `|| some_failure_logging_func`. If you think bash is not the best language, consider writing your script in Python or another better scripting language.

launcher/image/sync_experiments.sh
```diff
@@ -0,0 +1,41 @@
...
+readonly CS_PATH="${OEM_PATH}/confidential_space"
+readonly EXPERIMENTS_BINARY="confidential_space_experiments"
+readonly SYNC_RESULT_DIRECTORY="/tmp/container_launcher"
+readonly SYNC_RESULT_FILE="experiment_data"
+
+getImageDetails() {
+    # image will look like `projects/confidential-space-images/global/images/confidential-space-hardened-230600`
+    #                   or `projects/confidential-space-images/global/images/confidential-space-debug-230600`
+    #                   or `projects/confidential-space-images/global/images/confidential-space-debug-0-presubmit-abc123`
+    image=$(curl --silent "http://metadata.google.internal/computeMetadata/v1/instance/image" -H "Metadata-Flavor: Google")
```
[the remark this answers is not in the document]
alexmwu: That's a capital S, aka `--show-error` in case it's not clear

launcher/image/sync_experiments.sh
```diff
@@ -0,0 +1,41 @@
...
+}
+
+main() {
+    getImageDetails results
+    if [ ! -d "$SYNC_RESULT_DIRECTORY" ]; then
+        echo "$SYNC_RESULT_DIRECTORY does not exist, creating it."
+        mkdir $SYNC_RESULT_DIRECTORY
+    fi
+   
+    echo "calling $CS_PATH/$EXPERIMENTS_BINARY with flags: output=$SYNC_RESULT_DIRECTORY/$SYNC_RESULT_FILE, image_type=\"${results[0]}\" image_version=\"${results[1]}\" -image_build_hash=\"${results[2]}\""
```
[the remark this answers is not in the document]
alexmwu: https://pkg.go.dev/os/exec#Cmd.Run. This would make testing easier to implement since we can drop in a mock_experiments_binary.

internal/experiments/experiments.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return readJSONInput(f)
+}
+
+// ReadJSONInput  takes a reader and unmarshalls the contents into the experiments map.
+// If the unmarsahlling fails, the experiments map is set to an empty map.
+func readJSONInput(read io.Reader) (map[string]any, error) {
+	jsondata := make(map[string]any)
+	if err := json.NewDecoder(read).Decode(&jsondata); err != nil {
```
alexmwu, line 92: I think we should write our own struct if we're gonna be calling something like `GetBoolExperiment` anyway. The default decoder is not strict (although you definitely need a unit test) and will not unmarshal/fail on unknown fields.

Then caller code just inspects `launchSpec.Experiments.TestFeatureForImage`, which is a bool.

launcher/image/preload.sh
```diff
@@ -2,11 +2,21 @@
 
 readonly OEM_PATH='/usr/share/oem'
 readonly CS_PATH="${OEM_PATH}/confidential_space"
+readonly EXPERIMENTS_BINARY="confidential_space_experiments"
+readonly GCS_BUCKET="gs://confidential-space-images_third-party"
```
alexmwu, line 6: This is no longer used

launcher/launcher/main.go
```diff
@@ -28,6 +30,10 @@ const (
 	// panic() returns 2
 	rebootRC = 3 // reboot
 	holdRC   = 4 // hold
+	// hostTokenPath defined the directory in the host that will store attestation tokens and experiment data.
+	hostTokenPath = "/tmp/container_launcher/"
```
alexmwu, line 34: This is duplicated with container_runner.go. Any reason we don't pull it out into some constants or common package

launcher/image/sync_experiments.sh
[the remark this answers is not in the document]
JoshuaKrstic: This will be removed and put into the binary instead.

internal/experiments/experiments.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return f, true
+}
+
+// ReadExperimentsFile takes a filepath, opens the file, and calls ReadJsonInput with the contents
+// of the file.
+// If the file cannot be opened, the experiments map is set to an empty map.
+func ReadExperimentsFile(fpath string) (map[string]any, error) {
+	f, err := os.Open(fpath)
```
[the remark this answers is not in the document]
JoshuaKrstic: thanks for catching this.

launcher/launcher/main.go
```diff
@@ -72,6 +78,13 @@ func main() {
 		return
 	}
 
+	e, err := experiments.ReadExperimentsFile(path.Join(hostTokenPath, experimentDataFile))
+	if err != nil {
+		logger.Println(err)
```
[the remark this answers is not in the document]
JoshuaKrstic: done.

launcher/image/sync_experiments.sh
JoshuaKrstic: This script has been removed entirely.

launcher/image/sync_experiments.sh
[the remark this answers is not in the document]
JoshuaKrstic: Done

internal/experiments/experiments.go
```diff
@@ -0,0 +1,97 @@
...
+	"io"
+	"os"
+)
+
+// GetBooleanExperiment attempts to read the key from the map provided and does type checking on the
+// returned value, if any. Value is false on failure.
+func GetBooleanExperiment(e map[string]any, key string) (bool, bool) {
+	v, err := e[key]
+	if !err {
+		return false, false
```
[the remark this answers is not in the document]
JoshuaKrstic: Done.

internal/experiments/experiments.go
JoshuaKrstic: Removed in favor of generics.

internal/experiments/experiments_test.go
```diff
@@ -0,0 +1,251 @@
...
+	e, err := readJSONInput(&buffer)
+	if err != nil {
+		t.Errorf("Failed to parse input json, err: %v", err)
+	}
+
+	val, ok := GetStringExperiment(e, "FloatFeature")
+	if ok {
+		t.Errorf("expected to fail to get string experiment, got %v", val)
+	}
+}
```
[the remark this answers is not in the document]
JoshuaKrstic: Love these, thanks for the tip.

launcher/launcher/main.go
alexmwu: please rebase off of https://github.com/google/go-tpm-tools/pull/356

internal/experiments/experiments.go
alexmwu: I think this is a better approach since the type checking is in one place instead of spread around. Can we do this instead of generics? Sorry if I was unclear; I meant to remove the generics comments

JoshuaKrstic: Gotcha, that certainly clears things up. Made the change.

launcher/internal/experiments/experiments.go
```diff
@@ -0,0 +1,42 @@
...
+	return r, err
+}
+
+// ReadJSONInput  takes a reader and unmarshalls the contents into the experiments map.
+// If the unmarsahlling fails, the experiments map is set to an empty map.
+func readJSONInput(b []byte) (Experiments, error) {
+	var experiments Experiments
+	if err := json.Unmarshal(b, &experiments); err != nil {
+		// Return default values on failure.
+		return Experiments{}, fmt.Errorf("failed to unmarshall json: %w", err)
```
alexmwu, line 39: nit: here and elsewhere "unmarshall" -> "unmarshal"

launcher/image/preload.sh
```diff
@@ -2,11 +2,18 @@
...
 readonly OEM_PATH='/usr/share/oem'
 readonly CS_PATH="${OEM_PATH}/confidential_space"
+readonly EXPERIMENTS_BINARY="confidential_space_experiments"
 
 copy_launcher() {
   cp launcher "${CS_PATH}/cs_container_launcher"
 }
 
+copy_experiment_client() {
+  # EXPERIMENTS_BINARY is placed at  by the `gcloud` step in image/cloudbuild.yaml
```
alexmwu, line 12: nit: DownloadExpBinary creates the file at EXPERIMENTS_BINARY.

launcher/internal/experiments/experiments_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"testing"
+)
+
+func TestExperiments(t *testing.T) {
+	tests := []struct {
+		input string
+	}{
+		{input: "{\"EnableTestFeatureForImage\":true,\"EnableSignedContainerImage\":true}"},
```
alexmwu, line 11: Add a test for invalid JSON (e.g., missing a closing bracket).

launcher/internal/experiments/experiments_test.go
```diff
@@ -0,0 +1,51 @@
+package experiments
+
+import (
+	"testing"
+)
+
```
alexmwu, line 6: Add a test for empty string and empty JSON "{}".

launcher/launcher/main.go
```diff
@@ -72,6 +79,16 @@ func main() {
 		return
 	}
 
+	args := fmt.Sprintf("-output=%s", experimentDataFile)
+	exec.Command(binaryPath, args).Run()
```
alexmwu, line 83: You should also catch the err from Run()

```
if err != nil {
    logger.Printf("failed to fetch experiments: %v\n", err)
}
```

launcher/launcher/main.go
```diff
@@ -72,6 +79,16 @@ func main() {
...
+
+	e, err := experiments.New(path.Join(launcherfile.HostTmpPath, experimentDataFile))
+	if err != nil {
+		logger.Println("failed to read experiment file %w", err)
```
alexmwu, line 87: If you are using the a format verb, you need to use logger.Printf. The format verb %w is typically used when returning an error however, so you don't need to wrap the err. Just use %v or:
`logger.Println("Failed to read experiment file: ", err)`

launcher/launcher/main.go
```diff
@@ -72,6 +79,16 @@ func main() {
 		return
 	}
 
+	args := fmt.Sprintf("-output=%s", experimentDataFile)
```
alexmwu, line 82: Does output assume a prefix of `launcherfile.HostTmpPath`? Or do you need to `path.Join` with `experimentDataFile` here? (if so, you can pull out the path.Join below into a variable)

alexmwu (changes requested):

launcher/internal/experiments/experiments_test.go
JoshuaKrstic: This is in the next test case.

launcher/internal/experiments/experiments_test.go
JoshuaKrstic: Added the empty string, the empty bracket is in the next test case.

launcher/launcher/main.go
JoshuaKrstic: Gotcha, thanks for the tip. I think the println is an artifact from previous statement, thanks for catching that,

launcher/launcher/main.go
JoshuaKrstic: oopsie, it does need the combo. I appear to have rushed this commit :)

alexmwu (approved):

PR #363, opened 2023-09-18
Make launcher host tmp directory before experiment fetch

JoshuaKrstic:
This PR:

Updates main.go to create the directory where experiments data is saved for the script before it is called.
Fixes experiment-related log messages to be in the correct format

### DIFF
--- a/launcher/launcher/main.go
+++ b/launcher/launcher/main.go
@@ -79,17 +79,20 @@ func main() {
 		return
 	}
 
+	if err := os.MkdirAll(launcherfile.HostTmpPath, 0744); err != nil {
+		logger.Printf("failed to create %s: %v", launcherfile.HostTmpPath, err)
+	}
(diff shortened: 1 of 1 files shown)


launcher/launcher/main.go
```diff
@@ -79,17 +79,21 @@ func main() {
 		return
 	}
 
-	experimentsFile := path.Join(launcherfile.HostTmpPath, experimentDataFile)
+	err = os.MkdirAll(launcherfile.HostExperimentsPath, os.ModePerm)
```
alexmwu, line 82: I think your error is that you don't create the directory first. It's created in https://github.com/google/go-tpm-tools/blob/master/launcher/container_runner.go#L404.
So pull that mkdir here.

JoshuaKrstic: Done

alexmwu (approved):

PR #368, opened 2023-09-22
[cmd] Add new command genToken in the CLI tool

Ruide:
Command Description: Fetch an attestation report from GCE VM vTPM and send it to Google Attestation Service for an OIDC token. 

This command improves usability for a GCE VM user.

### DIFF
--- a/cmd/gen_token.go
+++ b/cmd/gen_token.go
@@ -18,25 +18,11 @@ import (
 	"github.com/google/go-tpm-tools/launcher/agent"
 	"github.com/google/go-tpm-tools/launcher/verifier"
 	"github.com/google/go-tpm-tools/launcher/verifier/rest"
+	"github.com/google/go-tpm/legacy/tpm2"
 	"github.com/spf13/cobra"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/option"
 )
-
-const (
-	successRC = 0 // workload successful (no reboot)
-	failRC    = 1 // workload or launcher internal failed (no reboot)
-	// panic() returns 2
-	rebootRC = 3 // reboot
-	holdRC   = 4 // hold
-)
-
-var rcMessage = map[int]string{
-	successRC: "workload finished successfully, shutting down the VM",
-	failRC:    "workload or launcher error, shutting down the VM",
-	rebootRC:  "rebooting VM",
-	holdRC:    "VM remains running",
-}
 
 var logger *log.Logger
 var mdsClient *metadata.Client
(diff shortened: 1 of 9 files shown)


Ruide:
@jkl73 @kongoshuu Could you take a look?

jkl73 (commented):
I'm still reviewing, one overall comment I have is:
Can we not include MDS as a dependency of CMD package? Since this is command line tool, the caller should be responsible for getting the token (using curl or other tools). And then pass the token as a file/parameter/stdin into our cmd tool. This should also make the testing easier (no mocking metadata server anymore)

Ruide:
> I'm still reviewing, one overall comment I have is: Can we not include MDS as a dependency of CMD package? Since this is command line tool, the caller should be responsible for getting the token (using curl or other tools). And then pass the token as a file/parameter/stdin into our cmd tool. This should also make the testing easier (no mocking metadata server anymore)

By the token, do you mean the GCP ID token? If so, it would require the customers to know how to fetch the token themselves. And also the MDS client retrieves ProjectID and Region information for the customers. So without the MDS dependency, the customers would need to know how to supply the GCP ID token, ProjectID and Region information to the cmd tool themselves. From the perspective of usability, and under the assumption that customer would use this command in a GCE CVM environment, I think it might be worthwhile for us to keep the dependency of MDS. What do you think?

jkl73:
> > I'm still reviewing, one overall comment I have is: Can we not include MDS as a dependency of CMD package? Since this is command line tool, the caller should be responsible for getting the token (using curl or other tools). And then pass the token as a file/parameter/stdin into our cmd tool. This should also make the testing easier (no mocking metadata server anymore)
> 
> By the token, do you mean the GCP ID token? If so, it would require the customers to know how to fetch the token themselves. And also the MDS client retrieves ProjectID and Region information for the customers. So without the MDS dependency, the customers would need to know how to supply the GCP ID token, ProjectID and Region information to the cmd tool themselves. From the perspective of usability, and under the assumption that customer would use this command in a GCE CVM environment, I think it might be worthwhile for us to keep the dependency of MDS. What do you think?

Yes, I mean the ID token (inside the principalfetcher). I still think the CMD tool should be lightweight (I know attest.go also use MDS, but I think that can also be removed).
Since it's a command line tool, the whole token generation can take several steps/commands (we can provide sample curl commands to fetch all token/regions, and pass those as parameters to the tool), this also make it more flexible (allow user to choose a different region). But let me try to run the cmd myself first.

Ruide:
> > > I'm still reviewing, one overall comment I have is: Can we not include MDS as a dependency of CMD package? Since this is command line tool, the caller should be responsible for getting the token (using curl or other tools). And then pass the token as a file/parameter/stdin into our cmd tool. This should also make the testing easier (no mocking metadata server anymore)
> > 
> > 
> > By the token, do you mean the GCP ID token? If so, it would require the customers to know how to fetch the token themselves. And also the MDS client retrieves ProjectID and Region information for the customers. So without the MDS dependency, the customers would need to know how to supply the GCP ID token, ProjectID and Region information to the cmd tool themselves. From the perspective of usability, and under the assumption that customer would use this command in a GCE CVM environment, I think it might be worthwhile for us to keep the dependency of MDS. What do you think?
> 
> Yes, I mean the ID token (inside the principalfetcher). I still think the CMD tool should be lightweight (I know attest.go also use MDS, but I think that can also be removed). Since it's a command line tool, the whole token generation can take several steps/commands (we can provide sample curl commands to fetch all token/regions, and pass those as parameters to the tool), this also make it more flexible (allow user to choose a different region). But let me try to run the cmd myself first.

Yea, I agree with the more flexible part. I think we might want to decide on the common scenario where users use this command. If users most likely use this command in a GCE CVM environment and fetch a token to prove to other parties/components the environment they are located at, then including MDS client makes sense. I could not picture a scenario where users might want to use this command for other purpose. Could you provide some other use cases? Also, since the command requires a GCP shielded VM vtpm device, and what the command could do is to prove one GCP CVM environment, I consider MDS client would normally exist?

cmd/gen_token.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		return nil
+	},
+}
+
+// getRESTClient returns a REST verifier.Client that points to the given address.
+// It defaults to the Attestation Verifier instance at
+// https://confidentialcomputing.googleapis.com.
+func getRESTClient(ctx context.Context, asAddr string, ProjectID string, Region string) (verifier.Client, error) {
```
jkl73, line 158: This is some repetitive code from "go-tpm-tools/launcher/container_runner.go"

Ideally we need to refactor the code to avoid this, but for now at least add some comment or TODO to make them point to each other. (Same for the principalFetcher)

jkl73:
> rties/components the environment they are located at, then including MDS client makes sense. I could not picture a scenario where users might want to use this command for other purpose. Could you provide some other use cases? Also, since the command requires a GCP

I see, for ease of use, I think it's fine to use MDS for now. We may want to change the behavior later (if we want to run this command on other than a gce VM).

PR #369, opened 2023-09-29
Changed name of Tdx Test in verify_test.go

Pranjali-2501:
This test should be skipped when snp staging tests are triggered.


alexmwu (commented):
I don't follow how this PR will ensure SNP staging tests skip TDX.

PR #373, opened 2023-10-13
Import latest version of go-tdx-guest

jrjatin:
The latest version of go-tdx-guest fixes the logger error that was displayed when using the CLI for verification.


jkl73 (approved):
LGTM, fyi we recently change the branch name from "master" to "main"

PR #374, opened 2023-10-13
Container launcher prints signature details instead of signature object

yawangwang:
From https://github.com/google/go-tpm-tools/blob/main/launcher/agent/agent.go#L105 https://screenshot.googleplex.com/7wxmTBPQubwZLoB, the launcher prints golang objects for any found signatures, we should print signature details instead.

### DIFF
--- a/launcher/internal/oci/cosign/signature.go
+++ b/launcher/internal/oci/cosign/signature.go
@@ -21,6 +21,8 @@ type Sig struct {
 	// Blob represents the opaque data uploaded to OCI registry associated with the layer.
 	// This contains the Simple Signing Payload as described in https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#tag-based-discovery.
 	Blob []byte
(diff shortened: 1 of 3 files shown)


launcher/internal/oci/cosign/signature.go
```diff
@@ -68,3 +70,12 @@ func (s Sig) PublicKey() ([]byte, error) {
 func (s Sig) SigningAlgorithm() (oci.SigningAlgorithm, error) {
 	return "", fmt.Errorf("not implemented")
 }
+
+// String returns signature details
+func (s Sig) String() string {
```
alexmwu, line 75: Please add a unit test.

alexmwu (changes requested):

launcher/internal/oci/cosign/signature.go
yawangwang: Done.

jkl73 (approved):
lgtm

alexmwu (approved):

PR #378, opened 2023-10-24
Add image tests for the experiments binary

JoshuaKrstic:
Adds image tests for the experiments binary

Tests added:
- VM started, LaunchSpec has the TestFeature enabled

Tests still needed:
- VM started, binary is replaced with a failing binary, LaunchSpec has the TestFeature disabled
- VM started, binary is replaced with binary that writes random experiments, LaunchSpec has the TestFeature disabled

### DIFF
--- a/launcher/image/test/scripts/test_experiment_value.sh
+++ b/launcher/image/test/scripts/test_experiment_value.sh
@@ -1,0 +1,21 @@
+ #!/bin/bash
+set -euo pipefail
+source util/read_serial.sh
+
+# This test requires the workload to run and print
+# corresponding messages to the serial console.
+SERIAL_OUTPUT=$(read_serial $2 $3) 
+print_serial=false
+
+if echo $SERIAL_OUTPUT | grep -q "EnableTestFeatureForImage:$1"
+then
+    echo "- test experiment verified $1"
+else
+    echo "FAILED: experiment status expected to be $1"
+    echo "TEST FAILED. Test experiment status expected to be $1" > /workspace/status.txt
+    print_serial=true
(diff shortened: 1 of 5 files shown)


launcher/image/test/scripts/test_experiment_value.sh
```diff
@@ -0,0 +1,21 @@
...
+SERIAL_OUTPUT=$(read_serial $2 $3) 
+print_serial=false
+
+if echo $SERIAL_OUTPUT | grep -q "EnableTestFeatureForImage:$1"
+then
+    echo "- test experiment verified $1"
+else
+    echo "FAILED: experiment status expected to be $1"
+    echo "TEST FAILED. Test experiment status expected to be $1" > /workspace/status.txt
+    print_serial=true
```
alexmwu, line 16: very nit, but you could just `echo $SERIAL_OUTPUT` here instead of have the `print_serial` var

alexmwu (approved):
Please have the tests run successfully before merging. (I think we may need to increase our quota)

launcher/image/test/scripts/test_experiment_value.sh
JoshuaKrstic: Cleaned up VMs for now and restarted these tests.
Will look into the quota now.

JoshuaKrstic: Yes definitely, Ill handle this in a follow-up CL. We do this in all of our scripts and we should clean them all up together, IMO.
PLMK if you disagree and I can handle this now.

PR #391, opened 2023-11-16
Enable memory monitoring in CS

yawangwang:
Enable memory monitoring in CS (see [go/monitoring-enablement-in-hardened-image-1-pager](http://goto.google.com/monitoring-enablement-in-hardened-image-1-pager)).
This PR includes changes to:
1. Unmask node-problem-detector.service for hardened image.
2. Override the default config file [system-stats-monitor.json](https://cos.googlesource.com/cos/overlays/board-overlays/+/refs/heads/master/project-lakitu/app-admin/node-problem-detector/files/system-stats-monitor.json) to collect memory/bytes_used metrics only for CS.
3. Configure preload.sh to allow hardened/debug images have this change.
4. Add a library `systemctl` to use dbus API to start `node-problem-detector.service`.
5. Introduce `tee.launch_policy.monitoring_memory_allow` to launch policy and `tee-monitoring-memory-enable` to launchSpec.
6. **Add a library `nodeproblemdetector` which provides methods to override node-problem-detector configurations. This library will not be used for now, but will be used in the future if operators want to opt-in more metrics monitoring options: cpu, disk, etc.**
7. Add image tests to check if memory monitoring is enabled.


launcher/internal/systemctl/systemctl.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Close disconnects from dbus.
+func (s *Systemctl) Close() { s.dbus.Close() }
+
+func runSystemdCmd(cmdFunc func(string, string, chan<- string) (int, error), cmd string, unit string) error {
+	progress := make(chan string, 1)
+
+	// Run systemd command in "replace" mode to start the unit and its dependencies,
+	// possibly replacing already queued jobs that conflict w∏ith this.
```
jkl73, line 51: extra character  "w∏ith"

launcher/container_runner.go
```diff
@@ -507,6 +508,23 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
+			return fmt.Errorf("failed to create systemctl client: %v", err)
+		}
+		defer s.Close()
+
+		if err := s.Start("node-problem-detector.service"); err != nil {
+			return fmt.Errorf("failed to start node-problem-detector.service: %v", err)
+		}
+		r.logger.Println("node-problem-detector.service successfully started.")
+	} else {
+		r.logger.Println("node-problem-detector.service disabled.")
```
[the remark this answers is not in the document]
yawangwang: Done.

launcher/internal/systemctl/systemctl.go
yawangwang: Done.

launcher/internal/healthmonitoring/nodeproblemdetector/systemstats_config.go
```diff
@@ -0,0 +1,55 @@
...
+
+type memoryStatsConfig struct {
+	MetricsConfigs map[string]metricConfig `json:"metricsConfigs"`
+}
+
+// SystemStatsConfig contains configurations for `System Stats Monitor`,
+// a problem daemon in node-problem-detector that collects pre-defined health-related metrics from different system components.
+// For now we only consider collecting memory related metrics.
+// View the comprehensive configuration details on https://github.com/kubernetes/node-problem-detector/tree/master/pkg/systemstatsmonitor#detailed-configuration-options
+type SystemStatsConfig struct {
```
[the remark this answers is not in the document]
yawangwang: Yes, your understanding is right - this `systemstats_config.go` is not used anywhere for now, but the value of having this is for future proof: if customers want to enable other monitoring options in the future (cpu, disk, os etc.,), this file will allow them to override the `system-stats-monitor.json` on the run time instead of image build time.

yawangwang:
@jkl73 Added three another config files in addition to `system-stats-monitor.json` to turn off `system-log-monitor` and `custom-plugin-monitor` for node-problem-detector. These two daemon processes were intended to monitor/report kernel logs, docker logs and boot process, but we decided to turn them off.

launcher/internal/systemctl/systemctl.go
```diff
@@ -0,0 +1,62 @@
...
+	conn, err := dbus.New()
+	if err != nil {
+		return nil, err
+	}
+	return &Systemctl{dbus: conn}, nil
+}
+
+// Start is the equivalent of `systemctl start $unit`.
+func (s *Systemctl) Start(unit string) error {
+	return runSystemdCmd(s.dbus.StartUnit, "start", unit)
```
jkl73, line 36: Seems like StartUnit is deprecated, try to use StartUnitContext instead https://github.com/coreos/go-systemd/blob/v22.5.0/dbus/methods.go#L116, (same for the following functions)

It's probably because you imported a old version of dbus

launcher/go.mod
```diff
@@ -30,7 +30,10 @@ require (
 	github.com/containerd/fifo v1.0.0 // indirect
 	github.com/containerd/ttrpc v1.1.0 // indirect
 	github.com/containerd/typeurl v1.0.2 // indirect
+	github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf // indirect
+	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
```
jkl73, line 34: Seems like should only keep the v22 version here

launcher/internal/systemctl/systemctl.go
```diff
@@ -0,0 +1,62 @@
...
+func runSystemdCmd(cmdFunc func(string, string, chan<- string) (int, error), cmd string, unit string) error {
+	progress := make(chan string, 1)
+
+	// Run systemd command in "replace" mode to start the unit and its dependencies,
+	// possibly replacing already queued jobs that conflict with this.
+	if _, err := cmdFunc(unit, "replace", progress); err != nil {
+		return fmt.Errorf("failed to run systemctl [%s] for unit [%s]: %v", cmd, unit, err)
+	}
+
+	if result := <-progress; result != "done" {
```
jkl73, line 56: This will block, do you know the timeout for those dbus function?

Also probably should log something in container_runner.go to indicate it is trying to do some systemctl operation.

jkl73 (approved):

launcher/go.mod
yawangwang: Done.

launcher/internal/systemctl/systemctl.go
yawangwang: dbus will call the underlying C library. You can check the timeout definitions at https://dbus.freedesktop.org/doc/api/html/group__DBusTimeout.html. The dbus config file usually locates at `/usr/share/dbus-1/system.conf` with a default timeout 25s.

Added systemctl loggings to container_runner.go.

PR #395, opened 2023-11-28
Integrate nonce verification as part of the TDX quote validation procedure.

jrjatin:
With the integration of the validate package, the go-tdx-guest library now supports TDX quote validation using nonce. I have implemented nonce verification in the TDX quote verification function and added corresponding testcases.


server/verify_test.go
```diff
@@ -374,21 +386,30 @@ func TestVerifyBasicAttestationWithTdx(t *testing.T) {
...
 		TrustedAKs: []crypto.PublicKey{ak.PublicKey()},
+		TEEOpts:    teeopts,
 	}); err != nil {
 		t.Errorf("failed to verify: %v", err)
 	}
 
+	if _, err := VerifyAttestation(attestation, VerifyOpts{
+		Nonce:      append(nonce, 0),
+		TrustedAKs: []crypto.PublicKey{ak.PublicKey()},
+		TEEOpts:    teeopts,
```
alexmwu, line 408: Please also test the code path where we change the *TEE* nonce in addition to the TPM  nonce

cmd/verify.go
```diff
@@ -52,8 +52,16 @@ var debugCmd = &cobra.Command{
...
+			if len(teeNonce) != 0 {
+				validateOpts = &server.VerifyTdxOpts{
+					Validation:   server.TdxDefaultValidateOpts(teeNonce),
+					Verification: tv.DefaultOptions(),
+				}
+			} else {
+				validateOpts = &server.VerifyTdxOpts{
+					Validation:   server.TdxDefaultValidateOpts(nonce),
+					Verification: tv.DefaultOptions(),
+				}
```
alexmwu, line 64: Please include a unit test.

server/verify_test.go
```diff
@@ -358,7 +370,7 @@ func TestVerifyBasicAttestationWithTdx(t *testing.T) {
 	}
 	defer ak.Close()
 
-	nonce := []byte("super secret nonce")
+	nonce := tdxReportData
```
alexmwu, line 373: I don't know the TDX specifics very well. Does "reportData" usually refer to a nonce?

server/verify_test.go
```diff
@@ -1140,10 +1161,9 @@ func TestVerifyAttestationWithTdx(t *testing.T) {
 	}
 	defer ak.Close()
 
-	nonce := []byte("super secret nonce")
```
alexmwu, line 1143: We should use a different value for the TPM nonce. That way we make sure the TPM and TEE nonces don't subtly depend on each other in attest. So, make a `tpmNonce` and a `teeNonce`. This makes the tests easier to read.

server/verify_test.go
```diff
@@ -1210,6 +1228,7 @@ func TestVerifyAttestationWithTdx(t *testing.T) {
 				Nonce:      nonce,
 				TrustedAKs: []crypto.PublicKey{ak.PublicKey()},
 				TEEOpts: &VerifyTdxOpts{
+					Validation:   TdxDefaultValidateOpts(nonce),
```
alexmwu, line 1231: We should have happy path without TEEOpts set as well

server/verify_test.go
```diff
@@ -1243,6 +1264,19 @@ func TestVerifyAttestationWithTdx(t *testing.T) {
...
+				TrustedAKs: []crypto.PublicKey{ak.PublicKey()},
+				TEEOpts: &VerifyTdxOpts{
+					Validation:   TdxDefaultValidateOpts([]byte("badNonce")),
+					Verification: tv.DefaultOptions(),
+				},
+			},
+			attest:  attestation,
+			wantErr: "quote field REPORT_DATA",
+		},
 	}
```
alexmwu, line 1280: In general, these tests are difficult to read due to the large structs getting created in the test table. Instead of creating them here, let's populate these tables with the actual variables we care about (`VerifyOpts.nonce`, the TEE nonce, `attestation`, `wantErr`, etc.). Then create the VerifyOpts object in the `t.Run` block.

server/verify_test.go
jrjatin: Yes!

server/verify_test.go
jrjatin: This code path is tested under `TestVerifyAttestationWithTdx` on line no. 1268 `name: "Wrong TEE Nonce"`

server/verify_test.go
```diff
@@ -40,6 +40,16 @@ import (
...
+var tdxReportData = []byte{
+	108, 98, 222, 193, 184, 25, 23, 73,
+	163, 29, 171, 73, 11, 229, 50, 163,
+	89, 68, 222, 164, 124, 174, 241, 249,
+	128, 134, 57, 147, 217, 137, 149, 69,
+	235, 116, 6, 163, 141, 30, 237, 49,
+	59, 152, 122, 70, 125, 172, 234, 214,
+	240, 200, 122, 109, 118, 108, 102, 246,
+	242, 159, 138, 203, 40, 31, 17, 19,
+}
```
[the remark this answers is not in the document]
jrjatin: Extracted a nonce value from a TDX sample quote for use in verifier tests. This value is now stored in an embedded binary file.

server/verify_tdx.go
```diff
@@ -1,25 +1,43 @@
...
+		TdQuoteBodyOptions: validate.TdQuoteBodyOptions{}}
+	policy.TdQuoteBodyOptions.ReportData = make([]byte, tabi.ReportDataSize)
+	copy(policy.TdQuoteBodyOptions.ReportData, tpmNonce)
+	return policy
+}
+
+// TdxDefaultOptions returns a default validation policy and verification options for TDX
 // attestation quote on GCE.
-func TdxDefaultOptions() *VerifyTdxOpts {
+func TdxDefaultOptions(tpmNonce []byte) *VerifyTdxOpts {
```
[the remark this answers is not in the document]
jrjatin: Done

server/verify_test.go
jrjatin: Done

server/verify_test.go
jrjatin: Added

server/verify_test.go
jrjatin: Done

server/verify_tdx.go
```diff
@@ -1,25 +1,43 @@
...
 
 // VerifyTdxOpts allows for customizing the functionality of VerifyAttestation's TDX verification.
 type VerifyTdxOpts struct {
+	Validation   *validate.Options
 	Verification *tv.Options
 }
 
-// TdxDefaultOptions returns a default verification options for TDX
+// TdxDefaultValidateOpts returns a default validation policy for TDX attestation quote on GCE.
+func TdxDefaultValidateOpts(tpmNonce []byte) *validate.Options {
```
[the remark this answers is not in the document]
jrjatin: Done

server/verify_test.go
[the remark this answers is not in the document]
jrjatin: I have added binary file under tdx_test_files

cmd/verify.go
jrjatin: Added a unit test to check complete attestation that includes fake tdx attestation.

cmd/verify_test.go
```diff
@@ -143,3 +149,82 @@ func TestHwAttestationPass(t *testing.T) {
 		})
 	}
 }
+
+func TestTdxAttestation(t *testing.T) {
+	file1, err := os.Create("attest")
```
alexmwu, line 154: Tests should use https://pkg.go.dev/testing#T.TempDirtemp then create files in that temp directory.

cmd/verify_test.go
```diff
@@ -143,3 +149,82 @@ func TestHwAttestationPass(t *testing.T) {
...
+	var out []byte
+	if format == "binarypb" {
+		out, err = proto.Marshal(attestation)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal attestation proto: %v", attestation)
+		}
+	} else {
+		out = []byte(marshalOptions.Format(attestation))
+	}
+	return out, nil
```
alexmwu, line 229: This function would make more sense returning an Attestation proto message rather than the marshaled proto. This should be test-specific logic, and AFAICT we don't use the format flag right now.

cmd/verify_test.go
alexmwu: Should have a more descriptive name like attestFile

cmd/verify_test.go
```diff
@@ -143,3 +149,82 @@ func TestHwAttestationPass(t *testing.T) {
 		})
 	}
 }
+
+func TestTdxAttestation(t *testing.T) {
+	file1, err := os.Create("attest")
+	if err != nil {
+		t.Fatal(err)
+	}
+	file2 := makeOutputFile(t, "verify")
```
alexmwu, line 158: verifyOutputFile or verifyFile

cmd/verify_test.go
```diff
@@ -143,3 +149,82 @@ func TestHwAttestationPass(t *testing.T) {
...
+	defer os.RemoveAll(file1.Name())
+	defer os.RemoveAll(file2)
+
+	tpmNonce := "1234"
+	teeNonce := "6c62dec1b8191749a31dab490be532a35944dea47caef1f980863993d9899545eb7406a38d1eed313b987a467dacead6f0c87a6d766c66f6f29f8acb281f1113"
```
alexmwu, line 163: Isn't this the same as TdxReportData? Why are we redefining it here?

server/verify_tdx.go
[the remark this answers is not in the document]
alexmwu: I still see tpmNonce

server/verify_tdx.go
```diff
@@ -1,25 +1,43 @@
...
-func TdxDefaultOptions() *VerifyTdxOpts {
+func TdxDefaultOptions(tdxNonce []byte) *VerifyTdxOpts {
 	return &VerifyTdxOpts{
+		Validation:   TdxDefaultValidateOpts(tdxNonce),
 		Verification: tv.DefaultOptions(),
 	}
 }
 
 // VerifyTdxAttestation checks that the TDX attestation quote is valid
 func VerifyTdxAttestation(attestation *tpb.QuoteV4, opts *VerifyTdxOpts) error {
```
alexmwu, line 35: You'll need to rebase on https://github.com/google/go-tpm-tools/pull/405

server/verify_tdx.go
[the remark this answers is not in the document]
jrjatin: Missed changing name in this function. Now I have renamed here as well.

cmd/verify_test.go
jrjatin: Thanks for this info. Created a temp dir and renamed it to attestFile

cmd/verify_test.go
jrjatin: Makes sense, I have changed function output to attestation proto message. I have added format flag as textproto in verify args so that marshalling logic is test specific.

cmd/verify_test.go
jrjatin: Done

cmd/verify_test.go
jrjatin: Yes, this should be a hex string.

server/verify_tdx.go
jrjatin: Done

internal/test/test_tpm.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	"encoding/binary"
 	"io"
 	"sync"
 	"testing"
 
 	"github.com/google/go-attestation/attest"
 	"github.com/google/go-tpm-tools/simulator"
 	"github.com/google/go-tpm/legacy/tpm2"
+	gotpm "github.com/google/go-tpm/tpm2"
```
alexmwu, line 12: Why did you rename the import here? Generally avoid this: https://google.github.io/styleguide/go/decisions#import-renaming

jrjatin: Since tpm2 is imported from 2 different packages, so I have added alias to one of tpm2 import. I have renamed the alias to gtpm2 to avoid any confusion.

alexmwu (approved):

PR #397, opened 2023-12-05
Fix teeserver context reset issue & add container signature cache

yawangwang:
Discussed offline, we decide to introduce container signatures cache to attestation agent.
So we should refresh the signatures cache every time the token refresher goroutine runs (~1/hr), and ODA will only read from the cache.
This PR also includes the following changes:
- fixes teeserver context reset issue
- removes the existing experiment flag `EnableSignedContainerImage`, add a new flag `EnableSignedContainerCache`

### DIFF
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -35,6 +35,7 @@ type principalIDTokenFetcher func(audience string) ([][]byte, error)
 type AttestationAgent interface {
 	MeasureEvent(cel.Content) error
 	Attest(context.Context, AttestAgentOpts) ([]byte, error)
+	Refresh(context.Context) error
 }
 
 // AttestAgentOpts contains user generated options when calling the
@@ -54,6 +55,7 @@ type agent struct {
 	cosCel           cel.CEL
 	launchSpec       spec.LaunchSpec
 	logger           *log.Logger
+	sigsCache        *sigsCache
 }
 
 // CreateAttestationAgent returns an agent capable of performing remote
@@ -72,6 +74,7 @@ func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher tpmKeyFetcher, ver
 		sigsFetcher:      sigsFetcher,
 		launchSpec:       launchSpec,
 		logger:           logger,
+		sigsCache:        &sigsCache{},
 	}
 }
 
@@ -111,12 +114,15 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
 		},
 	}
 
-	if a.launchSpec.Experiments.EnableSignedContainerImage {
-		signatures := fetchContainerImageSignatures(ctx, a.sigsFetcher, a.launchSpec.SignedImageRepos, a.logger)
-		if len(signatures) > 0 {
-			req.ContainerImageSignatures = signatures
-			a.logger.Printf("Found container image signatures: %v\n", signatures)
-		}
+	var signatures []oci.Signature
+	if a.launchSpec.Experiments.EnableSignedContainerCache {
+		signatures = a.sigsCache.get()
+	} else {
+		signatures = fetchContainerImageSignatures(ctx, a.sigsFetcher, a.launchSpec.SignedImageRepos, a.logger)
+	}
+	if len(signatures) > 0 {
+		req.ContainerImageSignatures = signatures
+		a.logger.Printf("Found container image signatures: %v\n", signatures)
 	}
 
 	resp, err := a.client.VerifyAttestation(ctx, req)
@@ -127,6 +133,17 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
 		a.logger.Printf("Partial errors from VerifyAttestation: %v", resp.PartialErrs)
 	}
 	return resp.ClaimsToken, nil
+}
+
+// Refresh refreshes the internal state of the attestation agent.
+// It will reset the container image signatures for now.
+func (a *agent) Refresh(ctx context.Context) error {
+	if a.launchSpec.Experiments.EnableSignedContainerCache {
+		signatures := fetchContainerImageSignatures(ctx, a.sigsFetcher, a.launchSpec.SignedImageRepos, a.logger)
+		a.sigsCache.set(signatures)
+		a.logger.Printf("Refreshed container image signature cache: %v\n", signatures)
+	}
+	return nil
 }
 
 func (a *agent) getAttestation(nonce []byte) (*pb.Attestation, error) {
@@ -148,7 +165,6 @@ func (a *agent) getAttestation(nonce []byte) (*pb.Attestation, error) {
 	return attestation, nil
 }
 
-// TODO: cache signatures so we don't need to fetch every time.
 func fetchContainerImageSignatures(ctx context.Context, fetcher signaturediscovery.Fetcher, targetRepos []string, logger *log.Logger) []oci.Signature {
 	signatures := make([][]oci.Signature, len(targetRepos))
 
@@ -172,4 +188,22 @@ func fetchContainerImageSignatures(ctx context.Context, fetcher signaturediscove
 		foundSigs = append(foundSigs, sigs...)
 	}
 	return foundSigs
+}
+
+type sigsCache struct {
+	mu    sync.RWMutex
+	items []oci.Signature
+}
+
+func (c *sigsCache) set(sigs []oci.Signature) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.items = make([]oci.Signature, len(sigs))
+	copy(c.items, sigs)
+}
+
+func (c *sigsCache) get() []oci.Signature {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.items
 }
--- a/launcher/agent/agent_test.go
+++ b/launcher/agent/agent_test.go
@@ -7,6 +7,8 @@ import (
 	"encoding/base64"
 	"fmt"
 	"log"
+	"runtime"
+	"sync"
 	"testing"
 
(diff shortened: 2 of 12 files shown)


launcher/image/test/test_discover_signatures.yaml
```diff
@@ -4,7 +4,7 @@ substitutions:
   '_CLEANUP': 'true'
   '_VM_NAME_PREFIX': 'discover-signatures'
   '_ZONE': 'us-west1-a'
-  '_WORKLOAD_IMAGE': 'us-west1-docker.pkg.dev/confidential-space-images-dev/cs-integ-test-images/basic-test:latest'
+  '_WORKLOAD_IMAGE': 'us-west1-docker.pkg.dev/confidential-space-images-dev/cs-integ-test-images/ipc/happypath:latest'
```
JoshuaKrstic, line 7: We should also have a test for the default token to ensure signed container works with it as well

yawangwang: Done. I created a separate test for ODA and signed container, and keep this test for the default token.

launcher/image/test/test_oda_with_signed_container.yaml
```diff
@@ -0,0 +1,73 @@
...
+          '-z', '${_ZONE}',
+        ]
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: TestCustomToken
+  entrypoint: 'bash'
+  args: ['scripts/test_custom_token.sh', "true", '${_VM_NAME_PREFIX}-${BUILD_ID}', '${_ZONE}']
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: BasicDiscoverSignaturesTest
+  entrypoint: 'bash'
+  args: ['scripts/test_launcher_workload_discover_signatures.sh', '${_VM_NAME_PREFIX}-${BUILD_ID}', '${_ZONE}', '2']
```
alexmwu, line 43: Why do we expect 2 signatures? I only see one cosign call above.

alexmwu (approved):

launcher/image/test/test_oda_with_signed_container.yaml
yawangwang: No, this is to check how many times `Found container image signatures` is being logged. Fetching signatures will occur on refresh the default token, and on teeserver receiving ODA requests, so the expected number is 2.

alexmwu: In that case better comments in the test and/or in `launcher/image/test/scripts/test_launcher_workload_discover_signatures.sh` would be good. When reading it, I assumed it was trying to find the number of signatures.

yawangwang: Done.

launcher/agent/agent_test.go
```diff
@@ -255,3 +253,58 @@ func TestFetchContainerImageSignatures(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		cache := &sigsCache{}
+		sdClient := signaturediscovery.NewFakeClient()
+		gotSigs := fetchContainerImageSignatures(ctx, sdClient, tc.targetRepos, log.Default())
+		cache.set(gotSigs)
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			// Read cache from a separate goroutine.
+			t.Parallel()
```
alexmwu, line 286: shouldn't we also read the cache before setting?

launcher/agent/agent_test.go
```diff
@@ -255,3 +253,58 @@ func TestFetchContainerImageSignatures(t *testing.T) {
 		})
 	}
 }
+
+func TestCache(t *testing.T) {
```
alexmwu, line 257: I'm not sure I follow why we need this test. I think I would ideally want to a few things for this change: synchronization on the mutex (throwing goroutines and trying to catch a race condition), making sure `Refresh` updates appropriately when the backing store changes (e.g., add sig, delete sig), and finally `refreshToken` does indeed update the default token and the cache (unit test and perhaps a manual test).

launcher/image/test/test_oda_with_signed_container.yaml
alexmwu: We should compare the signatures fetched are the same somehow

alexmwu (changes requested):

launcher/image/test/test_oda_with_signed_container.yaml
yawangwang: Done.

launcher/agent/agent.go
```diff
@@ -129,6 +131,16 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
 	return resp.ClaimsToken, nil
 }
 
+// Refresh refreshes the internal state of the attestation agent.
+// It will reset the container image signatures for now.
+func (a *agent) Refresh(ctx context.Context) error {
```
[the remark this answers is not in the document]
yawangwang: yes, adding error is for future-proof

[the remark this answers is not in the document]
yawangwang: Done.

launcher/agent/agent_test.go
yawangwang: See comments below.

launcher/agent/agent_test.go
yawangwang: Done. 
- Added unit tests `TestCacheConcurrentSet` and `TestCacheConcurrentGet` that spawn goroutines for synchronization checks.
- Added a unit test `TestRefreshTokenWithSignedContainerCacheEnabled` in `container_runner_test.go` to check `refreshToken` updates the default token and cache appropriately when the backing store changes (signatures get updated.)

launcher/agent/agent_test.go
```diff
@@ -255,3 +256,82 @@ func TestFetchContainerImageSignatures(t *testing.T) {
...
+	sdClient := signaturediscovery.NewFakeClient()
+	targetRepos := []string{signaturediscovery.FakeRepoWithSignatures}
+
+	var wg sync.WaitGroup
+	for i := 0; i < runtime.NumCPU(); i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			sigs := fetchContainerImageSignatures(ctx, sdClient, targetRepos, log.Default())
+			cache.set(sigs)
```
alexmwu, line 309: Since we aren't intermingling sets and gets in a concurrent manner, we aren't actually testing the sync primitive. Consider intermingling data accesses and using the `-race` flag: https://go.dev/doc/articles/race_detector

launcher/container_runner_test.go
```diff
@@ -131,6 +169,66 @@ func TestRefreshToken(t *testing.T) {
...
+	fakeAgent.attestFunc = func(context.Context, agent.AttestAgentOpts) ([]byte, error) {
+		return createJWTWithSignatures(t, fakeAgent.sigsCache), nil
+	}
+
+	runner := ContainerRunner{
+		attestAgent: fakeAgent,
+		logger:      log.Default(),
+	}
+
+	if err := os.MkdirAll(launcherfile.HostTmpPath, 0744); err != nil {
```
alexmwu, line 193: should use https://pkg.go.dev/testing#B.TempDir

launcher/container_runner_test.go
```diff
@@ -131,6 +169,66 @@ func TestRefreshToken(t *testing.T) {
...
+		t.Fatalf("Error creating host token path directory: %v", err)
+	}
+
+	_, err := runner.refreshToken(ctx)
+	if err != nil {
+		t.Fatalf("refreshToken returned with error: %v", err)
+	}
+
+	// simulate adding signatures.
```
alexmwu, line 202: nit: here and below capitalize comments

launcher/image/test/test_oda_with_signed_container.yaml
```diff
@@ -0,0 +1,50 @@
+substitutions:
+  '_IMAGE_NAME': ''
+  '_IMAGE_PROJECT': ''
+  '_CLEANUP': 'true'
+  '_VM_NAME_PREFIX': 'oda-signedcontainer'
+  '_ZONE': 'us-east1-b'
+  '_WORKLOAD_IMAGE': 'us-west1-docker.pkg.dev/confidential-space-images-dev/cs-integ-test-images/ipc/happypath@sha256:999831a7b8f8afd323e2359f3c1192206be2aa1d4f3b19f0739eff5f01f83b9e'
```
alexmwu, line 7: Add a comment here that if the workload image changes, the commit author should change the cosign signature as well.

alexmwu (approved):

launcher/agent/agent.go
```diff
@@ -129,6 +135,16 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
...
+	if a.launchSpec.Experiments.EnableSignedContainerCache {
+		signatures := fetchContainerImageSignatures(ctx, a.sigsFetcher, a.launchSpec.SignedImageRepos, a.logger)
```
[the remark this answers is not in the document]
yawangwang: Done.

launcher/agent/agent_test.go
yawangwang: Done.

launcher/container_runner_test.go
yawangwang: Since `refreshToken` will always write OIDC token to a fixed location, I think we'll have to create a fixed directory here instead of a random temp test directory.

launcher/container_runner_test.go
yawangwang: Done.

launcher/image/test/test_oda_with_signed_container.yaml
yawangwang: Done

PR #404, opened 2023-12-21
Add eventlog parse logics for memory monitoring

yawangwang:
Adding eventlog parse logics for memory monitoring.


jkl73:
We can split the change to 2 PRs, we first merge the eventlog parsing logic, and wait for it to deploy to the service. And then merge the measuring logic PR, so the image test can pass here.

yawangwang:
> We can split the change to 2 PRs, we first merge the eventlog parsing logic, and wait for it to deploy to the service. And then merge the measuring logic PR, so the image test can pass here.

Looks like confidential space VMs failed to start due to unknown COS event type errors returned by `VerifyAttestation` if we add measuring logics to this PR. 
Done.

server/eventlog.go
```diff
@@ -198,6 +199,10 @@ func getVerifiedCosState(coscel cel.CEL) (*pb.AttestedCosState, error) {
 			cosState.Container.OverriddenEnvVars[envName] = envVal
 		case cel.LaunchSeparatorType:
 			seenSeparator = true
+		case cel.MemoryMonitorType:
+			if len(cosTlv.EventContent) > 0 && cosTlv.EventContent[0] == uint8(1) {
```
jkl73, line 203: Can be more definitive here, by making sure the len(cosTlv.EventContent) == 1

jkl73: And for failsafe, I think the default value for cosState.HealthMonitoring.MemoryEnabled should be true, and when we see the eventContent == 0 here, we can turn it to false

yawangwang: > Can be more definitive here, by making sure the len(cosTlv.EventContent) == 1

Done.

yawangwang: > And for failsafe, I think the default value for cosState.HealthMonitoring.MemoryEnabled should be true, and when we see the eventContent == 0 here, we can turn it to false

Discussed offline, we agree to keep the default value as false.

jkl73 (approved):

PR #405, opened 2023-12-22
Use TDX quote provider to attest and verify.

vbalain:
go-tpm-tools uses go-tdx-guest's device driver to attest and verify quotes but with the recently upstreamed Unified ABI patch from Intel fetches attestation quotes via ConfigFS subsystem.  go-tdx-guest has incorporated this change and now fetches quote via go-configfs-tsm library instead of device driver's IOCTLs.  Added TDX quote provider in attest module which would be used instead of TDX device driver, and updated unit tests.


client/attest.go
```diff
@@ -19,9 +20,9 @@ const (
 )
 
 // TEEDevice is an interface to add an attestation report from a TEE technology's
-// attestation driver.
+// attestation driver or quote provider.
 type TEEDevice interface {
```
vbalain, line 24: Ideally this interface should be renamed to TEEQuoteProvider but not updating it to keep changes to a minimum.

client/attest.go
```diff
@@ -144,10 +145,17 @@ type SevSnpDevice struct {
 
 // TdxDevice encapsulates the TDX attestation device to add its attestation quote
 // to a pb.Attestation.
+// Deprecated: TdxDevice is deprecated. It is recommended to use TdxQuoteProvider.
```
vbalain, line 148: Deprecated!

client/attest.go
```diff
@@ -234,6 +238,73 @@ func (d *TdxDevice) Close() error {
 	return nil
 }
 
+// CreateTdxQuoteProvider creates the TDX quote provider and wraps it with behavior
+// that allows it to add an attestation quote to pb.Attestation.
+func CreateTdxQuoteProvider() (*TdxQuoteProvider, error) {
```
vbalain, line 243: TDX QuoteProvider uses same interface as a TEEDevice.

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-attestation v0.5.0
 	github.com/google/go-cmp v0.5.9
 	github.com/google/go-sev-guest v0.9.3
-	github.com/google/go-tdx-guest v0.2.3-0.20231011100059-4cf02bed9d33
+	github.com/google/go-tdx-guest v0.2.3-0.20231222042644-aeefcb0d0eb3
```
vbalain, line 9: go-tdx-guest version update to use latest code

server/verify_tdx.go
```diff
@@ -19,7 +18,7 @@ func TdxDefaultOptions() *VerifyTdxOpts {
 }
 
 // VerifyTdxAttestation checks that the TDX attestation quote is valid
-func VerifyTdxAttestation(attestation *tpb.QuoteV4, opts *VerifyTdxOpts) error {
+func VerifyTdxAttestation(attestation any, opts *VerifyTdxOpts) error {
```
vbalain, line 21: Making API future compatible with other quote formats.

jkl73: This (any) can be confused for people to call this function, could you add more description of "attestation" and potentially how to generate an "attestation"

cmd/attest_test.go
jkl73, line 1: Seems like TestHardwareAttestationPass is failing in CI, probably just need to update the wanterr

server/verify_tdx.go
vbalain: Added description.

cmd/attest_test.go
vbalain: Yes. Command **go test ./...** didn't run tests in go-tpm-tools/cmd directory. Updated expected errors in attest_test.go and verify_test.go in cmd. It should fix CI errors.

jkl73 (approved):

PR #409, opened 2024-01-24
Fix parsing memory monitoring event log

yawangwang:
The default measurement value for memory monitoring is false for now. So for the image version `231201` that supports memory monitoring but without its measurement, if operators enable the memory monitoring feature, the measurement is still false. 

Since the measurement should always reflect the actual state of the image, we should make the field as null for image versions that don't support memory monitoring measurement.

This PR includes a fix to make the proto field `memory_enabled` as optional such that the default value is a pointer instead a boolean value.


jkl73 (approved):
good catch!

PR #412, opened 2024-02-14
Use resizable integrity-fs with in-memory tags

alexmwu:
The Confidential Space image now uses dm-crypt with a backing dm-integrity where the integrity tags are stored in memory vs getting interleaved on the actual disk.

This changes comes with support for boot-time integrity tag initialization speed up and automatically resizing of the stateful partition on the boot disk.

### DIFF
--- a/launcher/image/preload.sh
+++ b/launcher/image/preload.sh
@@ -111,7 +111,9 @@ main() {
   # Install container launcher.
   copy_launcher
   setup_launcher_systemd_unit
-  append_cmdline "cos.protected_stateful_partition=e"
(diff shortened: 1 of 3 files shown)


launcher/image/preload.sh
```diff
@@ -111,7 +111,7 @@ main() {
   # Install container launcher.
   copy_launcher
   setup_launcher_systemd_unit
-  append_cmdline "cos.protected_stateful_partition=e"
+  append_cmdline "cos.protected_stateful_partition=m"
```
jkl73, line 114: Add a comment on the minimum COS version support on this feature

jkl73 (approved):

PR #375, opened 2023-10-14
[cmd] Add new command token in the CLI tool

Ruide:
Command Description: Fetch an attestation report from GCE VM vTPM and send it to Google Attestation Service for an OIDC token.

This command improves usability for a GCE VM user.

This PR replaces closed https://github.com/google/go-tpm-tools/pull/368. Branch name changed.

PR #417, opened 2024-02-27
add cloud log flag: add records to cloud logging when fetching token from attestation verifier

Ruide:


cmd/flags.go
```diff
@@ -130,6 +130,11 @@ func addAsAddressFlag(cmd *cobra.Command) {
 		"Attestation Service address")
 }
 
+// Lets this command enable Cloud logging
+func addCloudLoggingFlag(cmd *cobra.Command) {
+	cmd.Flags().Bool("cloudLog", false, "cloud logging switch")
+}
+
```
alexmwu, line 137: `cloudLog` and `asAddr` should be lower case and separated by hypens. For example `cloud-log` and `as-addr`. See the `hash-algo` flag.

alexmwu: `as-addr` is not very descriptive either. Consider something like `service-endpoint` or `verifier-endpoint`

cmd/token.go
```diff
@@ -101,8 +104,18 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 		}
 		gceAK.Close()
 
+		cloudLogEnabled, _ := cmd.Flags().GetBool("cloudLog")
+		if cloudLogEnabled {
+			cloudLogClient, err = logging.NewClient(ctx, projectID)
+			if err != nil {
+				return fmt.Errorf("failed to create cloud logging client: %w", err)
```
alexmwu, line 111: Cloud Logging should be capitalized here, as it is a proper service name.

cmd/token.go
```diff
@@ -101,8 +104,18 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
...
+			}
+			cloudLogger = cloudLogClient.Logger("go-tpm-tool")
```
alexmwu, line 113: `gotpm` is the current name of the CLI tool. Please extract a const variable for this name in case we change it.

cmd/token.go
```diff
@@ -142,7 +155,17 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 			}
 		}
 
+		if cloudLogEnabled {
+			cloudLogger.Log(logging.Entry{Payload: map[string]string{"token": string(token)}})
```
alexmwu, line 159: Logging the token seems like it could be dangerous, as it can be used as a bearer token. I thought we were only logging the attestation message

alexmwu: Please PM me the PDD

launcher/agent/agent.go
```diff
@@ -114,6 +117,11 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
 		},
 	}
 
+	if a.cloudLogger != nil {
+		a.cloudLogger.Log(logging.Entry{Payload: challenge})
```
alexmwu, line 121: As I commented on #375, using the agent is a bit of an odd choice since the agent is closely tied with the launcher (consider `MeasureEvent`). We should really refactor out the Attest command if it is useful to you.

cmd/flags.go
```diff
@@ -126,10 +126,15 @@ func addInputFlag(cmd *cobra.Command) {
 
 // Lets this command specify an Attestation Server Address.
 func addAsAddressFlag(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVar(&asAddress, "asAddr", "https://confidentialcomputing.googleapis.com",
+	cmd.PersistentFlags().StringVar(&asAddress, "verifier-endpoint", "https://confidentialcomputing.googleapis.com",
```
alexmwu, line 129: This flag should have documentation. Something like "the attestation verifier endpoint used to retrieve an attestation claims token"

cmd/token.go
alexmwu: As we discussed, I think that we should integrate with custom audience and require a custom audience for cloud logging. This ensures that auditors don't accidentally mint a bearer token that can be used to access other cloud resources.

cmd/flags.go
```diff
@@ -126,10 +126,15 @@ func addInputFlag(cmd *cobra.Command) {
...
 		"Attestation Service address")
 }
 
+// Lets this command enable Cloud logging
+func addCloudLoggingFlag(cmd *cobra.Command) {
+	cmd.Flags().Bool("cloud-log", false, "cloud logging switch")
```
alexmwu, line 135: "logs the attestation and token to Cloud Logging for auditing purposes. Requires the audience flag."

alexmwu: I don't see this change

cmd/flags.go
```diff
@@ -126,8 +127,19 @@ func addInputFlag(cmd *cobra.Command) {
...
+
+// Lets this command enable Cloud logging
+func addCloudLoggingFlag(cmd *cobra.Command) {
+	cmd.Flags().Bool("cloud-log", false, "cloud logging switch")
+}
+
+// Lets this command specify custom audience field of the attestation token
+func addAudienceFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&audience, "audience", "https://api.example.com",
+		"the audience field in the claims token")
```
alexmwu, line 142: the audience field in the claims token. Cannot be sts.googleapis.com.

cmd/flags.go
```diff
@@ -126,8 +127,19 @@ func addInputFlag(cmd *cobra.Command) {
...
 func addAsAddressFlag(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVar(&asAddress, "asAddr", "https://confidentialcomputing.googleapis.com",
-		"Attestation Service address")
+	cmd.PersistentFlags().StringVar(&asAddress, "verifier-endpoint", "https://confidentialcomputing.googleapis.com",
+		"the attestation verifier endpoint used to retrieve an attestation claims token")
+}
+
+// Lets this command enable Cloud logging
+func addCloudLoggingFlag(cmd *cobra.Command) {
+	cmd.Flags().Bool("cloud-log", false, "cloud logging switch")
```
alexmwu, line 136: This should be consistent with our other flags and use a global var like `cloudLog`:
cmd.Flags().Bool(&cloudLog, "cloud-log", false, "cloud logging switch")

cmd/token.go
```diff
@@ -101,11 +106,22 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 		}
 		gceAK.Close()
 
+		cloudLogEnabled, _ := cmd.Flags().GetBool("cloud-log")
```
alexmwu, line 109: This should use `cloudLog`

cmd/token.go
```diff
@@ -101,11 +106,22 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
...
+		if cloudLogEnabled {
+			audience = "Access and Audit"
+			cloudLogClient, err = logging.NewClient(ctx, projectID)
+			if err != nil {
+				return fmt.Errorf("failed to create Cloud Logging client: %w", err)
+			}
+			cloudLogger = cloudLogClient.Logger(toolName)
+			fmt.Fprintf(debugOutput(), "cloudLogger created for project: "+projectID+"\n")
+		}
+
```
alexmwu, line 119: We shouldn't override audiences the user sets. Just check 
```
if cloudLog && audience == "" {
  return errors.New("... ")
}
```

alexmwu (commented):
What about testing?

Also, please address the documentation comments on https://github.com/google/go-tpm-tools/pull/375. Either here or another PR is fine.

cmd/token.go
Ruide: Yes, we should not. I set the default audience just now to "https://api.example.com" and forgot to remove the override here. Thanks for pointing it out! And please let me know if you have other thoughts about default audience value.

launcher/agent/agent.go
Ruide: The refactor out would be in another PR. I think we might want to discuss on where to host verifier package and agent package. I would PM you later!

cmd/flags.go
Ruide: In theory, if the malicious actor really want this field to be sts.googleapis.com. He can use the api to fetch it anyway. Shall I put a note on "Cannot be sts.googleapis.com" or else? What do you recommend?

cmd/flags.go
Ruide: The function signature for cmd.Flags().Bool is func (*pflag.FlagSet).Bool(name string, value bool, usage string) *bool. It is different from cmd.PersistentFlags().StringVar(). So we could not use the pattern of  cmd.Flags().Bool(&cloudLog, "cloud-log", false, "cloud logging switch").

alexmwu: Sorry I meant cmd.Flags().BoolVar. See https://github.com/google/go-tpm-tools/blob/b22bad0e47403f90d754253de3927b3edea29613/cmd/root.go#L35

cmd/flags.go
alexmwu: The service rejects the https://sts.googleapis.com audience

cmd/token.go
alexmwu: the default should be an empty audience, so that users don't accidentally issue a token with https://api.example.com/ by default. An empty audience should mean we don't fill out the audience field

cmd/flags.go
Ruide: Thanks for the explanation! Confirmed. I have another question. Was it a bug that I could get "https://sts.googleapis.com/" before? Or is there a recent change for the service?

alexmwu: I think the trailing `/`

cmd/token_test.go
```diff
@@ -72,14 +80,78 @@ func TestTokenWithGCEAK(t *testing.T) {
 			}
 			defer mockAttestationServer.Stop()
 
-			RootCmd.SetArgs([]string{"token", "--algo", op.algo, "--output", secretFile1, "--asAddr", mockAttestationServer.server.URL})
+			//redirect cloud log from http request to stdout
+			old := os.Stdout
+			r, w, _ := os.Pipe()
+			os.Stdout = w
+
+			RootCmd.SetArgs([]string{"token", "--algo", op.algo, "--output", secretFile1, "--verifier-endpoint", mockAttestationServer.server.URL, "--cloud-log", "--audience", "https://api.test.com", "--unit-test"})
```
alexmwu, line 88: We have a bunch of mocks for the attestation verifier and MDS. Could we do the same for cloud logging?

cmd/token_test.go
```diff
@@ -72,14 +80,78 @@ func TestTokenWithGCEAK(t *testing.T) {
 			}
 			defer mockAttestationServer.Stop()
 
-			RootCmd.SetArgs([]string{"token", "--algo", op.algo, "--output", secretFile1, "--asAddr", mockAttestationServer.server.URL})
+			//redirect cloud log from http request to stdout
+			old := os.Stdout
+			r, w, _ := os.Pipe()
+			os.Stdout = w
```
alexmwu, line 86: We should be able to avoid the pipe redirection and sync primitives by using the output flag properly in a test directory https://pkg.go.dev/testing#T.TempDir

cmd/token.go
```diff
@@ -101,11 +108,37 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 		}
 		gceAK.Close()
 
+		var cloudLogClient *logging.Client
+		var cloudLogger *logging.Logger
+		if cloudLog {
+			if audience == "" {
+				return errors.New("cloud logging requires the --audience flag")
```
alexmwu, line 115: capitalize Cloud Logging, as it's a proper noun

cmd/token_test.go
```diff
@@ -72,7 +72,12 @@ func TestTokenWithGCEAK(t *testing.T) {
 			}
 			defer mockAttestationServer.Stop()
 
-			RootCmd.SetArgs([]string{"token", "--algo", op.algo, "--output", secretFile1, "--asAddr", mockAttestationServer.server.URL})
+			mockCloudLoggingServerAddress, err = newMockCloudLoggingServer()
```
alexmwu, line 75: I would have expected you to check the `loggingHandler.logs` output matches up.

alexmwu (approved):

cmd/token.go
Ruide: go-staticcheck would give "error strings should not be capitalized (ST1005)"

cmd/token_test.go
Ruide: The logs would need to contain a "pb.Attestation" struct from the TPM emulator. Do you have a reference or recommendation on how to create a deterministic "pb.Attestation"?

Ruide: The attestation would give different ak_cert since we are using tpm simulator.

Ruide: I create a new [PR](https://github.com/google/go-tpm-tools/pull/420) with the `loggingHandler.logs` output matches up. I use a simple field comparison for attestation.

PR #423, opened 2024-03-26
Add event-log flag to cmd package

Ruide:
Add an event-log flag to enable specifying custom event log file location.

### DIFF
--- a/cmd/open_other.go
+++ b/cmd/open_other.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os"
 
-	"github.com/google/go-tpm-tools/client"
 	"github.com/google/go-tpm/legacy/tpm2"
 )
 
@@ -16,15 +15,11 @@ var tpmPath string
 // tpmWrapper wrap tpm io.ReadWriteCloser and EventLogGetter interfaces together.
 type tpmWrapper struct {
 	io.ReadWriteCloser
-	client.EventLogGetter
 }
 
 // EventLog allows caller to call the EventLogGetter function
 // of the wrapped TPM.
 func (et tpmWrapper) EventLog() ([]byte, error) {
-	if ExternalTPM != nil {
-		return client.GetEventLog(et.ReadWriteCloser)
-	}
 	return os.ReadFile(eventLog)
 }
 
--- a/cmd/open_other.go
+++ b/cmd/open_other.go
@@ -7,10 +7,26 @@ import (
 	"io"
 	"os"
 
+	"github.com/google/go-tpm-tools/client"
 	"github.com/google/go-tpm/legacy/tpm2"
 )
 
 var tpmPath string
+
+// tpmWrapper wrap tpm io.ReadWriteCloser and EventLogGetter interfaces together.
+type tpmWrapper struct {
+	io.ReadWriteCloser
+	client.EventLogGetter
+}
+
+// EventLog allows caller to call the EventLogGetter function
+// of the wrapped TPM.
+func (et tpmWrapper) EventLog() ([]byte, error) {
+	if ExternalTPM != nil {
+		return client.GetEventLog(et.ReadWriteCloser)
+	}
+	return os.ReadFile(eventLog)
+}
 
 func init() {
 	RootCmd.PersistentFlags().StringVar(&tpmPath, "tpm-path", "",
@@ -18,13 +34,16 @@ func init() {
 }
 
 // On Linux, we have to pass in the TPM path though a flag
-func openImpl() (io.ReadWriteCloser, error) {
+func openImpl() (tpmWrapper, error) {
+	tw := tpmWrapper{}
+	var err error
 	if tpmPath == "" {
-		tpm, err := tpm2.OpenTPM("/dev/tpmrm0")
+		tw.ReadWriteCloser, err = tpm2.OpenTPM("/dev/tpmrm0")
 		if os.IsNotExist(err) {
-			tpm, err = tpm2.OpenTPM("/dev/tpm0")
+			tw.ReadWriteCloser, err = tpm2.OpenTPM("/dev/tpm0")
 		}
-		return tpm, err
+		return tw, err
 	}
-	return tpm2.OpenTPM(tpmPath)
+	tw.ReadWriteCloser, err = tpm2.OpenTPM(tpmPath)
+	return tw, err
 }
--- a/cmd/open_other.go
+++ b/cmd/open_other.go
@@ -12,13 +12,12 @@ import (
 
 var tpmPath string
 
-// tpmWrapper wrap tpm io.ReadWriteCloser and EventLogGetter interfaces together.
+// tpmWrapper wraps a TPM io.ReadWriteCloser that implements client.EventLogGetter.
 type tpmWrapper struct {
 	io.ReadWriteCloser
 }
 
-// EventLog allows caller to call the EventLogGetter function
-// of the wrapped TPM.
+// EventLog fetches the event log specified by the event-log flag.
 func (et tpmWrapper) EventLog() ([]byte, error) {
 	return os.ReadFile(eventLog)
 }
--- a/cmd/token.go
+++ b/cmd/token.go
@@ -5,9 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"log"
-	"os"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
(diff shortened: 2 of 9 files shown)


alexmwu (commented):
Why not implement this interface: https://github.com/google/go-tpm-tools/blob/cf9fd6dc7284b8a2ce9cc499870a4c373a25308f/client/eventlog.go#L17-L19?

Ruide:
> Why not implement this interface:
> 
> https://github.com/google/go-tpm-tools/blob/cf9fd6dc7284b8a2ce9cc499870a4c373a25308f/client/eventlog.go#L17-L19
> 
> ?

It may add complexity and might make it harder to debug in the future. But sure, I implemented the interface version.

cmd/token.go
```diff
@@ -26,6 +28,19 @@ var mockCloudLoggingServerAddress string
...
+
+// EventLog allows caller to call the EventLogGetter function
+// of the wrapped TPM.
+func (et tpmWrapper) EventLog() ([]byte, error) {
+	if ExternalTPM != nil {
+		return client.GetEventLog(et.ReadWriter)
+	}
+	return os.ReadFile(eventLog)
+}
+
```
alexmwu, line 43: I'm sorry. I should've been clearer in my last comment. Would it make sense to have the response of `openImpl` implement `EventLogGetter` and call `os.ReadFile(eventLog)` there?

Ruide: Just to clarify. Do you mean to create a new interface type like `tpmWrapper` which holds `io.ReadWriteCloser` and `EventLogGetter`? And change `openImpl` responce `io.ReadWriteCloser` to `tpmWrapper`?

Ruide: I put down the implementation in the new commit, please let me know if it is what you are suggesting. Thanks!

cmd/open_other.go
```diff
@@ -7,24 +7,43 @@ import (
...
+	"github.com/google/go-tpm-tools/client"
 	"github.com/google/go-tpm/legacy/tpm2"
 )
 
 var tpmPath string
 
+// tpmWrapper wrap tpm io.ReadWriteCloser and EventLogGetter interfaces together.
+type tpmWrapper struct {
+	io.ReadWriteCloser
+	client.EventLogGetter
```
alexmwu, line 19: this is unnecessary. You are already implementing the interface.

cmd/open_other.go
```diff
@@ -7,24 +7,43 @@ import (
...
+}
+
+// EventLog allows caller to call the EventLogGetter function
+// of the wrapped TPM.
+func (et tpmWrapper) EventLog() ([]byte, error) {
+	if ExternalTPM != nil {
+		return client.GetEventLog(et.ReadWriteCloser)
+	}
```
alexmwu, line 27: extTPMWrapper should be used when ExternalTPM is set. This should be used otherwise (see `openTpm`). We should not check ExternalTPM here.

jkl73 (dismissed):

cmd/token_test.go
```diff
@@ -89,6 +89,43 @@ func TestTokenWithGCEAK(t *testing.T) {
 	}
 }
 
+func TestCustomEventLogFile(t *testing.T) {
+	ExternalTPM = nil
```
jkl73, line 93: So in a test, you probably need to use
```
rwc := test.GetTPM(t)
defer client.CheckedClose(t, rwc)
ExternalTPM = rwc
```
to get a TPM

cmd/open_other.go
Ruide: Yes, thanks!

cmd/token_test.go
Ruide: Thanks for the comment! This test is not trying to test with the simulator. This test is to test the eventlog flag. And this test is expected to open /dev/tpm0 and /test-event-log. So this test is expected to return failed to open /dev/tpm0 error for environments with no TPM and failed to open /test-event-log for environments with a TPM. Please let me know if this makes sense!

cmd/open_other.go
```diff
@@ -12,19 +12,33 @@ import (
 
 var tpmPath string
 
+// tpmWrapper wrap tpm io.ReadWriteCloser and EventLogGetter interfaces together.
```
alexmwu, line 15: tpmWrapper wraps a TPM io.ReadWriteCloser that implements client.EventLogGetter.

cmd/open_other.go
```diff
@@ -12,19 +12,33 @@ import (
...
+type tpmWrapper struct {
+	io.ReadWriteCloser
+}
+
+// EventLog allows caller to call the EventLogGetter function
+// of the wrapped TPM.
```
alexmwu, line 21: EventLog fetches the event log specified by the `event-log` flag.

cmd/token_test.go
```diff
@@ -89,6 +89,43 @@ func TestTokenWithGCEAK(t *testing.T) {
...
+	mockAttestationServer, err := util.NewMockAttestationServer()
+	if err != nil {
+		t.Error(err)
+	}
+	defer mockAttestationServer.Stop()
+
+	RootCmd.SetArgs([]string{"token", "--verifier-endpoint", mockAttestationServer.Server.URL, "--event-log", "/test-event-log"})
+	if err := RootCmd.Execute(); err != nil {
+		if err.Error() != "failed to attest: failed to retrieve TCG Event Log: open /test-event-log: no such file or directory" &&
+			err.Error() != "connecting to TPM: stat /dev/tpm0: no such file or directory" {
```
alexmwu, line 123: Why does this check two error strings? Is the failure not deterministic?

alexmwu (approved):

jkl73:
Sorry I should ask this earlier: why do we need to have a flag for the eventlog? Do we expect a user to use an eventlog other than the default one?

And seems like all the existing eventlog fetching logic is built around this idea. And it's hard to add a flag and proper test without changing a lot of existing eventlog logic.

cmd/token_test.go
jkl73: Also my question. Seems like the test will always pass (at least in a CICD) because the /dev/tpm0 not found. So the actual eventlog behavior is not tested.

The behavior is depending on the test environment which is not ideal.

Ruide: I quoted my reply above "This test tests the eventlog flag. And this test is expected to open /dev/tpm0 and /test-event-log. So this test is expected to return failed to open /dev/tpm0 error for environments with no TPM and failed to open /test-event-log for environments with a TPM."

Ruide: We could also make this test to skip for for environments with no TPM. What do you think?

jkl73: yes I think skipping is better

Ruide: Got it, thanks!

cmd/token_test.go
```diff
@@ -89,6 +89,46 @@ func TestTokenWithGCEAK(t *testing.T) {
 	}
 }
 
+func TestCustomEventLogFile(t *testing.T) {
+	if _, err := os.Stat("/dev/tpm0"); os.IsNotExist(err) {
+		t.Skip("Skipping test: /dev/tpm0 not found")
+	}
+
+	ExternalTPM = nil
```
alexmwu, line 97: You don't need to set ExternalTPM; it defaults to nil.

cmd/token_test.go
```diff
@@ -89,6 +89,46 @@ func TestTokenWithGCEAK(t *testing.T) {
...
+	var dummyMetaInstance = util.Instance{ProjectID: "test-project", ProjectNumber: "1922337278274", Zone: "us-central-1a", InstanceID: "12345678", InstanceName: "default"}
```
alexmwu, line 98: Sorry, one more thing: I see now you mean for this test to run only in GCE environments with TPMs. In that case, another test to add would be to copy the default event log location to a https://pkg.go.dev/testing#T.TempDir and then point the `event-log` flag there. Please post a copy of a successful test run as a comment too.

alexmwu (approved):

cmd/token_test.go
jkl73: I don't think we run cmd tests in GCE yet, maybe we can manually test it?

jkl73 (approved):

cmd/token_test.go
Ruide: Without this assignment, we would have a concurrency error. See commit e5558fc74b06f943ce723d87f7884c990e584eeb vs commit 0f65c4c6b6250735d7cfdbdac506bca5478ec666

cmd/token_test.go
Ruide: To copy `eventlog` file, the `go test` would need root privilege. We might not want this?

Ruide: I added a TestCopiedCustomEventLogFile for running manually with the following commands:
`go test -c -v ./cmd/...`
`sudo ./cmd.test -test.run TestCopiedCustomEventLogFile`

PR #435, opened 2024-04-11
fix invalid check and restore workaround from #72

savely-krasovsky:
It seems this check was not valid for a while. This PR fixes that and restores unsealing for the old devices.

### DIFF
--- a/client/keys.go
+++ b/client/keys.go
@@ -388,7 +388,8 @@ func (k *Key) Unseal(in *pb.SealedBytes, opts UnsealOpts) ([]byte, error) {
 		// ECC signing key.
 		// We can detect this bug, as it triggers a RCInsufficient
 		// Unmarshaling error.
-		if paramErr, ok := certErr.(tpm2.ParameterError); ok && paramErr.Code == tpm2.RCInsufficient {
+		var handleError tpm2.HandleError
+		if errors.As(certErr, &handleError) && handleError.Code == tpm2.RCInsufficient {
 			signer, err := AttestationKeyECC(k.rw)
 			if err != nil {
 				return nil, fmt.Errorf("failed to create fallback signing key: %w", err)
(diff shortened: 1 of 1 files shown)


savely-krasovsky:
@jkl73 please review! I have found this bug after our discussion from #72.

client/keys.go
```diff
@@ -388,7 +388,8 @@ func (k *Key) Unseal(in *pb.SealedBytes, opts UnsealOpts) ([]byte, error) {
 		// ECC signing key.
 		// We can detect this bug, as it triggers a RCInsufficient
 		// Unmarshaling error.
-		if paramErr, ok := certErr.(tpm2.ParameterError); ok && paramErr.Code == tpm2.RCInsufficient {
+		var handleError tpm2.HandleError
+		if errors.As(certErr, &handleError) && handleError.Code == tpm2.RCInsufficient {
```
jkl73, line 392: Can you also still include the old ParameterError check in the condition? Seem like different TPM implementations will return slightly different error code on this one.

jkl73 (changes requested):

jkl73:
Also just out of my curiosity, do you the actual model and manufacturer of the TPM you facing the issue with? (is it a firmware TPM or a standalone TPM)

client/keys.go
savely-krasovsky: I've added ParameterError, but left `error.As` pattern, because in my opinion this check will be more future-proof, at somepoint someone could wrap error above.

savely-krasovsky:
@jkl73 this is ST Microelectronics (STM).
Manufacturer version: `73.4.17568.4452`
Spec version: `2.0`
PPI spec version: `1.3`
Subversion of TPM spec: `1.16 (9.21.2016)`
Spec version of client computer: `1.00`

Sorry, I could translate something wrong, user gave me only a screenshot of Windows Defender window in Russian:
<details><summary>Screenshot</summary>
</img>
</details>

jkl73:
Thanks for fixing this!

PR #436, opened 2024-04-12
Error message should return length of digest

hkolvenbach:
The error message should return the length of the digest to compare against the length of the hash, instead of returning the whole digest.


jkl73 (approved):
Thanks for the fix!

PR #442, opened 2024-04-26
Allow server verification to accept tpm attestation evidence in remote attestation request from tdx and sevsnp machines

Ruide:
The current server.VerifyAttestation function enforce itself to only accept hardware attestation report when the event log record (event type == NonhostInfo) is 3 or 4 (standing for GCEConfidentialTechnology_INTEL_TDX & GCEConfidentialTechnology_AMD_SEV_SNP).

When server.VerifyAttestation calls VerifyGceTechnology, server.VerifyAttestation returns error when it cannot find a pb.Attestation_SevSnpAttestation or pb.Attestation_TdxAttestation in the received attestation evidence (pb.Attestation).

However, we want to still allow vTPM attestation on tdx and sev-snp machines. This PR is for removing the enforcement.

See more context in https://docs.google.com/document/d/1bqy3MJdbQrZzn0RQTo1WWZ9wqTXl_y7Jq7Wvnw8Oobo/edit?resourcekey=0-Pfgmgvm7HqCp37cOAu8pbw&tab=t.0#heading=h.mf27ppdi03e3.

### DIFF
--- a/server/verify.go
+++ b/server/verify.go
@@ -47,6 +47,11 @@ type VerifyOpts struct {
 	// Which bootloader the instance uses. Pick UNSUPPORTED to skip this
 	// parsing or for unsupported bootloaders (e.g., systemd).
 	Loader Bootloader
+	// AllowGCESoftwareTEEAttestation indicates if GCESoftwareTEEAttestation is
+	// allowed. If it is allowed, then TEE (e.g. sevsnp, tdx) hardware
+	// attestation evidence is not required. And Attestation struct does not
+	// need TEE Attestation (e.g. sevsnp.Attestation, tdx.QuoteV4).
+	AllowGCESoftwareTEEAttestation bool
 	// TEEOpts allows customizing the functionality of VerifyTEEAttestation.
 	// Its type can be *VerifySnpOpts if the TEEAttestation is a SevSnpAttestation
 	// or can be *VerifyTdxOpts if the TEEAttestation is a TdxAttestation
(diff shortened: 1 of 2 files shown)


server/verify.go
```diff
@@ -342,7 +342,7 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
 		var snpOpts *VerifySnpOpts
 		tee, ok := attestation.TeeAttestation.(*pb.Attestation_SevSnpAttestation)
 		if !ok {
-			return fmt.Errorf("TEE attestation is %T, expected a SevSnpAttestation", attestation.GetTeeAttestation())
```
alexmwu, line 345: should this be behind a flag? something like `RequireHardwareAttestation` or `RequireTEEHardwareAttestation`?

Ruide: Yea, a flag makes sense, let me add that!

alexmwu: Sorry, just had this thought: should this be the other way around? That way the default requires hw attestation. (e.g., `AllowSoftwareTEEAttestation` or `AllowGCESoftwareTEEAttestation` since it's only for GCE-based sw attestation)

alexmwu: (you'll need to add some new tests either way)

Ruide:
After discussion with @KeithMoyer, we DONOT want to allow vTPM attestation solo. Closing this PR. Reopen in future if assumption changes.

PR #446, opened 2024-05-03
Add SEV-SNP policy for signed UEFI measurements

deeglaze:
Depends on PR#445

This adds an extra validation check beyond well-formedness that the verification step checks. If the reference values are available within the SEV-SNP attestation certificate chain, then verify the signature and check the report measurement against the golden values.


deeglaze:
@jkl73 not sure what to make of the CS presubmit failure.

jkl73:
> @jkl73 not sure what to make of the CS presubmit failure.

just ignore the error now.. I'm still trying to fix the the build

jkl73:
> > @jkl73 not sure what to make of the CS presubmit failure.
> 
> just ignore the error now.. I'm still trying to fix the the build

@deeglaze could you rebase this PR and run the CS presubmit, I think I fixed it (hopefully..)

proto/attest.proto
```diff
@@ -250,9 +257,35 @@ message PlatformPolicy {
   GCEConfidentialTechnology minimum_technology = 3;
 }
 
+// A policy about what parts of a RIM to compare against machine state as
+// reflected in a quote or (verified) event log. Reference measurements for
+// a component are expected to be addressable by the machine state's reported
+// digest, or otherwise presented as cached collateral with the attestation
+// itself. The method of delivery is vendor-specific.
+message RIMPolicy {
+  // If true, the signed measurement must be available and the target
```
alexmwu, line 266: available where? IIUC, this is the attestation certificate table, but we state it.

proto/attest.proto
```diff
@@ -250,9 +257,35 @@ message PlatformPolicy {
...
+  // measurement must be among the listed signed measurements.
+  // If false, then only error if there is a problem verifying the signed
+  // measurements when they are available.
+  bool require = 1;
```
alexmwu, line 270: How about cert_entry_required`?

server/policy.go
```diff
@@ -13,7 +19,69 @@ import (
...
+	// and apply the verification logic against the attestation measurement.
+	opts := &svalidate.Options{
+		GuestPolicy: defaultSevSnpGuestPolicy,
+		CertTableOptions: map[string]*svalidate.CertEntryOption{
+			gcesev.GCEFwCertGUID: {
+				Kind: kind,
+				Validate: gceverify.SNPValidateFunc(&gceverify.Options{
+					RootsOfTrust: uefirot,
+					Now:          time.Now(),
+					// TODO: No getter? Network fallback can be helpful.
```
alexmwu, line 82: What specifically do we need to do here? Is it to introduce a Getter/an option for a Getter?

server/policy_test.go
```diff
@@ -151,3 +166,113 @@ func TestEvaluatePolicyFailure(t *testing.T) {
...
+	}
+
+	nonprod.RootCmd.SetArgs([]string{"endorse",
+		"--uefi", fwPath,
+		"--key_dir", keydir,
+		"--bucket_root", certdir,
+		"--root_path", "root.crt",
+		"--add_snp",
+		"--out_root", outdir,
+		"--snapshot_dir", "snap",
```
alexmwu, line 202: Seems a bit odd that we have to test through the CLI instead of through an underlying library

server/policy_test.go
```diff
@@ -151,3 +166,113 @@ func TestEvaluatePolicyFailure(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	dents, err := os.ReadDir(outdir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	dents2, err := os.ReadDir(outdir2)
+	if err != nil {
+		t.Fatal(err)
+	}
```
alexmwu, line 214: I might be missing something, but I don't see where we use `outdir2`

go.mod
```diff
@@ -18,10 +19,17 @@ require (
 	github.com/google/go-configfs-tsm v0.2.2 // indirect
 	github.com/google/go-tspi v0.3.0 // indirect
 	github.com/google/uuid v1.6.0 // indirect
+	github.com/inconshreveable/mousetrap v1.1.0 // indirect
```
JoshuaKrstic, line 22: this is a really bizarre dependency to see

server/policy.go
```diff
@@ -13,7 +19,69 @@ import (
 // will describe in what way the state failed. See the Policy documentation for
 // more information about the specifics of different policies.
 func EvaluatePolicy(state *pb.MachineState, policy *pb.Policy) error {
```
JoshuaKrstic, line 21: Me want rego policy

Not related to your change, but would you be open to this being completely moved to a rego policy in the near future?

server/policy.go
```diff
@@ -13,7 +19,69 @@ import (
...
+			return nil, err
+		}
+		pool.AddCert(cert)
+	}
+	return pool, nil
+}
+
+// the SEV-SNP attestation signature is already verified by this point.
+func evaluateSevSnpPolicy(state *spb.Attestation, policy *pb.SevSnpPolicy) error {
+	if policy == nil {
```
JoshuaKrstic, line 53: This feels like a scary way to represent any "everything" policy to me. I feel like this is really easily done accidentally.
I don't know enough about protocol buffers to have a suggestion here, so perhaps when we have a different sort of policy language this concern will no longer be relevant.
WDYT?

server/policy_test.go
```diff
@@ -151,3 +166,113 @@ func TestEvaluatePolicyFailure(t *testing.T) {
...
+func TestSevSnpSignedUefi(t *testing.T) {
+	// Generate fake signed reference measurements.
+	fw := ovmfsev.CleanExample(t, 0x200000)
+	meas, err := gcesev.LaunchDigest(gcesev.LaunchOptionsDefault(), fw)
+	if err != nil {
+		t.Fatal(err)
+	}
+	certdir := t.TempDir()
+	keydir := t.TempDir()
+	outdir := t.TempDir()
```
JoshuaKrstic, line 179: nit: can you define these closer to where they are used?

server/policy_test.go
```diff
@@ -151,3 +166,113 @@ func TestEvaluatePolicyFailure(t *testing.T) {
...
+	outdir2 := t.TempDir()
```
JoshuaKrstic, line 180: nit: can you define these closer to where they are used?

server/policy_test.go
```diff
@@ -151,3 +166,113 @@ func TestEvaluatePolicyFailure(t *testing.T) {
...
+	}
+
+	pol := &pb.Policy{
+		SevSnp: &pb.SevSnpPolicy{
+			Uefi: &pb.RIMPolicy{
+				RootCerts: [][]byte{rootDER},
+			},
+		},
+	}
+	if err := EvaluatePolicy(ms, pol); err != nil {
```
JoshuaKrstic, line 275: Consider adding tests cases for nil policy + nil UEFI policy

JoshuaKrstic (approved):
Overall LGTM. I have a small concern about the nil policy check

proto/attest.proto
```diff
@@ -250,9 +257,35 @@ message PlatformPolicy {
...
+  // x.509 certificates in ASN.1 DER format.
+  repeated bytes root_certs = 2;
+}
+
+// Represent minimal decisions about attestation fields until necessary to
+// add for policy reasons.
+message SevSnpPolicy {
+  // The policy for checking the signed reference values for the UEFI at launch.
+  RIMPolicy uefi = 1;
```
jkl73, line 279: Is it possible to make this policy not just specific to snp?

I wonder later if tdx can have a similar firmware policy check

proto/attest.proto
deeglaze: It can be in the certificate table, a UEFI variable, a local file, downloadable given a URI, or any other means that could be specified by a vendor. I've updated the wording a little.

proto/attest.proto
deeglaze: SGTM

server/policy_test.go
deeglaze: It's a CLI library. Using it this way is less boilerplate-y.

server/policy.go
deeglaze: How do you feel about `EvaluatePolicyOpt` with a new `*PolicyOptions` argument? The `EvaluatePolicy` can use default options like `Getter: trust.DefaultHTTPSGetter()` and `Now: time.Now()` (Now moved to options to make evaluateSevSnpPolicy more deterministic and testable).

server/policy_test.go
deeglaze: Removing..

server/policy_test.go
deeglaze: Removed outdir2.

server/policy_test.go
deeglaze: Thanks, found a bug.

server/policy.go
deeglaze: policy == nil should be understood as &pb.SevSnpPolicy{}. I've removed this check. The real concern is the `policy.GetUefi() == nil` check, since if I treat nil as &policy.RIMPolicy{}, then the behavior is different.

I can change the behavior to be like the evaluatePlatformPolicy, where at least one root has to be present to care about any signers. No signers means any value is okay. That would make the nil check and empty message the same.

I wouldn't want to start failing all policies because there are no signed measurements deployed yet. Do you want to depend on signed reference values right now? I suppose you could but then have failing tests internally until the signed firmware package finally rolls out.

server/policy.go
deeglaze: I would certainly. The issue is primarily populating the Rego context to know that the signed values are even available and their verification status.

PR #449, opened 2024-05-09
Refresh SA auth token in signaturediscovery client before fetching container image signatures

yawangwang:
Currently we only fetch the VM SA once, for the purpose of downloading workload image from Artifact registry.
But it doesn't work well for long running workloads because the VM SA token will expire after 1 hr, and thus lead to 401 errors when the signaturediscovery client tries to fetch image signatures periodically.
We'll need to refresh the resolver for signaturediscovery client to authenticate with the target docker repo before fetching container signatures.

# Breaking changes:
- move `auth.go` to from package `launcher` a to a new sub-package `github.com/google/go-tpm-tools/launcher/registryauth` so that auth utilities can be used cross packages. Add a new method that takes in a metadata server client and returns a refreshed remote resolver.
- refactor of signaturesdiscovery client to be able to refresh resolver before pulling docker image.

### DIFF
--- a/launcher/internal/signaturediscovery/client.go
+++ b/launcher/internal/signaturediscovery/client.go
@@ -8,12 +8,18 @@ import (
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/remotes"
 	"github.com/google/go-tpm-tools/verifier/oci"
 	"github.com/google/go-tpm-tools/verifier/oci/cosign"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 const signatureTagSuffix = "sig"
+
+type (
+	remoteResolverFetcher func(context.Context) (remotes.Resolver, error)
+	imageFetcher          func(context.Context, string, ...containerd.RemoteOpt) (containerd.Image, error)
+)
 
 // Fetcher discovers and fetches OCI signatures from the target repository.
 type Fetcher interface {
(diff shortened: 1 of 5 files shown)


launcher/internal/signaturediscovery/client.go
```diff
@@ -8,31 +8,38 @@ import (
...
 	"github.com/google/go-tpm-tools/verifier/oci"
 	"github.com/google/go-tpm-tools/verifier/oci/cosign"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/oauth2"
 )
 
 const signatureTagSuffix = "sig"
 
+type (
+	oauth2TokenFetcher func(context.Context) (oauth2.Token, error)
```
jkl73, line 21: Instead of having the logic in the signature client, can we move it to auth.go package? So the resolver can handle the token fetching itself

launcher/auth/auth.go
```diff
@@ -1,4 +1,5 @@
-package launcher
+// Package auth contains functionalities to authenticate docker repo.
+package auth
```
jkl73, line 2: Like having a new Resolver in this package that takes in mds client instead of the token

launcher/internal/signaturediscovery/client.go
yawangwang: Done.

launcher/auth/auth.go
yawangwang: Done.

jkl73 (approved):

launcher/auth/auth.go
```diff
@@ -1,8 +1,10 @@
-package launcher
+// Package auth contains functionalities to authenticate docker repo.
```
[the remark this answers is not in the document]
alexmwu: Please comment this as a breaking change in the  PR description

launcher/internal/signaturediscovery/client.go
```diff
@@ -8,31 +8,37 @@ import (
...
 type Fetcher interface {
 	FetchImageSignatures(ctx context.Context, targetRepository string) ([]oci.Signature, error)
 }
 
 // Client is a wrapper of containerd.Client to interact with signed image manifest.
 type Client struct {
-	cdClient          *containerd.Client
 	OriginalImageDesc v1.Descriptor
-	RemoteOpts        []containerd.RemoteOpt
+	refreshResolver   remoteResolverFetcher
```
alexmwu, line 32: `remoteResolverFetcher` or `refreshedResolverFetcher` to match the name below.

alexmwu (approved):

launcher/internal/signaturediscovery/client.go
yawangwang: Done.

launcher/auth/auth.go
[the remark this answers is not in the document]
yawangwang: Done.

PR #451, opened 2024-05-17
add custom nonce flag to cmd package token subcommand

Ruide:
- add a custom-nonce flag to allow user supplied string array for Custom Nonce
- added the flag in token subcommand for generating tokens with the custom nonce field
- added the flag in unit test

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -14,18 +14,19 @@ import (
 )
 
 var (
-	output    string
-	input     string
-	nvIndex   uint32
-	nonce     []byte
-	teeNonce  []byte
-	keyAlgo   = tpm2.AlgRSA
-	pcrs      []int
-	format    string
-	asAddress string
-	audience  string
-	eventLog  string
-	cloudLog  bool
+	output      string
+	input       string
+	nvIndex     uint32
+	nonce       []byte
+	teeNonce    []byte
+	keyAlgo     = tpm2.AlgRSA
+	pcrs        []int
+	format      string
+	asAddress   string
+	audience    string
+	eventLog    string
+	cloudLog    bool
+	customNonce []string
 )
 
 type pcrsFlag struct {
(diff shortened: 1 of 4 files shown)


cmd/token.go
```diff
@@ -138,7 +138,7 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 			Challenge:      challenge,
 			GcpCredentials: principalTokens,
 			Attestation:    attestation,
-			TokenOptions:   verifier.TokenOptions{CustomAudience: audience, TokenType: "OIDC"},
+			TokenOptions:   verifier.TokenOptions{CustomAudience: audience, CustomNonce: customNonce, TokenType: "OIDC"},
```
JoshuaKrstic, line 141: Does this mean that default token will have this custom nonce every time? What is this used for?

Ruide: The default value for `customNonce` is nil. Thus, the default token would not contain custom nonce. This is used for on-demand attestation for CVMs for customer to build upon.

JoshuaKrstic: Ah yes sorry, totally missed the directory this was in.

alexmwu (changes requested):
Please add a test that uses custom nonce and the fake verifier

JoshuaKrstic (approved):

cmd/flags.go
```diff
@@ -144,6 +145,12 @@ func addAudienceFlag(cmd *cobra.Command) {
 		"the audience field in the claims token. Cannot be sts.googleapis.com.")
 }
 
+// Lets this command specify custom nonce field of the attestation token.
+func addCustomNonceFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringArrayVar(&customNonce, "custom-nonce", nil,
+		"the custom nonce field in the claims token. use this flag multiple times to add multiple custom nonce.")
```
alexmwu, line 151: "the custom nonce field in the claims token. Use this flag multiple times to add multiple custom nonces."

verifier/util/fake_attestation_server.go
```diff
@@ -71,3 +88,27 @@ func NewMockAttestationServer() (*MockAttestationServer, error) {
...
+	}
+	defer r.Body.Close()
+
+	if err := protojson.Unmarshal(body, req); err != nil {
+		return fmt.Errorf("error decoding attestation request: %v", err)
+	}
+
+	if req.TokenOptions.Nonce != nil || req.TokenOptions.Audience != "" {
+		if req.TokenOptions.Nonce[0] != FakeCustomNonce[0] || req.TokenOptions.Nonce[1] != FakeCustomNonce[1] {
+			return fmt.Errorf("error comparing custom nonce: %v", err)
```
alexmwu, line 107: what about failure mode tests?

alexmwu (approved):

PR #452, opened 2024-05-17
Fix an uint conversion

jkl73:
https://github.com/google/go-tpm-tools/security/code-scanning/1

### DIFF
--- a/server/policy_constants.go
+++ b/server/policy_constants.go
@@ -135,7 +135,7 @@ func ConvertSCRTMVersionToGCEFirmwareVersion(version []byte) (uint32, error) {
 		asciiVersion = append(asciiVersion, b)
 	}
 
-	versionNum, err := strconv.Atoi(string(asciiVersion))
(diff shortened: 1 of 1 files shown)


server/policy_constants.go
```diff
@@ -135,7 +135,7 @@ func ConvertSCRTMVersionToGCEFirmwareVersion(version []byte) (uint32, error) {
 		asciiVersion = append(asciiVersion, b)
 	}
 
-	versionNum, err := strconv.Atoi(string(asciiVersion))
+	versionNum, err := strconv.ParseUint(string(asciiVersion), 10, 32)
 	if err != nil {
 		return 0, fmt.Errorf("when parsing GCE firmware version: %w", err)
 	}
```
alexmwu, line 141: You can remove the unnecessary cast to uint32 in the line below this

alexmwu (approved):

server/policy_constants.go
jkl73: Atoi returns uint64, so still need to cast it to uint32

PR #453, opened 2024-05-17
Add clean up step to image build process.

michael-pregman:
This PR adds a step to the build/test process that removes "presubmit" images 30+ days old from the confidential-space-images-dev project.


jkl73:
Instead of putting the cleanup code when submitting a PR, can we have it somewhere like cloud function or borgcron that can run periodically itself?

PR #455, opened 2024-05-22
Add version information and fix cloudbuild

jkl73:
* gotpm cmd can now shows version info:
```
$ go-tpm-tools_linux_amd64_v1/go-tpm-tools -v
gotpm version 0.4.4, commit 2dca42d8bece8ce657cd972d4451e3ba803b8e96, built at 2024-05-22T20:46:34Z
- go-sev-guest version v0.3.1
- go-tdx-guest version v0.9.3
```

* launcher will also show build commit in the welcome message.

* update launcher cloudbuild to go1.21 to match its go.mod

* add "sync" at the end of the `preload.sh` file to fix (hopefully) the empty files in OEM partition when building the image.
```
/usr/share/oem/confidential_space $ ls -althr
total 80M
drwxr-xr-x 4 root root 4.0K May 28 20:51 ..
-rwxr-xr-x 1 root root  55M May 28 20:51 confidential_space_experiments
-rwxr-xr-x 1 root root  26M May 28 20:51 cs_container_launcher
-rw-r--r-- 1 root root    0 May 28 20:51 container-runner.service
-rwxr-xr-x 1 root root    0 May 28 20:51 exit_script.sh
-rw-r--r-- 1 root root    0 May 28 20:51 fluent-bit-cs.conf
-rw-r--r-- 1 root root    0 May 28 20:51 system-stats-monitor-cs.json
-rw-r--r-- 1 root root    0 May 28 20:51 boot-disk-size-consistency-monitor-cs.json
-rw-r--r-- 1 root root    0 May 28 20:51 docker-monitor-cs.json
-rw-r--r-- 1 root root    0 May 28 20:51 kernel-monitor-cs.json
drwxr-xr-x 2 root root 4.0K May 28 20:51 .
```

### DIFF
--- a/cmd/gotpm/main.go
+++ b/cmd/gotpm/main.go
@@ -2,12 +2,43 @@
 package main
 
 import (
+	"fmt"
 	"os"
+	"runtime/debug"
 
 	"github.com/google/go-tpm-tools/cmd"
 )
 
+// GoReleaser will populates those fields
+// https://goreleaser.com/cookbooks/using-main.version/
+var (
+	version = "dev"
+	commit  = "none"
+	date    = "unknown"
+)
+
+var (
+	tdxGuestVersion = "unknown"
+	sevGuestVersion = "unknown"
+	sevGuest        = "github.com/google/go-sev-guest"
+	tdxGuest        = "github.com/google/go-tdx-guest"
+)
+
 func main() {
+	if info, ok := debug.ReadBuildInfo(); ok {
+		for _, dep := range info.Deps {
+			switch dep.Path {
+			case sevGuest:
+				sevGuestVersion = dep.Version
+			case tdxGuest:
+				tdxGuestVersion = dep.Version
+			}
+		}
+	}
+
+	cmd.RootCmd.Version = fmt.Sprintf("%s, commit %s, built at %s\n- go-sev-guest version %s\n- go-tdx-guest version %s",
+		version, commit, date, tdxGuestVersion, sevGuestVersion)
+
 	if cmd.RootCmd.Execute() != nil {
 		os.Exit(1)
 	}
--- a/launcher/image/preload.sh
+++ b/launcher/image/preload.sh
@@ -128,6 +128,12 @@ main() {
          "Only 'debug' and 'hardened' are supported."
     exit 1
   fi
+
+  # Make sure cache is flushed for the OEM partition.
+  sync ${OEM_PATH}
+
+  # Verify the content before the OEM sealing step.
+  ls -lh ${CS_PATH}
 }
 
 main
--- a/launcher/launcher/main.go
+++ b/launcher/launcher/main.go
@@ -45,6 +45,9 @@ var rcMessage = map[int]string{
 	holdRC:    "VM remains running",
 }
 
+// BuildCommit shows the commit when building the binary, set by -ldflags when building
+var BuildCommit = "dev"
+
 var logger *log.Logger
 var mdsClient *metadata.Client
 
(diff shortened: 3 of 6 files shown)


cmd/gotpm/main.go
```diff
@@ -2,12 +2,42 @@
...
 
 import (
+	"fmt"
 	"os"
+	"runtime/debug"
 
 	"github.com/google/go-tpm-tools/cmd"
 )
 
+// GoReleaser will populates those fields
```
alexmwu, line 12: how does it do this? I see https://pkg.go.dev/runtime/debug#ReadBuildInfo, does GoReleaser fill that out?

cmd/gotpm/main.go
```diff
@@ -2,12 +2,42 @@
...
+var (
+	version = "dev"
+	commit  = "none"
+	date    = "unknown"
```
alexmwu, line 16: How does goreleaser populate these fields?

launcher/image/preload.sh
```diff
@@ -128,6 +128,12 @@ main() {
          "Only 'debug' and 'hardened' are supported."
     exit 1
   fi
+
+  # make sure files are copied
+  sync
```
alexmwu, line 133: Why is syncing required? Should we only sync the copy directory?

launcher/launcher/main.go
```diff
@@ -73,7 +76,7 @@ func main() {
 	defer serialConsole.Close()
 	logger.SetOutput(io.MultiWriter(os.Stdout, serialConsole))
 
-	logger.Println(welcomeMessage)
+	logger.Printf("%s, version: %s\n", welcomeMessage, Version)
```
alexmwu, line 79: I think we sometimes key off the welcomeMessage. I'm not sure if it's a regex or pattern match, but, to be safe, move it to another line. 
```
logger.Println(welcomeMessage)
logger.Printf("Build commit: %s\n", Commit)
```

launcher/launcher/main.go
```diff
@@ -45,6 +45,9 @@ var rcMessage = map[int]string{
 	holdRC:    "VM remains running",
 }
 
+// Version shows the commit version, set by -ldflags when building
+var Version = "dev"
```
alexmwu, line 49: nit: call this `Commit` or `BuildCommit`

cmd/gotpm/main.go
jkl73: yes, by default these three will be populated https://goreleaser.com/cookbooks/using-main.version/

cmd/gotpm/main.go
jkl73: when building the binary, using ldflags:
https://goreleaser.com/cookbooks/using-main.version/

launcher/image/preload.sh
jkl73: So the symptom we saw in some image builds are files with 0 length, I thought this might be because of file didn't copy through and the partition got umounted, but I'm not 100% sure.

launcher/launcher/main.go
jkl73: Done

launcher/launcher/main.go
jkl73: done

launcher/image/preload.sh
jkl73: I changed to sync the oem directory only

cmd/gotpm/main.go
alexmwu: consider adding a comment

alexmwu (approved):

PR #456, opened 2024-05-23
Increasing MAX_NV_BUFFER_SIZE to 2048

Daviiap:
To match most vTPMs implementations, the MAX_NV_BUFFER_SIZE has been increased to 2048


jkl73:
We use the simulator mainly for running tests.

Did you run into any out-of-nvbuffer issue?

Instead of changing the header file directly, I prefer to let it stay the same as the upstream. Could you add a CFLAGS in `internal_cgo.go` to  verwrite macro? like
```
// #cgo CFLAGS: -DMAX_NV_BUFFER_SIZE=2048
// #cgo CFLAGS: -DNV_MEMORY_SIZE=32768
```

PR #458, opened 2024-06-01
[launcher] Try to fix cloudbuild for launcher

jkl73:
Still experiencing empty files in OEM partition, try more logging and careful mount in the script.

Also moved `cloudbuild.yaml` from the root directory to launcher/ directory


launcher/image/cloudbuild.yaml
```diff
@@ -54,7 +54,4 @@ steps:
 timeout: '3000s'
 
 options:
-  logging: CLOUD_LOGGING_ONLY
   dynamic_substitutions: true
-  pool:
-    name: 'projects/confidential-space-images-dev/locations/us-west1/workerPools/cs-image-build-vpc'
```
jkl73, line 60: Remove this private pool requirement, so people can submit the image build in their own project

alexmwu: Why did we have this?

launcher/image/fixup_oem.sh
```diff
@@ -26,6 +26,19 @@ main() {
     sed -e 's|.*|[BEGIN_CS_CMDLINE]&[END_CS_CMDLINE]|g'
 
   umount /mnt/disks/efi
+
+  # Now the oem partition is sealed, we mount it to print it's content
+  if [[ ! -d /mnt/disks/oem ]]; then
+    mkdir /mnt/disks/oem
```
alexmwu, line 32: Cleanup this directory?

launcher/image/preload.sh
```diff
@@ -132,8 +132,12 @@ main() {
   # Make sure cache is flushed for the OEM partition.
   sync ${OEM_PATH}
 
+  # Remount as read-only to avoid unexpected changes
```
alexmwu, line 135: Why do we need this? Are you worried that some other thing is modifying the OEM partition after the fact? Wouldn't that get reflected in the dm-verity?

jkl73: yes, the behavior I saw is the file is all good before the preload.sh finish, but somehow got corrupted before the dm-verity sealing. So dm-verity looks all good, but the files are still corrupted.

Remount it as read-only (as it should be before this script) just to make sure it didn't get written accidently

launcher/image/fixup_oem.sh
jkl73: I think it is fine as this is not get persistent, just a tmp build environment

launcher/image/cloudbuild.yaml
jkl73: Used to have this to build images in the private pool. But I realize it's fine to have the image building step without private pool, just put tests in the private pool.

Good thing is people can now use this cloudbuild file directly to build image in their own project for testing and debugging

alexmwu (approved):

PR #460, opened 2024-06-10
Add explicit SEV-SNP measurements policy

deeglaze:
When reference values come from a small database, allow checking against values that are known ahead of time, without needing to check signatures.

This allows the Google Attestation Verifier to use its reference measurements database without needing to trawl all the signed reference values we published.


proto/attest.proto
```diff
@@ -277,6 +277,9 @@ message RIMPolicy {
 message SevSnpPolicy {
   // The policy for checking the signed reference values for the UEFI at launch.
   RIMPolicy uefi = 1;
+  // The policy that any unsigned reference measurement must be within this
```
JoshuaKrstic, line 280: nit: is this a policy?

PR #463, opened 2024-06-25
Populate the SNP/TDX Machine State field with the verified SNP/TDX attestation data + use a stable COS image version

yawangwang:
This PR includes changes to populate the machinState proto with the verified SNP/TDX attestation data. The attestation verifier service will leverage the verified machineState to make claims around.

# Breaking changes:
- Unexport `verifyGceTechnology` because GCETechnology event log is closely related to TPM, this function is directly called after the event log is parsed.
- Use a stable COS image version `cos-dev-117-18374-0-0` rather than the latest COS image version to avoid CS image build issues.

### DIFF
--- a/launcher/cloudbuild.yaml
+++ b/launcher/cloudbuild.yaml
@@ -1,7 +1,7 @@
 substitutions:
   # using this base image for now, because there is an issue causing the newest COS dev
   # image not booting with cs.
-  '_BASE_IMAGE': '' # left empty means using the latest image in the family
+  '_BASE_IMAGE': 'cos-dev-117-18374-0-0' # Use a stable COS version to avoid CS image build issues
   '_BASE_IMAGE_FAMILY': 'cos-dev' # base image family
   '_OUTPUT_IMAGE_PREFIX': 'confidential-space'
   '_OUTPUT_IMAGE_SUFFIX': ''
--- a/server/verify.go
+++ b/server/verify.go
@@ -8,6 +8,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/google/go-sev-guest/proto/sevsnp"
+	"github.com/google/go-tdx-guest/proto/tdx"
 	"github.com/google/go-tpm-tools/internal"
 	pb "github.com/google/go-tpm-tools/proto/attest"
 	tpmpb "github.com/google/go-tpm-tools/proto/tpm"
@@ -148,14 +150,9 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
 
 		// Parse event logs and replay the events against the provided PCRs
 		pcrs := quote.GetPcrs()
-		state, err := parsePCClientEventLog(attestation.GetEventLog(), pcrs, opts.Loader)
+		state, err := parseMachineStateFromTPM(attestation, pcrs, opts)
 		if err != nil {
-			lastErr = fmt.Errorf("failed to validate the PCClient event log: %w", err)
-			continue
-		}
-
-		if err := VerifyGceTechnology(attestation, state.Platform.GetTechnology(), &opts); err != nil {
-			lastErr = fmt.Errorf("failed to verify memory encryption technology: %w", err)
+			lastErr = fmt.Errorf("failed to parse machine state from TCG event log: %w", err)
 			continue
 		}
 
@@ -318,10 +315,10 @@ func makePool(certs []*x509.Certificate) *x509.CertPool {
 	return pool
 }
 
-// VerifyGceTechnology checks the GCE-specific GceNonHost event's Trusted Execution Technology (TEE)
+// verifyGceTechnology checks the GCE-specific GceNonHost event's Trusted Execution Technology (TEE)
 // claim using attestation reports if the technology supports them, and only then validates that a
 // particular technology has proven that it is in use.
-func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTechnology, opts *VerifyOpts) error {
+func verifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTechnology, opts *VerifyOpts) error {
 	switch tech {
 	case pb.GCEConfidentialTechnology_NONE: // Nothing to verify
 		if opts.TEEOpts != nil {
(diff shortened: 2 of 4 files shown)


JoshuaKrstic (approved):

server/verify.go
```diff
@@ -321,54 +325,68 @@ func makePool(certs []*x509.Certificate) *x509.CertPool {
 // VerifyGceTechnology checks the GCE-specific GceNonHost event's Trusted Execution Technology (TEE)
 // claim using attestation reports if the technology supports them, and only then validates that a
 // particular technology has proven that it is in use.
-func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTechnology, opts *VerifyOpts) error {
+func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTechnology, opts *VerifyOpts) (*pb.MachineState, error) {
```
alexmwu, line 328: This should go in `parsePCClientEventLog` since it depends on the resulting state. Long term, we will have branching to attempt to get a full MachineState from TeeAttestation. If successful, we shouldn't need to read TPM-related data. If unsuccessful (e.g., TDX machine without RTMR support or SNP machine right now), then we merge in TPM MachineState.

alexmwu (changes requested):

server/verify.go
yawangwang: Discussed offline, added a wrapper function around `parsePCClientEventLog` and `VerifyGceTechnology`

server/verify.go
```diff
@@ -372,3 +376,48 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
 	}
 	return fmt.Errorf("unknown GCEConfidentialTechnology: %v", tech)
 }
+
+// parseTCGEventLog is a wrapper function around `parsePCClientEventLog` method to:
+// 1. parse partial machine state from TPM TCG event logs.
+// 2. verify GceTechnology since the GCE Technology event is directly related to the TPM.
+func parseTCGEventLog(attestation *pb.Attestation, pcrs *tpmpb.PCRs, opts VerifyOpts) (*pb.MachineState, error) {
```
alexmwu, line 383: I liked your suggestion of `parseMachineStateFromTPM` better than `parseTCGEventLog`. Since it is also doing event log replay and event parsing. The term "parse" is a bit overloaded unfortunately.

server/verify.go
```diff
@@ -176,6 +179,7 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
 
 		proto.Merge(machineState, celState)
 		proto.Merge(machineState, state)
+		proto.Merge(machineState, teeState)
```
alexmwu, line 182: We should avoid proto.Merge where possible as it can override existing values.

server/verify.go
```diff
@@ -372,3 +376,48 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
...
+		return nil, fmt.Errorf("failed to validate the PCClient event log: %w", err)
+	}
+
+	if err := verifyGceTechnology(attestation, ms.GetPlatform().Technology, &opts); err != nil {
+		return nil, fmt.Errorf("failed to verify memory encryption technology: %w", err)
+	}
+	return ms, nil
+}
+
+// parseTEEAttestation parses a machineState from TeeAttestation.
```
alexmwu, line 395: The point of the wrapper func for parsePCClientEventlog is to avoid proto.Merge. proto.Merge is less readable since it implicitly copies every field from src to dst. This requires our implementation to be mutually exclusive, or we risk overwriting. We originally used proto.Merge to merge the state from two separate event logs, but we could just populate the TeeAttestation field directly in the wrapper.

server/verify.go
```diff
@@ -318,10 +322,10 @@ func makePool(certs []*x509.Certificate) *x509.CertPool {
 	return pool
 }
 
-// VerifyGceTechnology checks the GCE-specific GceNonHost event's Trusted Execution Technology (TEE)
```
alexmwu, line 321: If we unexport this, it's a backwards incompatible change. Please add a breaking comment to the PR description.

server/verify.go
alexmwu: IOW, this can be inline in parseTCGEventLog.

server/verify_test.go
```diff
@@ -1068,10 +1068,14 @@ func TestVerifyAttestationWithSevSnp(t *testing.T) {
 			}
```
alexmwu, line 1068: Please add tests for the failure cases

alexmwu (changes requested):

server/verify.go
yawangwang: Done.

server/verify.go
yawangwang: My take from "inline" is call `parseTEEAttestation` from `parseMachineStateFromTPM` after `verifyGceTechnolgy` is done. LMK if this is not what you want.

server/verify.go
yawangwang: Done.

server/verify.go
yawangwang: Done.

server/verify_test.go
yawangwang: Added a new test `TestParseTEEAttestation` to cover both successful and failure cases.

launcher/cloudbuild.yaml
```diff
@@ -1,7 +1,7 @@
 substitutions:
   # using this base image for now, because there is an issue causing the newest COS dev
   # image not booting with cs.
-  '_BASE_IMAGE': '' # left empty means using the latest image in the family
+  '_BASE_IMAGE': 'cos-dev-117-18374-0-0' # Use a stable COS version to avoid CS image build issues
```
alexmwu, line 4: Please do not submit until you get approval from @jkl73 for this change.

server/verify.go
```diff
@@ -372,3 +369,56 @@ func VerifyGceTechnology(attestation *pb.Attestation, tech pb.GCEConfidentialTec
...
+	ms.TeeAttestation = state.TeeAttestation
+	return ms, nil
+}
+
+// parseTEEAttestation parses a machineState from TeeAttestation.
+// For now it simply populates the machineState TeeAttestation field with the verified TDX/SNP data.
+// In long term, it should parse a full machineState from TeeAttestation.
+func parseTEEAttestation(attestation *pb.Attestation, tech pb.GCEConfidentialTechnology) (*pb.MachineState, error) {
+	switch tech {
+	case pb.GCEConfidentialTechnology_AMD_SEV_SNP:
```
alexmwu, line 401: I may be missing something: don't we already do these checks in verifyGceTechnology?? Can we  have that return pb.MachineState or just remove the call to parseTEEAttestation and do
`ms.TeeAttestation = attestation.TeeAttestation`?

alexmwu (approved):

launcher/cloudbuild.yaml
jkl73: lgtm, this is the same base image on the last dev build.

jkl73 (approved):

server/verify.go
yawangwang: `ms.TeeAttestation = attestation.TeeAttestation` will not work because they're two different proto types.
`parseTEEAttestation` doesn't do any checks on attestation. This function, in short term, just populates the TeeAttestation field using a deep copy of the verified attestation data, but in long run, we'll want to get a full machineState from TDX/SNP attestation. The intent of having this function is for future-proof purpose.

PR #471, opened 2024-07-25
Adding EV_EVENT_TAG support for PCR9

yawangwang:
We see new Tagged Event log events using EV_EVENT_TAG measured in PCR9 in the latest Ubuntu 24.04 Kernel https://github.com/torvalds/linux/blob/1722389b0d863056d78287a120a1d6cadb8d4f7b/drivers/firmware/efi/libstub/efi-stub-helper.c#L263-L274. We should update our eventlog parser to support that new event type.


server/eventlog.go
```diff
@@ -425,8 +425,8 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
 			continue
 		}
 
-		if event.GetUntrustedType() != IPL {
-			return nil, fmt.Errorf("invalid event type for PCR%d, expected EV_IPL", index)
+		if event.GetUntrustedType() != IPL && event.GetUntrustedType() != EventTag {
```
jkl73, line 428: Should we explicitly allow/ignore "EventTag" only on PCR9?

Here if PCR is 8 it'll not throw an error and may fail later in this loop.

@alexmwu

alexmwu: I think what we have here is fine. according to the PFP (https://trustedcomputinggroup.org/wp-content/uploads/TCG-PC-Client-Platform-Firmware-Profile-Version-1.06-Revision-52_pub-2.pdf), EV_EVENT_TAG is "Used for PCRs defined for OS and
application usage." Which are PCRs 8 and higher. https://github.com/systemd/systemd/blob/main/docs/TPM2_PCR_MEASUREMENTS.md uses it for PCR 12.

alexmwu (approved):

server/eventlog.go
alexmwu: Oh I see your point. This shouldn't parse the EV_EVENT_TAG event, since it likely comes from Linux. Right now we put the digest of the EV_EVENT_TAG event into the GRUB files, which isn't right

alexmwu (changes requested):
Also I notice we don't have a MachineState test. Please add one and confirm the digest of `LOADED_IMAGE::LoadOptions` is not there

server/eventlog.go
alexmwu: So, put another way. If we see an EV_EVENT_TAG, we should both not throw an error AND skip this event.

yawangwang:
> Also I notice we don't have a MachineState test. Please add one and confirm the digest of `LOADED_IMAGE::LoadOptions` is not there

Done

server/eventlog.go
yawangwang: Change the logic a bit to skip parsing EV_EVENT_TAG

server/eventlog.go
```diff
@@ -425,6 +425,11 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
 			continue
 		}
 
+		// skip parsing EV_EVENT_TAG event since it likely comes from Linux.
+		if event.GetUntrustedType() == EventTag {
```
jkl73, line 429: I'm still little bit doubt to skip an event based on the event type. As the event type is unverified and thus untrusted. An attack may try to change a IPL event type to a EvenTag event type to skip the grub state check.

So I think we should also verify the digest and content of this event here (assuming it is possible), and then we can skip the event.

server/policy_constants.go
```diff
@@ -18,13 +18,18 @@ import (
...
 	SCRTMVersion               uint32 = 0x00000008
 	IPL                        uint32 = 0x0000000D
 	NonhostInfo                uint32 = 0x00000011
 	EFIBootServicesApplication uint32 = 0x80000003
 	EFIAction                  uint32 = 0x80000007
 )
 
+// EventTagHex used with type "EV_EVENT_TAG".
+// This corresponds to a TLV struct of type LOAD_OPTIONS_EVENT_TAG_ID (0x8F3B22ED, reversed endian), length 0x1a (26), value `LOADED_IMAGE::LoadOptions\n`.
+const EventTagHex = "ed223b8f1a0000004c4f414445445f494d4147453a3a4c6f61644f7074696f6e7300"
```
jkl73, line 31: nit: EventTagLoadedImageHex to be more specific

jkl73 (approved):

server/policy_constants.go
yawangwang: Done.

yawangwang:
> Also I notice we don't have a MachineState test. Please add one and confirm the digest of `LOADED_IMAGE::LoadOptions` is not there

Done

.github/workflows/ci.yml
```diff
@@ -101,8 +101,11 @@ jobs:
         with:
           version: latest
           working-directory: ${{ matrix.dir }}
+          # Temporarily disable staticcheck till `elliptic.GenerateKey` is updated to a proper method.
```
alexmwu, line 104: We should fix the issue or disable the check for that line rather than disable staticcheck altogether. It's okay if you need to submit with that failing here. It looks like we can't do a simple migration since P224 is not supported by crypto/ecdh. So just add the comment to import.go:134
```	//nolint:staticcheck crypto/ecdh does not support P-224, while we do.
	// We need to remove support for curve 224 to remove this comment.
```.

Same with line 137

File: server/eventlog.go [hunk 5]
```diff
@@ -416,6 +416,11 @@ func getSecureBootState(attestEvents []attest.Event) (*pb.SecureBootState, error
 	}, nil
 }
 
+func isValidEventTag(event *pb.Event) bool {
+	// TODO: verify the event digest once https://lore.kernel.org/all/20240308085754.476197-7-ardb+git@google.com/T/ is patched.
```
alexmwu, line 420: We can't do this since it's not compatible with any unpatched images. Either we need to check for both the buggy measurement AND the fixed measurement, or we need to do what we're doing now and skip the check. I still maintain the check is not security critical, so I'd just remove the TODO.

@jkl73

server/eventlog.go
```diff
@@ -425,8 +430,14 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
 			continue
 		}
 
-		if event.GetUntrustedType() != IPL {
-			return nil, fmt.Errorf("invalid event type for PCR%d, expected EV_IPL", index)
+		// Check if the event type is EV_IPL or a valid EV_EVENT_TAG.
+		if event.GetUntrustedType() != IPL && !isValidEventTag(event) {
```
alexmwu, line 434: This logic is very complex. I don't think we need to check the validity of the event. Not all events have digests that match the event data. See the "GrubFile" parsing. Really, we should check these events are all pre-ExitBootServices if we want to be sure all of these events came from a BootLoader. There is no *guarantee* it's from GRUB even then. For this, we rely on SecureBoot or checking the measurements in EfiState. See getEfiState.

alexmwu (changes requested):

server/eventlog.go
jkl73: sgtm

.github/workflows/ci.yml
yawangwang: Done. Did some experiments on its format, it turned out 
```
//nolint:staticcheck
// crypto/ecdh does not support P-224, while we do. We need to remove support for curve 224 to remove this comment.
```
is the correct one to skip staticcheck.

server/eventlog.go
yawangwang: Done.

server/eventlog.go
yawangwang: Done.
Instead of checking the validity of the EV_EVENT_TAG, we'll just skip this type of event.

internal/test/test_data.go
```diff
@@ -18,6 +18,8 @@ var (
 	Ubuntu2104NoDbxEventLog []byte
 	//go:embed eventlogs/ubuntu-2104-no-secure-boot.bin
 	Ubuntu2104NoSecureBootEventLog []byte
+	//go:embed eventlogs/ubuntu-2404-amd-sevsnp.bin
+	Ubuntu2404AmdSevSnptEventLog []byte
```
alexmwu, line 22: Ubuntu2404AmdSevSnpEventLog

server/eventlog.go
```diff
@@ -599,3 +604,11 @@ func getGrubKernelCmdlineSuffix(grubCmd []byte) int {
 	}
 	return -1
 }
+
+func decodeHex(hexStr string) []byte {
```
alexmwu, line 608: nit: put this back in the test file.

alexmwu (approved):

internal/test/test_data.go
yawangwang: Good catch.

server/eventlog.go
yawangwang: Done

PR #474, opened 2024-08-09
Send client logs with the cloud logging library

jessieqliu:
Replaces our existing logging (with the "log" package) with a custom logging library that wraps (1) the Cloud Logging library (for logging to Cloud Logs) and (2) slog (for logging to the serial console). Using the Cloud Logging library allows for structured payloads and properly displayed severity (ex. ERROR logs) in the logging UI.

Note this does NOT include logging redirection from the workload - that is still sent via `StdOut` (in the case of Cloud Logging).

## Examples
Cloud Logs
Screenshot 2024-10-11 3 05 42 PM
image

Serial Logs
Screenshot 2024-10-11 3 05 17 PM

# Breaking Changes
- Changes `*log.Logger` type to `logging.Logger` in `agent.CreateAttestationAgent()`, `agent.fetchContainerimageSignatures()`, `ContainerRunner.logger`, `runner.NewRunner()`, `spec.GetLaunchSpec()`, `spec.fetchExperiments()`, and `teeserver.New()`

### DIFF
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -187,10 +187,10 @@ func fetchContainerImageSignatures(ctx context.Context, fetcher signaturediscove
 				},
 				retry(),
 				func(err error, _ time.Duration) {
-					logger.Printf("Failed to fetch container image signatures from repo %q: %v", targetRepo, err)
+					logger.Error("Failed to fetch container image signatures from repo: "+err.Error(), slog.String("repo", targetRepo))
 				})
 			if err != nil {
-				logger.Printf("Failed all attempts to refresh container signatures from repo %q: %v", targetRepo, err)
+				logger.Error("Failed all attempts to refresh container signatures from repo: "+err.Error(), slog.String("repo", targetRepo))
 			} else {
 				signatures[index] = sigs
 			}
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -186,10 +186,10 @@ func fetchContainerImageSignatures(ctx context.Context, fetcher signaturediscove
 				},
 				retry(),
 				func(err error, _ time.Duration) {
-					logger.Error("Failed to fetch container image signatures from repo: "+err.Error(), "repo", targetRepo)
+					logger.Error(fmt.Sprintf("Failed to fetch container image signatures from repo: %v", err.Error()), "repo", targetRepo)
 				})
 			if err != nil {
-				logger.Error("Failed all attempts to refresh container signatures from repo: "+err.Error(), "repo", targetRepo)
+				logger.Error(fmt.Sprintf("Failed all attempts to refresh container signatures from repo: %v", err.Error()), "repo", targetRepo)
 			} else {
 				signatures[index] = sigs
 			}
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -11,8 +11,6 @@ import (
 	"crypto"
 	"fmt"
 	"io"
-	"log"
-	"log/slog"
 	"net/http"
 	"sync"
 	"time"
@@ -20,6 +18,7 @@ import (
 	"github.com/cenkalti/backoff/v4"
 	"github.com/google/go-tpm-tools/cel"
 	"github.com/google/go-tpm-tools/client"
+	"github.com/google/go-tpm-tools/launcher/internal/logging"
 	"github.com/google/go-tpm-tools/launcher/internal/signaturediscovery"
 	"github.com/google/go-tpm-tools/launcher/spec"
 	pb "github.com/google/go-tpm-tools/proto/attest"
@@ -59,7 +58,7 @@ type agent struct {
 	sigsFetcher      signaturediscovery.Fetcher
 	cosCel           cel.CEL
 	launchSpec       spec.LaunchSpec
-	logger           *slog.Logger
+	logger           logging.Logger
 	sigsCache        *sigsCache
 }
 
@@ -70,7 +69,7 @@ type agent struct {
 // - principalFetcher is a func to fetch GCE principal tokens for a given audience.
 // - signaturesFetcher is a func to fetch container image signatures associated with the running workload.
 // - logger will log any partial errors returned by VerifyAttestation.
-func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher, verifierClient verifier.Client, principalFetcher principalIDTokenFetcher, sigsFetcher signaturediscovery.Fetcher, launchSpec spec.LaunchSpec, logger *slog.Logger) (AttestationAgent, error) {
+func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher, verifierClient verifier.Client, principalFetcher principalIDTokenFetcher, sigsFetcher signaturediscovery.Fetcher, launchSpec spec.LaunchSpec, logger logging.Logger) (AttestationAgent, error) {
 	// Fetched the AK and save it, so the agent doesn't need to create a new key everytime
 	ak, err := akFetcher(tpm)
 	if err != nil {
@@ -168,7 +167,7 @@ func (a *agent) Refresh(ctx context.Context) error {
 	return nil
 }
 
-func fetchContainerImageSignatures(ctx context.Context, fetcher signaturediscovery.Fetcher, targetRepos []string, retry func() backoff.BackOff, logger *log.Logger) []oci.Signature {
+func fetchContainerImageSignatures(ctx context.Context, fetcher signaturediscovery.Fetcher, targetRepos []string, retry func() backoff.BackOff, logger logging.Logger) []oci.Signature {
 	signatures := make([][]oci.Signature, len(targetRepos))
 
 	var wg sync.WaitGroup
@@ -187,10 +186,10 @@ func fetchContainerImageSignatures(ctx context.Context, fetcher signaturediscove
 				},
 				retry(),
 				func(err error, _ time.Duration) {
-					logger.Error("Failed to fetch container image signatures from repo: "+err.Error(), slog.String("repo", targetRepo))
+					logger.Error("Failed to fetch container image signatures from repo: "+err.Error(), "repo", targetRepo)
 				})
 			if err != nil {
-				logger.Error("Failed all attempts to refresh container signatures from repo: "+err.Error(), slog.String("repo", targetRepo))
+				logger.Error("Failed all attempts to refresh container signatures from repo: "+err.Error(), "repo", targetRepo)
 			} else {
 				signatures[index] = sigs
 			}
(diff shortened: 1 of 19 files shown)


launcher/launcher/main.go
```diff
@@ -42,76 +44,81 @@ var rcMessage = map[int]string{
...
-	if err = startLauncher(ctx, launchSpec, serialConsole); err != nil {
-		logger.Println(err)
+	if err = startLauncher(launchSpec, logger.SerialConsoleFile()); err != nil {
+		logger.Error(err.Error())
 	}
 
+	workloadDuration := time.Since(start)
+	logger.Info("Workload completed",
+		"workload", launchSpec.ImageRef,
+		"latency_sec", workloadDuration.Seconds(),
```
jkl73, line 119: "elapsed_sec" ?

launcher/launcher/main.go
```diff
@@ -171,18 +193,26 @@ func startLauncher(ctx context.Context, launchSpec spec.LaunchSpec, serialConsol
 	}
 	gceAk.Close()
 
-	token, err := registryauth.RetrieveAuthToken(ctx, mdsClient)
+	token, err := registryauth.RetrieveAuthToken(context.Background(), mdsClient)
+	if err != nil {
+		logger.Info(fmt.Sprintf("failed to retrieve auth token: %v, using empty auth for image pulling\n", err))
+	}
+
+	uptime, err := getUptime()
```
jkl73, line 201: why need to get the system up time here? I thought we want the seconds since `main()` started?

launcher/launcher/main.go
```diff
@@ -42,76 +44,81 @@ var rcMessage = map[int]string{
...
+var logger logging.Logger
 var mdsClient *metadata.Client
 
 var welcomeMessage = "TEE container launcher initiating"
 var exitMessage = "TEE container launcher exiting"
 
+var start time.Time
+
 func main() {
+	start = time.Now()
```
jkl73, line 56: seems like this `start` is never used? it will be overwritten in line 215

launcher/launcher/main.go
jkl73, line 119: not sure if "latency_sec" is set somewhere else, but maybe "elapsed_sec" is more suitable in the context?

launcher/launcher/main.go
```diff
@@ -171,18 +193,26 @@ func startLauncher(ctx context.Context, launchSpec spec.LaunchSpec, serialConsol
...
-	ctx = namespaces.WithNamespace(ctx, namespaces.Default)
+	ctx := namespaces.WithNamespace(context.Background(), namespaces.Default)
 	r, err := launcher.NewRunner(ctx, containerdClient, token, launchSpec, mdsClient, tpm, logger, serialConsole)
 	if err != nil {
 		return err
 	}
 	defer r.Close(ctx)
 
+	// Start tracking time for workload execution.
+	start = time.Now()
```
jkl73, line 215: Technically this is not where the workload actually starts (in `r.Run()`, it will still spend time measuring eventlog, creating attestation etc... before start the workload task), but the measurement is still useful here

launcher/launcher/main.go
jessieqliu: main() only takes a few seconds to run (at least in my test machines), I thought the time from VM start would be more useful but if not it can be switched back. wdyt?

launcher/launcher/main.go
jessieqliu: Got it, is it worth adding another timer when the workload actually starts and/or relabeling this to be more accurate?

launcher/launcher/main.go
alexmwu: I think both are useful to track how long boot takes and how long the launch itself takes

launcher/agent/agent.go
```diff
@@ -69,7 +69,7 @@ type agent struct {
 // - principalFetcher is a func to fetch GCE principal tokens for a given audience.
 // - signaturesFetcher is a func to fetch container image signatures associated with the running workload.
 // - logger will log any partial errors returned by VerifyAttestation.
-func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher, verifierClient verifier.Client, principalFetcher principalIDTokenFetcher, sigsFetcher signaturediscovery.Fetcher, launchSpec spec.LaunchSpec, logger *log.Logger) (AttestationAgent, error) {
+func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher, verifierClient verifier.Client, principalFetcher principalIDTokenFetcher, sigsFetcher signaturediscovery.Fetcher, launchSpec spec.LaunchSpec, logger logging.Logger) (AttestationAgent, error) {
```
alexmwu, line 72: Here and elsewhere please note the breaking change.

launcher/agent/agent.go
```diff
@@ -193,10 +193,10 @@ func fetchContainerImageSignatures(ctx context.Context, fetcher signaturediscove
 				},
 				retry(),
 				func(err error, _ time.Duration) {
-					logger.Printf("Failed to fetch container image signatures from repo %q: %v", targetRepo, err)
+					logger.Error("Failed to fetch container image signatures from repo: "+err.Error(), "repo", targetRepo)
```
alexmwu, line 196: should this use Sprintf?

launcher/container_runner.go
```diff
@@ -99,22 +99,24 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
...
+		"operator_override_cmd", launchSpec.Cmd,
+	)
 
 	imageConfig, err := getImageConfig(ctx, image)
 	if err != nil {
 		return nil, err
 	}
 
-	logger.Printf("Exposed Ports:             : %v\n", imageConfig.ExposedPorts)
+	logger.Info(fmt.Sprintf("Exposed Ports:             : %v\n", imageConfig.ExposedPorts))
```
alexmwu, line 114: nit: can we pretty print programmatically vs with hardcoded spaces? maybe a function that calculates the longest string and pads every other one?

alexmwu: I realize this isn't the point of your PR so it's okay if we don't do this.

launcher/internal/logging/logging.go
```diff
@@ -0,0 +1,255 @@
...
+	pl := payload{}
+	addArgs(pl, args)
+
+	if len(msg) > 0 {
+		pl[payloadMessageKey] = msg
+	}
+
+	if len(l.instanceName) > 0 {
+		// Needed for backwards compatibility with Cloudbuild tests.
+		pl[payloadHostnameKey] = l.instanceName
```
alexmwu, line 171: You can feel free to change the tests. That was just the quickest way I had to write that script

launcher/internal/logging/logging.go
```diff
@@ -0,0 +1,255 @@
...
+		return nil, err
+	}
+
+	// Configure Serial Console logger.
+	serialConsole, err := os.OpenFile(serialConsoleFile, os.O_WRONLY, 0)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open serial console for writing: %v", err)
+	}
+
+	slg := slog.New(slog.NewTextHandler(serialConsole, nil))
```
alexmwu, line 90: Can you show an example log? I assume it doesn't break our cloud bulid tests, but let's make sure they also don't break our e2e tests.

alexmwu (commented):
The general logic LGTM. I'd love to see an example structured log in the description and also cloud build tests

launcher/container_runner.go
jessieqliu: In the interest of getting this PR submitted soon (syncing to main is a huge hassle given how many files this changes) I'll add formatting changes - including a better attestation token log - in a different PR.

launcher/launcher/main.go
jessieqliu: Looking more into the code, I think we can either:

1. Keep the logs as is (so we're including some setup steps in the "workload duration"), which is slightly inaccurate but simple
2. Include the setup steps in `r.Run()` (event log measurements, attestation) in the launcher duration log (so we log the launcher's duration in `r.Run()` instead of `startLauncher()` - biggest pain here is we'll need to provide the `main()` start time to the container runner (or the logging library) to calculate the duration, which is doable but a bit clumsy.
3. Separately time/log the duration for the `r.Run()` setup, and start timing the workload when it actually starts (after `task.Wait()` returns successfully). This adds yet another metric/field we have to filter for, so I'd only prefer to do this if it's worth monitoring as its own step ("workload setup"?).

I'm generally in favor of 1 or 3 for ease of implementation, but lmk what you think

launcher/internal/logging/logging.go
jessieqliu: The field is still worth logging for filtering purposes but changed the key.

launcher/internal/logging/logging.go
jessieqliu: Added screenshots for what each type of log will look like in the PR description.

launcher/internal/logging/logging.go
jessieqliu: Scratch that - I haven't been able to get workload log redirection to Cloud Logging working. Going to make that future work and leave the keys as-is for test backcompat.

launcher/launcher/main.go
jkl73: I think 3 makes sense. The setup itself is also worth measuring.

launcher/internal/logging/logging.go
```diff
@@ -0,0 +1,255 @@
...
+	slg.Info("Serial Console logger initialized")
+
+	// This is necessary for DEBUG logs to propagate properly.
+	slog.SetDefault(slg)
```
jkl73, line 94: Seems like if called NewLogger(), then we cannot use SimpleLogger? (because the default will be set to a serial log)

launcher/internal/logging/logging.go
```diff
@@ -0,0 +1,255 @@
...
+
+	if len(l.instanceName) > 0 {
+		// Needed for backwards compatibility with Cloudbuild tests.
+		pl[payloadInstanceNameKey] = l.instanceName
+	}
+
+	logEntry.Payload = pl
+
+	l.cloudLogger.Log(logEntry)
+	l.cloudLogger.Flush()
```
jkl73, line 177: Seems like Flush() may return error, we may wan to Log the error serial

launcher/internal/logging/logging.go
```diff
@@ -0,0 +1,255 @@
...
+	// Write to serial console.
+	switch severity {
+	case clogging.Info, clogging.Notice:
+		l.serialLogger.Info(msg, args...)
+	case clogging.Warning:
+		l.serialLogger.Warn(msg, args...)
+	case clogging.Error, clogging.Critical, clogging.Alert, clogging.Emergency:
+		l.serialLogger.Error(msg, args...)
+	default:
+		slog.Debug(msg, args...)
```
jkl73, line 188: Can we include Debug to Info?

jkl73 (approved):
Mostly nits, feel free to submit

launcher/internal/logging/logging.go
jessieqliu: That's true, but SimpleLogger mostly exists for testing reasons and we don't anticipate it being used anywhere in tandem with NewLogger.

PR #475, opened 2024-08-14
Release lock if generating attestation returns error

yawangwang:
Release lock if tpm attestation returns error, which will avoid potential deadlock.


jkl73 (approved):
good catch..

PR #480, opened 2024-08-23
Remove curve P224 support for creating ECC seed

yawangwang:
Per discussions from https://github.com/google/go-tpm-tools/pull/471#discussion_r1725944593, remove curve P224 as it is not supported by crypto/ecdh


server/import.go
```diff
@@ -131,25 +132,45 @@ func createECCSeed(ek tpm2.Public) (seed, encryptedSeed []byte, err error) {
 	if err != nil {
 		return nil, nil, err
 	}
-	priv, x, y, err := elliptic.GenerateKey(curve, rand.Reader)
+
+	ecdsaPriv, err := ecdsa.GenerateKey(curve, rand.Reader)
```
alexmwu, line 136: We're doing ecdh here so you should use the utilities in crypto/ecdh instead

alexmwu (commented):
I am not an expert here, so I'm not sure when someone would want to pick a given curve. However, we should not remove support for the deprecated library given the GCP vTPM supports P224 unless there's a good reason besides the deprecated library.

```
sudo tpm2_getcap ecc-curves
TPM2_ECC_NIST_P256: 0x3
TPM2_ECC_NIST_P521: 0x5
TPM2_ECC_NIST_P384: 0x4
TPM2_ECC_NIST_P224: 0x2
```

server/import.go
yawangwang: We need EC points {x, y} to compute ECC seed in the following operations. Using `crypto/ecdh` utilities is an ideal way, however they don't provide a method to expose the EC points {x, y} in public key. 

crypto/ecdsa provides a method to do that, also there are methods which convert each key type to their counterparts in the the ecdh package. See line 141 that converts an ecdsa private key to its corresponding ecdh private key.

yawangwang:
> I am not an expert here, so I'm not sure when someone would want to pick a given curve. However, we should not remove support for the deprecated library given the GCP vTPM supports P224 unless there's a good reason besides the deprecated library.
> 
> ```
> sudo tpm2_getcap ecc-curves
> TPM2_ECC_NIST_P256: 0x3
> TPM2_ECC_NIST_P521: 0x5
> TPM2_ECC_NIST_P384: 0x4
> TPM2_ECC_NIST_P224: 0x2
> ```

I agree to keep the deprecated library till P224 is supported by crypto/ecdh. There is a proposal https://github.com/golang/go/issues/59783 that seeks for this support. So I'll leave this PR open for now.

server/import.go
alexmwu: Makes sense. LGTM

PR #481, opened 2024-08-26
Log detailed errors if refreshing SA credential goes wrong

yawangwang:
We have flaky long lived because container launcher failed to fetch container signatures with 403 errors.

Looks like the detailed error message is swallowed by [pullTargetImage](https://github.com/google/go-tpm-tools/blob/5816a17931c753f5efbf5057ef74211c199af285/launcher/internal/signaturediscovery/client.go#L89) method where refreshing SA credentials first returns an error and then failover to fetching a public docker repo.

We should log the error message accordingly to determine what exactly went wrong.

### DIFF
--- a/launcher/internal/signaturediscovery/client.go
+++ b/launcher/internal/signaturediscovery/client.go
@@ -28,23 +28,23 @@ type Fetcher interface {
 
 // Client is a wrapper of containerd.Client to interact with signed image manifest.
 type Client struct {
-	OriginalImageDesc     v1.Descriptor
-	remoteResolverFetcher remoteResolverFetcher
-	imageFetcher          imageFetcher
+	OriginalImageDesc v1.Descriptor
+	refreshResolver   remoteResolverFetcher
+	imageFetcher      imageFetcher
 }
 
 // New creates a new client that implements Fetcher interface.
 func New(originalImageDesc v1.Descriptor, resolverFetcher remoteResolverFetcher, imageFetcher imageFetcher) Fetcher {
 	return &Client{
-		OriginalImageDesc:     originalImageDesc,
-		remoteResolverFetcher: resolverFetcher,
-		imageFetcher:          imageFetcher,
+		OriginalImageDesc: originalImageDesc,
+		refreshResolver:   resolverFetcher,
+		imageFetcher:      imageFetcher,
 	}
 }
 
 // FetchSignedImageManifest fetches a signed image manifest using a tag-based discovery mechanism.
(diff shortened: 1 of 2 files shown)


launcher/internal/signaturediscovery/client.go
```diff
@@ -80,13 +80,17 @@ func (c *Client) FetchImageSignatures(ctx context.Context, targetRepository stri
 func (c *Client) pullTargetImage(ctx context.Context, targetRepository string) (containerd.Image, error) {
 	targetImageRef := fmt.Sprint(targetRepository, ":", formatSigTag(c.OriginalImageDesc))
 
+	// Pull signature image from a public repository.
+	if c.remoteResolverFetcher == nil {
+		return c.imageFetcher(ctx, targetImageRef)
+	}
+
 	// Refresh resolver before pulling container image.
-	if resolver, err := c.remoteResolverFetcher(ctx); err == nil {
```
JoshuaKrstic, line 84: as a small nit I'd like to rename "remoteResolverFetcher" to perhaps "refreshResolver"

It looks like in the way we use it there is only "refreshing" happening. It is a function after all, not an object. So I like having a verb in it.

Anyway, its a nit, but this was a bit hard to read.

If you're down, feel free to close these and open in another PR

launcher/internal/signaturediscovery/client.go
```diff
@@ -80,13 +80,17 @@ func (c *Client) FetchImageSignatures(ctx context.Context, targetRepository stri
 func (c *Client) pullTargetImage(ctx context.Context, targetRepository string) (containerd.Image, error) {
 	targetImageRef := fmt.Sprint(targetRepository, ":", formatSigTag(c.OriginalImageDesc))
 
+	// Pull signature image from a public repository.
+	if c.remoteResolverFetcher == nil {
+		return c.imageFetcher(ctx, targetImageRef)
```
JoshuaKrstic, line 85: same here, would you be open to calling this pullImage?
Or fetch image? Something to denote this is a func that is getting the image.

JoshuaKrstic (changes requested):

launcher/internal/signaturediscovery/client.go
yawangwang: Done.

launcher/internal/signaturediscovery/client.go
yawangwang: This method aims to fetch signature image based on the digest of the signed container, so `pullSignatureImage` seems more appropriate here.

JoshuaKrstic (approved):

PR #482, opened 2024-08-28
Move CEL encoding behind mutex lock

JoshuaKrstic:
The only other method behind this tpmMu is the MeasureEvent method. MeasureEvent updates the cel records array as the last step in its operation.

When getting the quote from the tpm, we first encode the CEL before locking + fetching data from the TPM.

I think this process should match MeasureEvent - lock THEN encode cel, do other TPM operations.
If not, please respond with why not for posterity.


JoshuaKrstic:
Test failures are from an expired cert in gce-tcb-verifier

alexmwu:
We only write to the CEL before we launch the container, so this is not necessary until we support multiple runs of the launcher/multiple workloads after the first launch.

@jkl73 does that sound right to you?

jkl73:
> We only write to the CEL before we launch the container, so this is not necessary until we support multiple runs of the launcher/multiple workloads after the first launch.
> 
> @jkl73 does that sound right to you?

Right, for now we don't need to worry about the CEL, as it's only measured once. But I think this might help in the future

jkl73:
Could you rebase from the main so the test can be fixed?

PR #483, opened 2024-09-03
Removed experiment flags that we would no longer consider rolling back

JoshuaKrstic:
Removing a number of experiments that we would not consider rolling back.

### DIFF
--- a/launcher/internal/experiments/experiments.go
+++ b/launcher/internal/experiments/experiments.go
@@ -12,9 +12,6 @@ import (
 // to treat experiment flags as their default value. The error should still be checked.
 type Experiments struct {
 	EnableTestFeatureForImage  bool
-	EnableSignedContainerImage bool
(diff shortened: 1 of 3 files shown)


launcher/internal/experiments/experiments.go
```diff
@@ -12,9 +12,6 @@ import (
 // to treat experiment flags as their default value. The error should still be checked.
 type Experiments struct {
 	EnableTestFeatureForImage  bool
-	EnableSignedContainerImage bool
-	EnableOnDemandAttestation  bool
-	EnableMemoryMonitoring     bool
 	EnableSignedContainerCache bool
 	EnableMeasureMemoryMonitor bool
```
yawangwang, line 16: `EnableSignedContainerCache` + `EnableMeasureMemoryMonitor` have been in prod for a while, i think we can possibly remove them as well.

jkl73 (approved):

PR #484, opened 2024-09-06
[launcher] Add ipv6 support on ingress network

jkl73:
Refactor the port opening logic and add support for accepting IPv6 ingress traffic.

using `-m multiport` in `iptables` command, so only one command is needed if opening multiple ports.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


launcher/container_runner.go
```diff
@@ -609,34 +610,61 @@ func initImage(ctx context.Context, cdClient *containerd.Client, launchSpec spec
...
+		default:
+			return "", "", fmt.Errorf("received unsupported protocol: got %s, expected tcp or udp", protocol)
 		}
 	}
+	// sort to make the list deterministic, make the test easier
+	sort.Strings(tcpPorts)
+	sort.Strings(udpPorts)
 
-	return nil
+	return strings.Join(tcpPorts, ","), strings.Join(udpPorts, ","), nil
```
yawangwang, line 667: nit: Since you're using the named values to convey the meaning of return values, you can possibly just need 
```
tcpPortStr, udpPortStr = strings.Join(tcpPorts, ","), strings.Join(udpPorts, ",")
return
```

launcher/container_runner.go
```diff
@@ -609,34 +610,61 @@ func initImage(ctx context.Context, cdClient *containerd.Client, launchSpec spec
...
+
+	for _, command := range []string{"iptables", "ip6tables"} {
+		for protocol, ports := range portMap {
+			if ports == "" {
+				continue
+			}
+			cmd := exec.Command(command, "-A", "INPUT", "-p", protocol, "-m", "multiport", "--dport", ports, "-j", "ACCEPT")
+			out, err := cmd.CombinedOutput()
+			if err != nil {
+				return fmt.Errorf("failed to set firewall policy using %v, for %s %s: %v %s", cmd, ports, protocol, err, out)
```
yawangwang, line 632: Do we want to collect all errors and then return, or return on error?

launcher/container_runner.go
jkl73: updated

launcher/container_runner.go
jkl73: I think here one error is fine, all four commands will probably result in the same error if there is one.

yawangwang (approved):

launcher/container_runner_test.go
```diff
@@ -632,6 +632,77 @@ func TestMeasureCELEvents(t *testing.T) {
 	}
 }
 
+func TestGetPortsList(t *testing.T) {
```
alexmwu, line 635: add a failure test

launcher/container_runner.go
```diff
@@ -609,34 +610,61 @@ func initImage(ctx context.Context, cdClient *containerd.Client, launchSpec spec
...
+		case "tcp":
+			tcpPorts = append(tcpPorts, port)
+		case "udp":
+			udpPorts = append(udpPorts, port)
+		default:
+			return "", "", fmt.Errorf("received unsupported protocol: got %s, expected tcp or udp", protocol)
 		}
 	}
+	// sort to make the list deterministic, make the test easier
+	sort.Strings(tcpPorts)
```
alexmwu, line 663: this is a string sort, not a numeric sort. If we must sort that seems a bit odd. Why can't we return a list of strings, have `openPorts` join the strings, and have the test do the sorting?

launcher/container_runner.go
```diff
@@ -609,34 +610,61 @@ func initImage(ctx context.Context, cdClient *containerd.Client, launchSpec spec
...
+			if err != nil {
+				return fmt.Errorf("failed to set firewall policy using %v, for %s %s: %v %s", cmd, ports, protocol, err, out)
+			}
 		}
+	}
+	return nil
+}
 
+// getPortsList takes in a docker file style port mapping and output string of comma separated port numbers
+func getPortsList(protocolAndports map[string]struct{}) (tcpPortStr, udpPortStr string, err error) {
```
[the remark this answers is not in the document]
jkl73: done

launcher/container_runner.go
```diff
@@ -609,34 +610,61 @@ func initImage(ctx context.Context, cdClient *containerd.Client, launchSpec spec
...
+	if err != nil {
+		return err
+	}
+
+	for _, command := range []string{"iptables", "ip6tables"} {
+		for protocol, ports := range portMap {
+			if ports == "" {
+				continue
+			}
+			cmd := exec.Command(command, "-A", "INPUT", "-p", protocol, "-m", "multiport", "--dport", ports, "-j", "ACCEPT")
```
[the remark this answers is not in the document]
jkl73: Yes, this is my intention to do the refector.
But after some tests I realized multiport option only supports 15 ports at most in one command. So this won't work. I will create a new PR to do the simple fix.

jkl73:
So seems like "multiport" only support at most 15 ports in one command which is not going to work in our use cases.

Created a new pr to address the ipv6 problem (pr/487) first
The refactoring/improve performance will be handled later.

PR #487, opened 2024-09-09
[launcher] add IPv6 support when opening ports

jkl73:
Previous iptables operation only operates on IPv4 traffic. Adding the support to allow IPv6 traffic by using ip6tables.


JoshuaKrstic (approved):

jkl73:
The cloudbuild integration test is failing, because the org policy firewall in our test project blocks IPv6 traffic. I'm working on the exemption, will submit this after the test can pass.

PR #489, opened 2024-09-13
Add retry to container signature fetch in agent

JoshuaKrstic:
Adding a retry mechanism to fetch container signatures.

Given that we support "partial" errors from the GAV, I went with a fixed 300ms interval a max of 3 times. The goal of this retry is to handle transient errors.
Longer waits can be done, but in the interest of ODA tokens returning quickly, I decided to not allow this to wait more than a second for any 1 repo.

Happy to change that if needed.

### DIFF
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -14,7 +14,9 @@ import (
 	"log"
 	"net/http"
 	"sync"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/google/go-tpm-tools/cel"
 	"github.com/google/go-tpm-tools/client"
 	"github.com/google/go-tpm-tools/launcher/internal/signaturediscovery"
@@ -138,7 +140,7 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
 	if a.launchSpec.Experiments.EnableSignedContainerCache {
 		signatures = a.sigsCache.get()
 	} else {
-		signatures = fetchContainerImageSignatures(ctx, a.sigsFetcher, a.launchSpec.SignedImageRepos, a.logger)
+		signatures = fetchContainerImageSignatures(ctx, a.sigsFetcher, a.launchSpec.SignedImageRepos, defaultRetryPolicy(), a.logger)
 	}
 	if len(signatures) > 0 {
 		req.ContainerImageSignatures = signatures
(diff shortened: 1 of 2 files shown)


launcher/agent/agent_test.go
```diff
@@ -338,6 +344,170 @@ func TestFetchContainerImageSignatures(t *testing.T) {
...
+		numTimes: numTimes,
+	}
+}
+
+func (f *failingClient) FetchImageSignatures(_ context.Context, targetRepository string) ([]oci.Signature, error) {
+	attempt := f.numTimes[targetRepository]
+	r := f.results[targetRepository][intMin(attempt, len(f.results[targetRepository])-1)]
+	f.numTimes[targetRepository] = attempt + 1
+	return r.result, r.err
+}
```
yawangwang, line 373: nit: Looks like signatures are returned in a round robin fashion, could you add some comments to this method?

Alternatively, you might consider changing the implementation to something like w/o calling `intMin`
```
func (f *failingClient) FetchImageSignatures(_ context.Context, targetRepository string) ([]oci.Signature, error) {
	attempt := f.numTimes[targetRepository]
	r := f.results[targetRepository][attempt]
	f.numTimes[targetRepository] = (attempt + 1)  % len(f.results[targetRepository])
	return r.result, r.err
}
```

yawangwang (approved):

launcher/agent/agent_test.go
JoshuaKrstic: Yeah fair, that makes more sense. Made the change.

JoshuaKrstic: Only change I am going to make is not modulo, as its not exactly round robin. Design here is to return the last value forever. So basically I'm just taking the nicer setting of the value from your implementation :)

launcher/agent/agent_test.go
```diff
@@ -338,6 +344,170 @@ func TestFetchContainerImageSignatures(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tpm := test.GetTPM(t)
+			defer client.CheckedClose(t, tpm)
```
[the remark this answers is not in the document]
JoshuaKrstic: Removed thanks for catching this.

launcher/agent/agent_test.go
```diff
@@ -338,6 +344,170 @@ func TestFetchContainerImageSignatures(t *testing.T) {
...
+			repos := []string{}
+			wantSigs := []oci.Signature{}
+			for k, v := range tc.resultmap {
+				repos = append(repos, k)
+				for _, result := range v {
+					if result.err == nil {
+						wantSigs = append(wantSigs, result.result...)
+					}
+				}
+			}
```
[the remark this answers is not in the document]
JoshuaKrstic: I think you're right for table driven tests, will make the change.
I wish table driven tests were different :(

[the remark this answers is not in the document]
JoshuaKrstic: I will handle this in a follow up (maybe)
I think I prefer this style of not having to copy a bunch of strings in the input.

Since I want to move things around here anyway, lets talk about this in our 1 on 1.
You are almost certainly right because this is what the tests look like now, but lets chat.,

launcher/agent/agent.go
```diff
@@ -165,27 +167,40 @@ func (a *agent) attest(nonce []byte, cel []byte) (*pb.Attestation, error) {
...
 		go func(targetRepo string, index int) {
 			defer wg.Done()
-			sigs, err := fetcher.FetchImageSignatures(ctx, targetRepo)
+
+			// backoff independently per repo
+			var sigs []oci.Signature
+			err := backoff.RetryNotify(
+				func() error {
+					s, err := fetcher.FetchImageSignatures(ctx, targetRepo)
+					sigs = s
```
jkl73, line 191: nit: can omit this assign

jkl73 (approved):

launcher/agent/agent.go
JoshuaKrstic: This one is needed as its the only place we actually get the return value from FetchImageSignatures.

Verified by removing it and watching tests fail

PR #495, opened 2024-09-23
Remove the x86 build/test from the CI.

qinkunbao:
Remove the x86 build/test from the CI from the CI for the following reasons.

1. Our CVMs only support x86-64.
2. Some dependency libraries (go-tdx-guest, go-sev-guest) are only tested with x86-64.
3. go-tpm-tools releases only have x86-64.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,6 @@ jobs:
         go-version: [1.21.x]
         # TODO: Get this working on windows-latest
         os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


jkl73:
CVM is not the only platform this tool/lib can run.
Majority features on gotpm cli and libs are not depending on go-tdx-guest/go-sev-guest
Users may still want to build the binary on a 32 bit system.

jkl73:
So I think we probably should keep x86 build/test in CI, unless it's so broken and beyond fix.

PR #496, opened 2024-09-25
Instantiate backoff strategy per goroutine

yawangwang:
Instantiate a new backoff object per goroutine since the current retry library is not thread-safe.

This PR also adds those tests related to `RefreshToken` + `FetchContainerSignatures` to a test suite with race condition check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,8 +77,8 @@ jobs:
         run: |
           GO_EXECUTABLE_PATH=$(which go)
           sudo $GO_EXECUTABLE_PATH test -v -run "TestFetchImageSignaturesDockerPublic" ./launcher
-      - name: Run specific tests to capture potential data race
-        run: go test ./launcher/agent -race -run TestCacheConcurrentSetGet
+      - name: Run all tests in launcher to capture potential data race
+        run: go test -v -race ./launcher/...
         if: (runner.os == 'Linux' || runner.os == 'macOS') && matrix.architecture == 'x64'
       - name: Test all modules
         run: go test -v ./... ./cmd/... ./launcher/... ./verifier/... -skip='TestCacheConcurrentSetGet|TestHwAttestationPass|TestHardwareAttestationPass'
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -407,13 +407,13 @@ func (r *ContainerRunner) refreshToken(ctx context.Context) (time.Duration, erro
 
 // ctx must be a cancellable context.
 func (r *ContainerRunner) fetchAndWriteToken(ctx context.Context) error {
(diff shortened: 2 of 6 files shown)


jkl73 (approved):

launcher/container_runner.go
```diff
@@ -472,10 +472,10 @@ defaultRetryPolicy retries as follows:
 Given the following arguments, the retry sequence will be:
 
 	RetryInterval = 60 sec
-	RandomizationFactor = 0.5
+	JitterPercentage = 50
 	Multiplier = 2
-	MaxInterval = 3600 sec
-	MaxElapsedTime = 0 (never stops retrying)
+	WithCappedDuration = 3600 sec
+	MaxRetries = 8
```
jkl73, line 478: Should we keep the unlimited retry?

yawangwang: my concern is the launcher will hang forever with unlimited retry if `refreshToken` keeps returning error. We might want the launcher exit in this case.

yawangwang: nvm, this PR will use the original library, so we still keep the unlimited retry.

launcher/container_runner.go
```diff
@@ -439,7 +439,7 @@ func (r *ContainerRunner) fetchAndWriteTokenWithRetry(ctx context.Context,
 						duration, err = r.refreshToken(ctx)
 						return err
 					},
-					retry,
+					newExponentialBackOff(retry),
```
JoshuaKrstic, line 442: Why not just make a new one like we did in agent.go?

JoshuaKrstic (changes requested):

launcher/container_runner.go
JoshuaKrstic: lets do this as a fast follow up so we can release

JoshuaKrstic (approved):

.github/workflows/ci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           GO_EXECUTABLE_PATH=$(which go)
           sudo $GO_EXECUTABLE_PATH test -v -run "TestFetchImageSignaturesDockerPublic" ./launcher
       - name: Run specific tests to capture potential data race
-        run: go test ./launcher/agent -race -run TestCacheConcurrentSetGet
+        run: go test -v ./launcher/... -race -run 'TestCacheConcurrentSetGet|^.*TokenRefresh.*$|^TestFetchContainerImageSignatures'
```
alexmwu, line 81: Given the recent concurrency issues, we should just run everything in launcher. The perf hit is only for testing/CI and it's orders of magnitudes faster than the image build/test anyway

https://go.dev/doc/articles/race_detector#Runtime_Overheads

alexmwu (approved):

.github/workflows/ci.yml
yawangwang: Done.

PR #497, opened 2024-09-25
Add GPU driver installation support in launcher

meetrajvala:
This PR adds the following changes for supporting the GPU driver installation in launcher component:

- Installation of open sourced GPU drivers using the `cos-gpu-installer`. 
- Changes in `container_runnner` to ensure that GPU device is accessible to the workload container.
- Utility functions for listing the GPU device files and remounting the installation directory.
- Image tests with required setup and validation scripts.
- Unit tests


launcher/container_runner.go
```diff
@@ -159,6 +160,20 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 		specOpts = append(specOpts, oci.WithDevShmSize(launchSpec.DevShmSize))
 	}
 
+	if launchSpec.Experiments.EnableGpuDriverInstallation && launchSpec.InstallGpuDriver {
+		appendGpuDriverMounts(mounts)
+		specOpts = append(specOpts, oci.WithMounts(mounts))
+
+		gpuDeviceFiles, err := listFilesWithPrefix("/dev", "nvidia")
```
jkl73, line 167: consider using https://pkg.go.dev/path/filepath#WalkDir to get all nvidia devices, may make the code more concise

launcher/launcher/main.go
```diff
@@ -155,6 +157,28 @@ func startLauncher(ctx context.Context, launchSpec spec.LaunchSpec, serialConsol
...
 	defer containerdClient.Close()
 
+	if launchSpec.Experiments.EnableGpuDriverInstallation && launchSpec.InstallGpuDriver {
+		gpuType, err := deviceinfo.GetGPUTypeInfo()
+		if err != nil {
+			return fmt.Errorf("failed to get the gpu type info: %v", err)
+		}
+
+		if !gpuType.OpenSupported() {
+			return fmt.Errorf("open-sourced gpu kernel modules are not supported for the gpu type %s", gpuType.String())
```
jkl73, line 167: maybe list all supported GPUs in the error mesg

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,122 @@
...
+	status := <-statusC
+	code, _, _ := status.Result()
+	di.logger.Printf("Gpu driver installation task exited with status: %d\n", code)
+	return nil
+}
+
+func getInstallerImageReference() (string, error) {
+	installerImageRefBytes, err := exec.Command("cos-extensions", "list", "--", "--gpu-installer").Output()
+	installerImageRef := strings.TrimSpace(string(installerImageRefBytes))
+	if err != nil {
```
jkl73, line 118: move the error check up, also doesn't make sense to have installerImageRef in the error message

launcher/image/test/scripts/test_gpu_workload.sh
```diff
@@ -0,0 +1,17 @@
+ #!/bin/bash
+set -euo pipefail
+source util/read_serial.sh
+
+# This test requires the workload to run and printing
+# corresponding messages to the serial console.
+SERIAL_OUTPUT=$(read_serial $1 $2) 
+print_serial=false
+
+if echo $SERIAL_OUTPUT | grep -q 'TEST PASSED'
```
jkl73, line 10: where is this "TEST PASSED" coming from, from the workload?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	container, err := di.cdClient.NewContainer(
+		ctx,
+		installerContainerID,
+		containerd.WithImage(image),
+		containerd.WithNewSnapshot(installerSnapshotID, image),
+		containerd.WithNewSpec(oci.WithImageConfig(image),
+			oci.WithPrivileged,
+			oci.WithProcessArgs("/cos-gpu-installer", "install", "-version=default", fmt.Sprintf("-host-dir=%s", InstallationHostDir)),
```
jkl73, line 83: Are you sure it's installing to the right directory? I saw on line 66, it's mounting the root to /root/ in the container

launcher/internal/experiments/experiments.go
```diff
@@ -11,10 +11,11 @@ import (
...
 type Experiments struct {
-	EnableTestFeatureForImage  bool
-	EnableSignedContainerCache bool
-	EnableMeasureMemoryMonitor bool
-	EnableTempFSMount          bool
+	EnableTestFeatureForImage   bool
+	EnableSignedContainerCache  bool
+	EnableMeasureMemoryMonitor  bool
+	EnableTempFSMount           bool
+	EnableGpuDriverInstallation bool
```
jkl73, line 18: Have you try to allowlist the test project on the experiment? otherwise the cloudbuild test might not pass

launcher/image/test/scripts/test_gpu_workload.sh
meetrajvala: yes, it is coming from workload.

launcher/internal/experiments/experiments.go
meetrajvala: yes, test project is in the allowlist.

launcher/internal/gpu/driverinstaller.go
meetrajvala: root is getting mounted to /root for installer container. we expect drivers to be installed under "/var/lib/nvidia" which is getting set by the host-dir flag where drivers will be installed on the host.

launcher/image/testworkloads/gpu/cuda-vector-add/Dockerfile
```diff
@@ -0,0 +1,7 @@
+# From current directory:
+# gcloud builds submit --tag us-west1-docker.pkg.dev/confidential-space-images-dev/cs-integ-test-images/gpu/cuda-vector-add:latest --project confidential-space-images-dev
+FROM gcr.io/google_containers/cuda-vector-add:v0.1
```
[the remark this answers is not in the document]
meetrajvala: With container registry to artifact registry migration, url will remain same. This existing url will continue to work after the migration.

launcher/launcher/main.go
meetrajvala: done

launcher/internal/gpu/driverinstaller.go
meetrajvala: done

launcher/internal/gpu/deviceinfo/deviceinfo.go
```diff
@@ -0,0 +1,99 @@
+// Package deviceinfo provides functionality to retrieve information about the GPU
+// attached to the host machine. It supports identifying GPU types and determining
+// whether they have support for open-source drivers.
+package deviceinfo
```
yawangwang, line 4: Why not just use cos.googlesource.com/cos/tools.git/src/cmd/cos_gpu_installer/deviceinfo? You don't have to duplicate the same logics here.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,123 @@
...
+	return &DriverInstaller{
+		cdClient:   cdClient,
+		launchSpec: launchSpec,
+		logger:     logger,
+	}
+}
+
+// InstallGPUDrivers installs the GPU driver on host machine using the cos-gpu-installer container.
+// It uses containerd package to launch and manage the driver installer container.
+func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
```
yawangwang, line 43: Can we just run the command using `exec.Command` from https://pkg.go.dev/cos.googlesource.com/cos/tools.git@v0.0.0-20241008015903-8431fe581b1f/src/cmd/cos_gpu_installer#section-readme to install GPU driver?

launcher/internal/gpu/deviceinfo/deviceinfo.go
meetrajvala: Done.

launcher/image/test/scripts/test_gpu_unsupported_gputype.sh
```diff
@@ -0,0 +1,17 @@
+ #!/bin/bash
+set -euo pipefail
+source util/read_serial.sh
+
+# This test requires the workload to run and printing
+# corresponding messages to the serial console.
+SERIAL_OUTPUT=$(read_serial $1 $2) 
+print_serial=false
+
+if echo $SERIAL_OUTPUT | grep -q 'unsupported gpu type'
```
yawangwang, line 10: could you refactor your script to accept another param to check if the expected message exists in serial logs? That way you don't need three scripts with same logics.

yawangwang:
Please squash your commits to have a cleaner history.

launcher/image/test/scripts/test_gpu_unsupported_gputype.sh
meetrajvala: along expected messages, there are different success and failure messages so I didn't parameterized it and kept it separate.

launcher/internal/gpu/driverinstaller.go
meetrajvala: Done. Added comment with README link.

launcher/util_test.go
```diff
@@ -79,3 +81,58 @@ func (t *testRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
...
+		},
+		{
+			dir:     "tmpdir",
+			pattern: "tmpfile",
+			want:    nil,
+			wantErr: true,
+		},
+	}
+
+	os.Mkdir("testdata", 0755)
```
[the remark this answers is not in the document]
meetrajvala: Done

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if err := task.Start(ctx); err != nil {
+		return fmt.Errorf("failed to start gpu driver installation task: %v", err)
+	}
+
+	status := <-statusC
+	code, _, _ := status.Result()
+	di.logger.Printf("Gpu driver installation task exited with status: %d\n", code)
+
+	err = remountAsExecutable(InstallationHostDir)
```
jkl73, line 134: This feels weird,

can you try one thing? try to create the directory before launching the install container? And see if the result driver is executable without remounting

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to remount the installed drivers: %v", err)
+	}
+	return nil
+}
+
+func getInstallerImageReference() (string, error) {
+	installerImageRefBytes, err := exec.Command("cos-extensions", "list", "--", "--gpu-installer").Output()
```
jkl73, line 142: can we hardcode this ref? I run the command I saw "us.gcr.io/cos-cloud/cos-gpu-installer:v2.4.1"
I assume this won't change if in the same image?

launcher/container_runner.go
```diff
@@ -159,6 +160,21 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 		specOpts = append(specOpts, oci.WithDevShmSize(launchSpec.DevShmSize))
 	}
 
+	if launchSpec.Experiments.EnableGpuDriverInstallation && launchSpec.InstallGpuDriver {
+		mounts = appendGpuDriverMounts(mounts)
```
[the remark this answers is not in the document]
meetrajvala: Yes, specOpts contains `mounts`. Updated it to directly append `gpuMounts`.

launcher/util.go
```diff
@@ -28,3 +31,25 @@ func FetchImpersonatedToken(ctx context.Context, serviceAccount string, audience
 
 	return []byte(token.AccessToken), nil
 }
+
+func listFilesWithPrefix(targetDir string, prefix string) ([]string, error) {
+	targetFiles := make([]string, 0)
+
+	err := filepath.WalkDir(targetDir, func(path string, _ os.DirEntry, err error) error {
```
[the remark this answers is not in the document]
meetrajvala: Done

launcher/internal/gpu/driverinstaller.go
meetrajvala: As we use image family (e.g. cos-113-lts`) field in cloudbuild, build will use the latest image from the family which may have the installer version updated and may break if we hardcode it.  So it is recommended to get the version using `cos-extensions` for given base cos image as `cos-gpu-installer` is not guaranteed to be backward compatible.

launcher/util.go
```diff
@@ -28,3 +31,20 @@ func FetchImpersonatedToken(ctx context.Context, serviceAccount string, audience
 
 	return []byte(token.AccessToken), nil
 }
+
+func listFilesWithPrefix(targetDir string, prefix string) ([]string, error) {
+	targetFiles := make([]string, 0)
```
yawangwang, line 36: nit: use nil initialization: `var targetFiles []string`

launcher/util.go
```diff
@@ -28,3 +31,20 @@ func FetchImpersonatedToken(ctx context.Context, serviceAccount string, audience
...
+	err := filepath.WalkDir(targetDir, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
```
yawangwang, line 40: The error handling logics should be carefully evaluated. From https://pkg.go.dev/io/fs#WalkDirFunc looks like simply returning a non-nil error will stop `WalkDir` walking the entire tree, is this the intended behavior?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,162 @@
...
+			oci.WithHostResolvconf,
+			oci.WithEnv([]string{fmt.Sprintf("HOSTNAME=%s", hostname)})))
+	if err != nil {
+		return fmt.Errorf("failed to create gpu driver installer container: %v", err)
+	}
+
+	task, err := container.NewTask(ctx, cio.NewCreator(cio.WithStdio))
+	if err != nil {
+		return fmt.Errorf("failed to create gpu driver installation task: %v", err)
+	}
```
yawangwang, line 124: cleanup task after its done.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,162 @@
...
+			Options:     []string{"rbind", "rw"},
+		},
+	}
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		di.logger.Printf("cannot get hostname: %v", err)
+	}
+
+	container, err := di.cdClient.NewContainer(
```
yawangwang, line 101: cleanup container when it's done.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to wait for gpu driver installation task: %v", err)
+	}
+
+	if err := task.Start(ctx); err != nil {
+		return fmt.Errorf("failed to start gpu driver installation task: %v", err)
+	}
+
+	status := <-statusC
+	code, _, _ := status.Result()
```
yawangwang, line 136: what if the returned code is not 0, do we just log the code and return nil? We should apply retry to non-zero code and return an error instead of nil.

launcher/launcher/main.go
```diff
@@ -155,6 +157,28 @@ func startLauncher(ctx context.Context, launchSpec spec.LaunchSpec, serialConsol
...
+		installer := gpu.NewDriverInstaller(containerdClient, launchSpec, logger)
+		err = installer.InstallGPUDrivers(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to install gpu drivers: %v", err)
+		}
+
+		err = remountAsExecutable(gpu.InstallationHostDir)
+		if err != nil {
+			return fmt.Errorf("failed to remount the installed drivers: %v", err)
+		}
```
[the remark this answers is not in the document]
yawangwang: sry but I'm still not seeing any reason blocking doing install + gpu consumption in the same place. You can possibly create a new container with a different ID from the existing workload container and do the installation from there.

launcher/image/test/scripts/test_gpu_unsupported_gputype.sh
yawangwang: then could you put those three separate files into a new directory something like `scripts/gpu` since they're highly related to gpu tests.

yawangwang (changes requested):

launcher/internal/gpu/driverinstaller.go
meetrajvala: done

launcher/internal/gpu/driverinstaller.go
meetrajvala: done

launcher/launcher/main.go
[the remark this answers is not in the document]
meetrajvala: gpu consumption related code snippet needs to be part of `NewRunner` function of `container_runnner` as the workload container gets created (not launched) as part of this constructor method. Calling the `InstallGpuDrivers` method from this `NewRunner` would add more responsibility to this method as this method should only create the runner object with required fields.

so `InstallGpuDrivers` function should be called before we call `NewRunner` method, so kept it as part of `startLauncher`.

launcher/util.go
meetrajvala: We will skip the directories for which `err` argument is set to non-nil during the `WalkDirFunc` call.

launcher/internal/gpu/driverinstaller.go
meetrajvala: Make sense to log and return an error. Returned error would not be retryable as we do not want to retry on the gpu driver installation failure.

launcher/util.go
```diff
@@ -28,3 +31,20 @@ func FetchImpersonatedToken(ctx context.Context, serviceAccount string, audience
 
 	return []byte(token.AccessToken), nil
 }
+
+func listFilesWithPrefix(targetDir string, prefix string) ([]string, error) {
+	var targetFiles []string
+	err := filepath.WalkDir(targetDir, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
```
yawangwang, line 38: I guess you also need to do a `d.IsDir` check, otherwise `WalkDir` will skip the path's parent directory, resulting in missing reading potential target files.

launcher/launcher/main.go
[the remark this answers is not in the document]
yawangwang: If you view gpu installation as a preliminary work for creating a runner object, then I'm okay with your approach here.

launcher/launcher/main.go
```diff
@@ -155,6 +156,14 @@ func startLauncher(ctx context.Context, launchSpec spec.LaunchSpec, serialConsol
 	}
 	defer containerdClient.Close()
 
+	if launchSpec.Experiments.EnableGpuDriverInstallation && launchSpec.InstallGpuDriver {
```
yawangwang, line 159: Imagine a scenario when experiment is off but `InstallGpuDriver` is on, do we want to continue the launcher execution, or log the error and then exit early? I'll leave it to you to make the call.

launcher/util.go
meetrajvala: Yes, Good catch.

launcher/launcher/main.go
meetrajvala: It would good to exit early and instead of waiting for the workload to fail.

launcher/cloudbuild.yaml
```diff
@@ -242,6 +242,37 @@ steps:
...
+  - 'PROJECT_ID=$PROJECT_ID'
+  script: |
+    #!/usr/bin/env bash
+
+    cd launcher/image/test
+    echo "running gpu driver installation image tests on ${OUTPUT_IMAGE_PREFIX}-hardened-${OUTPUT_IMAGE_SUFFIX}"
+    gcloud builds submit --config=test_gpu_driver_installation_cloudbuild.yaml --region us-west1 \
+      --substitutions _IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-hardened-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
+    exit
+# TODO: Enable these tests for debug image once gpu qouta is setup for the build project.
```
yawangwang, line 260: nit: create a bug item to track TODOs.

yawangwang (approved):
Please ensure to get approvals from other reviewers before merging this PR. Their expertise in CS image will also provide valuable insights.

launcher/internal/gpu/driverinstaller.go
meetrajvala: If we just create the directory with required permission before installation step, then it still fails. But if we run the same `remountAsExecutable` function before the installation steps, then it works fine.

jkl73 (approved):

launcher/image/test/create_gpu_vm.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
+local OPTIND
```
alexmwu, line 2: I don't understand why this needs to be a separate script rather than extending the existing util.

launcher/image/test/scripts/gpu/test_gpu_nogpu.sh
```diff
@@ -0,0 +1,17 @@
...
+print_serial=false
+
+if echo $SERIAL_OUTPUT | grep -q 'failed to get the gpu type info'
+then
+    echo "- no gpu verified"
+else
+    echo "FAILED: gpu not detected"
+    echo 'TEST FAILED.' > /workspace/status.txt
+    print_serial=true
+fi
```
alexmwu, line 17: nit: missing newline

launcher/launcher/main.go
```diff
@@ -155,6 +156,19 @@ func startLauncher(ctx context.Context, launchSpec spec.LaunchSpec, serialConsol
 	}
 	defer containerdClient.Close()
 
+	if launchSpec.InstallGpuDriver {
```
alexmwu, line 159: This shouldn't be done at the beginning of startLauncher. At minimum it should happen before the TPM check but probably after retrieving the auth token as well.

launcher/launcher/main.go
```diff
@@ -155,6 +156,19 @@ func startLauncher(ctx context.Context, launchSpec spec.LaunchSpec, serialConsol
...
+		if launchSpec.Experiments.EnableGpuDriverInstallation {
+			installer := gpu.NewDriverInstaller(containerdClient, launchSpec, logger)
+			err = installer.InstallGPUDrivers(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to install gpu drivers: %v", err)
+			}
+		} else {
+			logger.Println("Gpu installation experiment flag is not enabled for this project. Ensure that it is enabled when tee-install-gpu-driver is set to true")
```
alexmwu, line 167: Users can't do anything about this:

"GPU installation is not enabled for this project. Please contact support."

launcher/launcher/main.go
```diff
@@ -155,6 +156,19 @@ func startLauncher(ctx context.Context, launchSpec spec.LaunchSpec, serialConsol
...
+			return fmt.Errorf("gpu installation experiment flag is not enabled")
```
alexmwu, line 168: in comments here and everywhere else, GPU should be capitalized

launcher/util.go
```diff
@@ -28,3 +31,20 @@ func FetchImpersonatedToken(ctx context.Context, serviceAccount string, audience
...
 	return []byte(token.AccessToken), nil
 }
+
+func listFilesWithPrefix(targetDir string, prefix string) ([]string, error) {
+	var targetFiles []string
+	err := filepath.WalkDir(targetDir, func(path string, d os.DirEntry, err error) error {
+		if err != nil && d != nil && d.IsDir() {
+			return filepath.SkipDir
+		}
+		if d != nil && !d.IsDir() && strings.HasPrefix(filepath.Base(path), prefix) {
```
alexmwu, line 41: In what circumstances can err be non-nil and this path still be valid? It worries me that we're skipping a non-nil err check

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// InstallGPUDrivers installs the GPU driver on host machine using the cos-gpu-installer container.
+// This function performs the same steps specified in this README file:
+// https://pkg.go.dev/cos.googlesource.com/cos/tools.git@v0.0.0-20241008015903-8431fe581b1f/src/cmd/cos_gpu_installer#section-readme
+// README specifies docker command where this function uses containerd for launching and managing the gpu driver installer container.
+func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
+	err := remountAsExecutable(InstallationHostDir)
+	if err != nil {
+		return fmt.Errorf("failed to remount the installation directory: %v", err)
```
alexmwu, line 57: nit: this additional wrapping string is likely unnecessary. You can just return err.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	ctx = namespaces.WithNamespace(ctx, namespaces.Default)
+	installerImageRef, err := getInstallerImageReference()
+	if err != nil {
+		di.logger.Printf("failed to get the installer container image reference: %v", err)
+		return err
+	}
+
+	di.logger.Printf("cos gpu installer version : %s", installerImageRef)
```
alexmwu, line 76: COS to be capitalized, same with GPU. For consistency, don't add a space before the colon:

"COS GPU installer version: %s", installerImageRef

launcher/internal/gpu/driverinstaller.go
alexmwu: This could be hardcoded by running this command at build time. That's because it will always be built on the same COS version. This should be fixed by GA; please add a TODO.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,169 @@
...
+	container, err := di.cdClient.NewContainer(
+		ctx,
+		installerContainerID,
+		containerd.WithImage(image),
+		containerd.WithNewSnapshot(installerSnapshotID, image),
+		containerd.WithNewSpec(oci.WithImageConfig(image),
+			oci.WithPrivileged,
+			oci.WithProcessArgs("/cos-gpu-installer", "install", "-version=default", fmt.Sprintf("-host-dir=%s", InstallationHostDir)),
+			oci.WithAllDevicesAllowed,
+			oci.WithHostDevices,
```
alexmwu, line 110: Do we need this if we mount /dev?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,169 @@
...
+			oci.WithMounts(mounts),
+			oci.WithHostNamespace(specs.NetworkNamespace),
+			oci.WithHostNamespace(specs.PIDNamespace),
+			oci.WithHostHostsFile,
+			oci.WithHostResolvconf,
```
alexmwu, line 115: since a lot of this is duplicated, it would be useful to have helper functions that help mount with hostnamespace, hosts file and resolve conf.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -0,0 +1,169 @@
...
+		containerd.WithImage(image),
+		containerd.WithNewSnapshot(installerSnapshotID, image),
+		containerd.WithNewSpec(oci.WithImageConfig(image),
+			oci.WithPrivileged,
+			oci.WithProcessArgs("/cos-gpu-installer", "install", "-version=default", fmt.Sprintf("-host-dir=%s", InstallationHostDir)),
+			oci.WithAllDevicesAllowed,
+			oci.WithHostDevices,
+			oci.WithMounts(mounts),
+			oci.WithHostNamespace(specs.NetworkNamespace),
+			oci.WithHostNamespace(specs.PIDNamespace),
```
alexmwu, line 113: Why do we need the same PID namespace?

alexmwu (commented):
Mostly minor comments.

launcher/internal/gpu/driverinstaller.go
meetrajvala: Done

launcher/internal/gpu/driverinstaller.go
meetrajvala: Right, this can be removed as we mount `/dev`.

launcher/cloudbuild.yaml
```diff
@@ -242,6 +242,37 @@ steps:
...
+#     #!/usr/bin/env bash
+
+#     cd launcher/image/test
+#     echo "running gpu driver installation image tests on ${OUTPUT_IMAGE_PREFIX}-hardened-${OUTPUT_IMAGE_SUFFIX}"
+#     gcloud builds submit --config=test_gpu_driver_installation_cloudbuild.yaml --region us-west1 \
+#       --substitutions _IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-hardened-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
+#     exit
+# TODO: Enable these tests for debug image once gpu qouta is setup for the build project.
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: GpuDriverInstallationDebugImageTests
```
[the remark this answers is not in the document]
meetrajvala: We have to update the non confidential GPU quota to address this. As hardened image is more important, we enabled GpuDriverInstallationHardenedImageTests. GpuDriverInstallationDebugImageTests will be enabled once GPU quota is updated.

launcher/image/test/create_gpu_vm.sh
meetrajvala: We can extend the existing script. making this change as part of same PR would require updating existing tests too and increases this PR size. So I am planning to address it in follow-up PR.

launcher/internal/gpu/driverinstaller.go
meetrajvala: We are trying to convert [this usage command](https://cos.googlesource.com/cos/tools/+/refs/heads/master/src/cmd/cos_gpu_installer#usage) of `cos_gpu_installer` to go code and I added it as per the usage command.

launcher/internal/gpu/driverinstaller.go
meetrajvala: Will address this in follow-up PR.

PR #505, opened 2024-10-11
[launcher] launcher can expose IPv6 ports as well

jkl73:
Previously, port opening only works on IPv4 traffic, now enabling it for IPv6 as well.


alexmwu (approved):
Why did we need to separate this out from #487?

jkl73:
> Why did we need to separate this out from #487?

#487 contains the new integration test which is currently failing due the firewall setting the project. I will rebase #487 so it will only contains the test and submit it once the firewall setting is fixed.

PR #507, opened 2024-10-16
Start NPD after LaunchSpec Verification

jessieqliu:
[PR 479](https://github.com/google/go-tpm-tools/pull/479) starts NPD before the spec is verified, meaning it only starts NPD based on the spec and not the policy. This PR moves that logic to after verification is properly done. The previous PR was still passing cloudbuild tests because `spec.Verify` was still emitting the right logs, but they weren't actually reflective of whether NPD was started.

The previous PR also replaced Memory Monitoring labels in cloud build tests with new ones. However, we should keep the previous tests to make sure we don't break the old labels until they're officially removed - tests for the new labels will be added separately.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -123,24 +123,8 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 		return nil, err
 	}
 
-	if launchSpec.MonitoringEnabled != spec.None {
-		logger.Printf("Health Monitoring is enabled by the VM operator")
-
-		if launchSpec.MonitoringEnabled == spec.All {
-			logger.Printf("All health monitoring metrics enabled")
-			if err := nodeproblemdetector.EnableAllConfig(); err != nil {
-				logger.Printf("Failed to enable full monitoring config: %v", err)
-				return nil, err
-			}
-		} else if launchSpec.MonitoringEnabled == spec.MemoryOnly {
-			logger.Printf("memory/bytes_used enabled")
-		}
-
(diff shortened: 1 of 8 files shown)


launcher/spec/launch_policy.go
```diff
@@ -218,7 +231,7 @@ func verifyMonitoringConfig(policy MonitoringType, spec MonitoringType) error {
 	case None:
 		// If policy is 'None', spec must also be 'None'.
 		if spec != None {
-			return fmt.Errorf("spec configured for %v but policy is none", spec)
+			return fmt.Errorf("spec configured for %v but policy is none", toString(spec))
```
yawangwang, line 234: nit: you can make MonitotringType as a receiver type and implement the method `func (mt MonitoringType) String() string` instead.  Then doing `return fmt.Errorf("spec configured for %s but policy is none", spec)` should be just fine.

launcher/container_runner.go
```diff
@@ -122,6 +123,26 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
...
+			logger.Printf("memory/bytes_used enabled")
+		}
+
+		if err := nodeproblemdetector.StartService(logger); err != nil {
+			logger.Print(err)
+		}
+	} else {
+		logger.Printf("Health Monitoring is disabled")
+	}
+
```
yawangwang, line 145: nit: consider extracting this chunk of logics to a new function as `NewRunner` is becoming too complicated.

yawangwang (approved):

PR #511, opened 2024-11-02
Apply retry logics in confidential computing API + workload image puller

yawangwang:
This PR serves the purpose of mitigating the 404 error `metadata entry test/token not found` from client prober tests.

After investigation, several factors contribute to this 404 error:
- 401 unauthorized error when pulling container image. see [cloud loggings](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22gce_instance%22%0Aresource.labels.instance_id%3D%222754575401503739344%22%0Alog_name%3D%22projects%2Fconfidentialcomputing-e2e%2Flogs%2Fconfidential-space-launcher%22;cursorTimestamp=2024-10-08T18:01:31.362267Z;startTime=2024-10-08T03:02:41.974Z;endTime=2024-10-14T18:02:41.974825Z?project=confidentialcomputing-e2e&e=13803378&mods=allow_workbench_image_override)
- 500 internal error when calling `v1.GetLocation`.
```
I1004 11:07:17.754233   13330 instance.go:634] [LogVMSerialConsole] 2024/10/04 18:05:54 failed to create REST verifier client: invalid region "southamerica-east1", available regions are [africa-south1, asia-east1, asia-east2, asia-northeast1, asia-northeast2, asia-northeast3, asia-south1, asia-south2, asia-southeast1, asia-southeast2, australia-southeast1, australia-southeast2, europe-central2, europe-north1, europe-southwest1, europe-west1, europe-west10, europe-west12, europe-west2, europe-west3, europe-west4, europe-west6, europe-west8, europe-west9, me-central1, me-central2, me-west1, northamerica-northeast1, northamerica-northeast2, southamerica-east1, southamerica-west1, us-central1, us-east1, us-east4, us-east5, us-east7, us-south1, us-west1, us-west2, us-west3, us-west4]: googleapi: Error 500: Internal error encountered.
```
- 500 internal error when calling `v1.VerifyAttestation`. see [cloud loggings]( https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22gce_instance%22%0Aresource.labels.instance_id%3D%229183949205174332022%22%0Alog_name%3D%22projects%2Fconfidentialcomputing-e2e%2Flogs%2Fconfidential-space-launcher%22;cursorTimestamp=2024-10-10T17:54:26.254584Z;startTime=2024-10-08T03:02:41.974Z;endTime=2024-10-14T18:02:41.974825Z?project=confidentialcomputing-e2e&e=13803378&mods=allow_workbench_image_override)
- 502 bad gateway error when pulling container image. see [cloud loggings](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22gce_instance%22%0Aresource.labels.instance_id%3D%223428719970864247249%22%0Alog_name%3D%22projects%2Fconfidentialcomputing-e2e%2Flogs%2Fconfidential-space-launcher%22;cursorTimestamp=2024-10-13T00:08:54.063503Z;startTime=2024-10-08T03:02:41.974Z;endTime=2024-10-14T18:02:41.974825Z?project=confidentialcomputing-e2e&e=13803378&mods=allow_workbench_image_override)

To mitigate the 404 test flakes, we should apply retry strategies to the following places:
- when the launcher pulls the workload image.
- when the rest client calls `GetLocation`.
- when the rest client calls `VerifyAttestation`.


launcher/container_runner.go
```diff
@@ -621,17 +631,39 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 	return nil
 }
 
+func pullImageWithRetries(f func() (containerd.Image, error), retry func() backoff.BackOff) (containerd.Image, error) {
```
JoshuaKrstic, line 634: This retry method seems very generic, with only the return in the func being passed being specific to pulling images.
Is it possible to make this more generic? Or less generic?

verifier/rest/retry.go
JoshuaKrstic, line 1: Have you considered passing non-default call options instead of having our own retries? 

https://github.com/googleapis/google-cloud-go/blob/confidentialcomputing/v1.7.2/confidentialcomputing/apiv1/confidential_computing_client.go#L69

JoshuaKrstic (changes requested):

verifier/rest/retry.go
JoshuaKrstic: Oh actually it looks like we aren't able to. Ill file a bug on the library generation team to ask for this, I don't see why we can't override this.

JoshuaKrstic: This is not overridable: https://github.com/googleapis/google-cloud-go/blob/confidentialcomputing/v1.7.2/confidentialcomputing/apiv1/confidential_computing_client.go#L232

verifier/rest/rest.go
```diff
@@ -89,7 +95,7 @@ func NewClient(ctx context.Context, projectID string, region string, opts ...opt
 }
 
 type restClient struct {
-	v1Client *v1.Client
+	v1Client *retryableClient
```
jkl73, line 98: Why can't we do the retry in the `restClient`'s functions?

The retry behavior is specific to the rest client, and then we don't need to create a wrapper around v1.Client anymore

JoshuaKrstic: +1 but even deeper, I would like to do the retries within the actual client provided by the API library generation. See my other comment.

Its currently unsupported tho, so we would need to open a feature request and re-gen the libraries, if they even agree to do the work at all,.

verifier/rest/retry.go
JoshuaKrstic: Looks like we can just manipulate the CallOptions directly in the returned client: https://yaqs.corp.google.com/eng/q/8957170376165556224#a2

We should probably just do that except for VerifyAttestation, as we discussed in the sync

launcher/container_runner.go
yawangwang: The intention is to make specific to pulling images. The reason I abstract this method is to make unit tests easier so that I can mock containerd API. Since cotainerd will make a network call to an actual docker registry, I want to avoid it in unit tests.

verifier/rest/retry.go
yawangwang: Done.
The default CallOptions is now overridden by our specific retry strategy. Also we retry the attestation flow when 500 errors originate from VerifyAttestation API.

verifier/rest/rest.go
yawangwang: Done. We manipulated the CallOptions directly in the returned client.

JoshuaKrstic (approved):
Looks great, its really nice that we are able to override callOptions like that.

launcher/agent/agent.go
```diff
@@ -101,10 +106,49 @@ func (a *agent) MeasureEvent(event cel.Content) error {
...
+
+			// Otherwise, save the error and exit the retry.
+			err = doErr
+			return nil
+		},
+		retry(),
+	)
+
+	// If retryErr is set, it means we retried maxAttempts for VerifyAttestation without success.
+	if retryErr != nil {
```
JoshuaKrstic, line 137: can't we just return the retryerr here? And combine this if with the next to say 
if retryErr != nil || err != nil {...} ?

launcher/agent/agent.go
```diff
@@ -101,10 +107,44 @@ func (a *agent) MeasureEvent(event cel.Content) error {
...
 }
 
-// Attest fetches the nonce and connection ID from the Attestation Service,
+// Attest is a thin wrapper of AttestWithRetries with defaultRetryPolicy.
+func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error) {
+	return a.AttestWithRetries(ctx, opts, defaultRetryPolicy)
+}
+
+// Attest executes doAttest with retries when 500 errors originate from VerifyAttestation API.
+func (a *agent) AttestWithRetries(ctx context.Context, opts AttestAgentOpts, retry func() backoff.BackOff) ([]byte, error) {
```
jkl73, line 116: I still feel we shouldn't do the retries in the `agent`. Instead this is the job for the caller, in our case, it should be handled in container_runner

verifier/rest/rest.go
```diff
@@ -6,40 +6,58 @@ import (
...
 )
 
-// BadRegionError indicates that:
-//   - the requested Region cannot be used with this API
-//   - other Regions _can_ be used with this API
-type BadRegionError struct {
-	RequestedRegion  string
-	AvailableRegions []string
-	err              error
+func confComputeBackoffPolicy() backoff.BackOff {
```
jkl73, line 28: Seems like this not not needed, it's not referenced anywhere

verifier/rest/rest.go
```diff
@@ -6,40 +6,58 @@ import (
...
+	Maximum interval = 1000 milliseconds
+	Maximum retries = 2
+*/
+func confComputeCallOptions() *v1.CallOptions {
+	callOption := []gax.CallOption{
+		gax.WithTimeout(1000 * time.Millisecond),
+		gax.WithRetry(func() gax.Retryer {
+			return gax.OnCodes([]codes.Code{
+				codes.Unavailable,
+				codes.Internal,
```
jkl73, line 47: Seems like your retry option only going to retry on these two codes. Do we need to worry about other codes (like in here: https://cloud.google.com/storage/docs/retry-strategy#go)

verifier/rest/rest.go
```diff
@@ -51,6 +69,9 @@ func NewClient(ctx context.Context, projectID string, region string, opts ...opt
 		return nil, fmt.Errorf("can't create ConfidentialComputing v1 API client: %w", err)
 	}
 
+	// Override the default retry CallOptions with specific retry policies.
```
jkl73, line 72: Is there a reference to the default retry CallOptions, why can't we just rely on the default retry options?

verifier/rest/rest.go
yawangwang: These codes are actually the mapping of gRPC error codes in https://github.com/grpc/grpc/blob/master/doc/statuscodes.md. And yes, these are only two error codes we want to retry on because 1. code.Unavailable is the default error code to retry 2. we only observed 500 internal error codes so far which caused test flakiness.

verifier/rest/rest.go
yawangwang: The default retry options can be found in here [google3/google/cloud/confidentialcomputing/confidentialcomputing_v1_grpc_service_config.json](https://source.corp.google.com/piper///depot/google3/google/cloud/confidentialcomputing/confidentialcomputing_v1_grpc_service_config.json) and only retry on the code "UNAVAILABLE".

verifier/rest/rest.go
yawangwang: Good catch, removed.

verifier/rest/rest.go
jkl73: cool, we may want to change the default json config later in the API, especially if we want to add more retry codes.

Add the OSS version here for reference:
https://github.com/googleapis/googleapis/blob/master/google/cloud/confidentialcomputing/v1/confidentialcomputing_v1_grpc_service_config.json

jkl73 (approved):

PR #515, opened 2024-12-14
Add AWS Principal Tag type to launcher

JoshuaKrstic:
This PR:

- Adds support for the new AWS Principal Tag token type with the updated API
- Adds support for the new TokenOptions field to modify the behavior of AWSPrincipalTag tokens
- Moves the definition of the TokenOptions to a shared models area within the internal package
- Cleans up the teeserver class a bit to make HTTP error handling + logging more consistent
- Cleans up the teeserver class to use HTTP constants instead of strings

AWSPrincipalTag tokens are guarded by the same server side experiment as limited AWS. We will allow the server to reject requests instead of guarding requests in the image.


JoshuaKrstic:
!gcbrun

client/pcr.go
```diff
@@ -26,7 +26,7 @@ const (
 // CertifyHashAlgTpm is the hard-coded algorithm used in certify PCRs.
 const CertifyHashAlgTpm = tpm2.AlgSHA256
 
-func min(a, b int) int {
+func minimum(a, b int) int {
```
JoshuaKrstic, line 29: these needed to change to make the Revive linter happy

internal/pcrs.go
```diff
@@ -16,13 +16,13 @@ import (
 const minPCRIndex = uint32(0)
 
 func maxPCRIndex(p *pb.PCRs) uint32 {
-	max := minPCRIndex
+	high := minPCRIndex
```
JoshuaKrstic, line 19: same here, these are for the revive linter

client/pcr.go
deeglaze: Min is a builtin golang function starting 1.21. Do you need to define this?

internal/cert_test.go
```diff
@@ -1,3 +1,4 @@
+// Package internal contains private helper functions and models needed in client and server
```
deeglaze, line 1: You only need the package comment once per package. You can put it in doc.go idiomatically if you have multiple files defining a package.

verifier/client.go
```diff
@@ -5,6 +5,7 @@ package verifier
 import (
 	"context"
 
+	models "github.com/google/go-tpm-tools/internal/models"
```
deeglaze, line 8: You don’t need to bind to “models” when the package basename is “models”.

deeglaze (commented):
Go readability comments.

verifier/client.go
JoshuaKrstic: You're right, I was importing something else first in an earlier revision. Fixed.

JoshuaKrstic: I thought you were right and made the change but the package is ofc "internal", so this wont work. WDYT, should I make a new package called "models" instead of co-opting this one?

client/pcr.go
JoshuaKrstic: Oh cool, no we don't. Ill remove it.

internal/cert_test.go
JoshuaKrstic: Fixed thank you

verifier/client.go
JoshuaKrstic: I made the package

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/google/go-tpm-tools
 
-go 1.20
+go 1.21
```
jkl73, line 3: Try not to upgrade the go version in this PR, should reduce the size of the PR

internal/models/token_options.go
```diff
@@ -0,0 +1,43 @@
+// Package models contains models needed in client and server
+package models
```
jkl73, line 2: This probably not belong to the /internal package, because it has nothing to do with TPM.
This might belong to the verifier or the launcher package

launcher/teeserver/tee_server_test.go
```diff
@@ -135,7 +137,14 @@ func TestCustomToken(t *testing.T) {
 			body: `{
 				"audience": "audience",
 				"nonces": ["thisIsAcustomNonce"],
-				"token_type": "OIDC"
+				"token_type": "OIDC",
+				"aws_principal_tag_options" : {
```
jkl73, line 141: Can you add a new test case for this? Instead of updating an existing test case

verifier/rest/rest_test.go
```diff
@@ -194,3 +196,222 @@ func TestConvertTDXProtoToREST(t *testing.T) {
...
+	for _, tc := range testCases {
+		pbTokenOpts := convertTokenOptionsToREST(tc.tokenOptions)
+		ok, result := compareTokenOptionsPbs(tc.wantpb, pbTokenOpts)
+		if !ok {
+			t.Errorf("%v: %s", tc.name, result)
+		}
+	}
+}
+
+func compareTokenOptionsPbs(want *ccpb.TokenOptions, got *ccpb.TokenOptions) (bool, string) {
```
jkl73, line 357: Can we do cmp.Equal on the structs, so we don't need to have a customized comparator.

verifier/rest/rest.go
```diff
@@ -142,22 +144,49 @@ func convertChallengeFromREST(chal *confidentialcomputingpb.Challenge) (*verifie
...
+			AwsPrincipalTagsOptions: &ccpb.TokenOptions_AwsPrincipalTagsOptions{}}
+		if request.PrincipalTagOptions.HasAllowedPrincipalTags() {
+			tokenOptionspb.GetAwsPrincipalTagsOptions().AllowedPrincipalTags = &ccpb.TokenOptions_AwsPrincipalTagsOptions_AllowedPrincipalTags{}
+			if request.PrincipalTagOptions.AllowedPrincipalTags.HasContainerImageSignatures() {
+				tokenOptionspb.GetAwsPrincipalTagsOptions().GetAllowedPrincipalTags().ContainerImageSignatures = &ccpb.TokenOptions_AwsPrincipalTagsOptions_AllowedPrincipalTags_ContainerImageSignatures{
+					KeyIds: request.PrincipalTagOptions.AllowedPrincipalTags.ContainerImageSignatures.KeyIDs,
+				}
+			}
+		}
+	}
```
jkl73, line 181: Can you move this to the switch statement above, seem like this only affects AWS_PRINCIPALTAGS  token

jkl73: Also this looks very complicated because of the long variable names, if you can simplify the logic it will be great

internal/models/token_options.go
JoshuaKrstic: moved to verifier

launcher/teeserver/tee_server_test.go
JoshuaKrstic: Done.

verifier/rest/rest_test.go
JoshuaKrstic: okay yeah my bad, just had to set a var to ignore the internal state of the objs being compared. Good call

verifier/rest/rest.go
JoshuaKrstic: Oops yeah this fell apart in a small refactor I did

JoshuaKrstic: Updated, LMK what you think

go.mod
JoshuaKrstic: Tried to do this but we need 1.21. The protobuf upgrade takes in new libraries that rely on 1.21 features: 

../../../../go/pkg/mod/go.opentelemetry.io/otel@v1.29.0/attribute/set.go:7:2: package cmp is not in GOROOT (/opt/homebrew/Cellar/go@1.20/1.20.14/libexec/src/cmp)
note: imported by a module that requires go 1.21
../../../../go/pkg/mod/google.golang.org/grpc@v1.67.1/experimental/stats/metricregistry.go:22:2: package maps is not in GOROOT (/opt/homebrew/Cellar/go@1.20/1.20.14/libexec/src/maps)
note: imported by a module that requires go 1.21
../../../../go/pkg/mod/google.golang.org/grpc@v1.67.1/clientconn.go:27:2: package slices is not in GOROOT (/opt/homebrew/Cellar/go@1.20/1.20.14/libexec/src/slices)
note: imported by a module that requires go 1.21

Gonna keep it as is

JoshuaKrstic: Reverted this

launcher/teeserver/tee_server.go
```diff
@@ -156,3 +143,17 @@ func (s *TeeServer) Shutdown(ctx context.Context) error {
 	}
 	return nil
 }
+
+func validateTokenOptions(opts models.TokenOptions) error {
```
jkl73, line 147: This validate function only applies to the "POST" request,  I think it's better to keep it where it was, otherwise we should rename the function name

launcher/teeserver/tee_server_test.go
```diff
@@ -170,3 +191,139 @@ func TestCustomToken(t *testing.T) {
...
+		{
+			testName: "TestKeyIdsReadSuccessfullyEvenWithInvalidTokenTypeMatch",
+			body: `{
+				"audience": "audience",
+				"nonces": ["thisIsAcustomNonce"],
+				"token_type": "OIDC",
+				"aws_principal_tag_options" : {
+					"allowed_principal_tags": {
+						"container_image_signatures" : {
+							"key_ids": ["test1", "test2"]
```
jkl73, line 212: I think there is limit on how many key_ids? Is it enforced on the service side?

verifier/rest/rest.go
```diff
@@ -13,10 +13,12 @@ import (
 	tabi "github.com/google/go-tdx-guest/abi"
 	"github.com/google/go-tdx-guest/proto/tdx"
 	"github.com/google/go-tpm-tools/verifier"
+	"github.com/google/go-tpm-tools/verifier/models"
 	"github.com/google/go-tpm-tools/verifier/oci"
 
 	v1 "cloud.google.com/go/confidentialcomputing/apiv1"
-	confidentialcomputingpb "cloud.google.com/go/confidentialcomputing/apiv1/confidentialcomputingpb"
+	"cloud.google.com/go/confidentialcomputing/apiv1/confidentialcomputingpb"
+	ccpb "cloud.google.com/go/confidentialcomputing/apiv1/confidentialcomputingpb"
```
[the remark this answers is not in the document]
JoshuaKrstic: weird that it stayed, fixed.

launcher/teeserver/tee_server.go
```diff
@@ -77,65 +72,57 @@ func (a *attestHandler) getToken(w http.ResponseWriter, r *http.Request) {
...
-	case "GET":
+	case http.MethodGet:
 		// this could call Attest(ctx) directly later.
 		data, err := os.ReadFile(a.defaultTokenFile)
 
 		if err != nil {
-			a.logger.Error(err.Error())
-			w.WriteHeader(http.StatusNotFound)
-			w.Write([]byte("failed to get the token"))
+			err = fmt.Errorf("failed to get the token")
```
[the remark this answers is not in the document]
JoshuaKrstic: fixed

verifier/rest/rest.go
```diff
@@ -142,22 +144,37 @@ func convertChallengeFromREST(chal *confidentialcomputingpb.Challenge) (*verifie
 	}, nil
 }
 
-func convertRequestToREST(request verifier.VerifyAttestationRequest) *confidentialcomputingpb.VerifyAttestationRequest {
-	idTokens := make([]string, len(request.GcpCredentials))
-	for i, token := range request.GcpCredentials {
-		idTokens[i] = string(token)
+func convertTokenOptionsToREST(request *models.TokenOptions) *ccpb.TokenOptions {
```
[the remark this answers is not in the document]
JoshuaKrstic: renamed to tokenOpts

launcher/teeserver/tee_server.go
JoshuaKrstic: Fair enough, moved it back.

verifier/models/token_options.go
```diff
@@ -0,0 +1,43 @@
...
+
+// HasAllowedPrincipalTags returns true if AllowedPrincipalTags is not nil.
+func (a AWSPrincipalTagsOptions) HasAllowedPrincipalTags() bool {
+	return a.AllowedPrincipalTags != nil
+}
+
+// HasContainerImageSignatures returns true if ContainerImageSignatures is not nil.
+func (a AllowedPrincipalTags) HasContainerImageSignatures() bool {
+	return a.ContainerImageSignatures != nil
+}
```
jkl73, line 43: Do we really need these three functions? can we just do the nil check directly

verifier/rest/rest.go
```diff
@@ -271,34 +283,56 @@ func convertOCISignatureToREST(signature oci.Signature) (*confidentialcomputingp
...
-	return &confidentialcomputingpb.VerifyAttestationRequest_TdCcel{
-		TdCcel: &confidentialcomputingpb.TdxCcelAttestation{
+	return &ccpb.VerifyAttestationRequest_TdCcel{
+		TdCcel: &ccpb.TdxCcelAttestation{
 			TdQuote: rawQuote,
 		},
 	}, nil
 }
+
+func setAwsPrincipalTagOptions(request *models.TokenOptions, optsPb *ccpb.TokenOptions) {
```
jkl73, line 318: `request` can get rename to `requestTokenOptions` or something similar

jkl73 (commented):
mostly nits

launcher/teeserver/tee_server.go
```diff
@@ -77,65 +72,65 @@ func (a *attestHandler) getToken(w http.ResponseWriter, r *http.Request) {
...
 			return
 		}
 
-		tok, err := a.attestAgent.Attest(a.ctx,
-			agent.AttestAgentOpts{
-				Aud:       tokenReq.Audience,
-				Nonces:    tokenReq.Nonces,
-				TokenType: tokenReq.TokenType,
-			})
+		// Do not check that TokenTypeOptions matches TokenType in the launcher.
```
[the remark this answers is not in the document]
JoshuaKrstic: Yes that's exactly right

launcher/teeserver/tee_server.go
```diff
@@ -77,65 +72,65 @@ func (a *attestHandler) getToken(w http.ResponseWriter, r *http.Request) {
...
 		decoder := json.NewDecoder(r.Body)
 		decoder.DisallowUnknownFields()
 
-		err := decoder.Decode(&tokenReq)
+		err := decoder.Decode(&tokenOptions)
 		if err != nil {
-			a.logger.Error(err.Error())
-			w.WriteHeader(http.StatusBadRequest)
-			w.Write([]byte(err.Error()))
+			a.logAndWriteHTTPError(w, http.StatusBadRequest, err)
```
alexmwu, line 94: You should prefix this error with some more context like `fmt.Errorf("failed to parse POST body as TokenOptions: %v", err)`. Otherwise the resulting error is not very understandable.

verifier/models/token_options.go
JoshuaKrstic: After removing that test method you are right. removed this

launcher/teeserver/tee_server.go
JoshuaKrstic: Good catch

verifier/rest/rest.go
```diff
@@ -258,34 +270,56 @@ func convertOCISignatureToREST(signature oci.Signature) (*confidentialcomputingp
...
-	return &confidentialcomputingpb.VerifyAttestationRequest_TdCcel{
-		TdCcel: &confidentialcomputingpb.TdxCcelAttestation{
+	return &ccpb.VerifyAttestationRequest_TdCcel{
+		TdCcel: &ccpb.TdxCcelAttestation{
 			TdQuote: rawQuote,
 		},
 	}, nil
 }
+
+func setAwsPrincipalTagOptions(requestTokenOptions *models.TokenOptions, optsPb *ccpb.TokenOptions) {
```
yawangwang, line 305: Readability: this method seems to convert `*models.TokenOptions` to `*ccpb.TokenOptions` and then set, it would be clearer to rename the method to `awsPrincipalTagOptions` and have `*ccpb.TokenOptions` as a return value. That way the caller will be responsible to set TokenOptions field.

yawangwang (approved):

go.mod
jkl73: Seems like go mod tidy think it needs to upgrade to go1.21 (I don't know why), I think it's okay to upgrade the go version then...

jkl73 (approved):
Make sure cloudbuild can finish before merging

verifier/rest/rest.go
JoshuaKrstic: yeah good call, made the change

go.mod
JoshuaKrstic: I also do not know why :(

PR #517, opened 2024-12-21
[launcher] Allow multiple Roots of Trust in Agent

jkl73:
Launcher Attestation Agent now can support multiple Root of Trust.
    Extend operation will extend measurements to all RoTs.
    Attest will generate quote/report using a technology specific
    RoT (like TDX RTMR), and fallback to use TPM if no such technology
    specific RoT can be found.

### DIFF
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -51,7 +51,10 @@ type AttestationAgent interface {
 
 type attestRoot interface {
 	// Extend measures the cel content into a measurement register and appends to the CEL.
-	Extend(cel.Content, *cel.CEL) error
+	Extend(cel.Content) error
+	// GetCEL fetches the CEL with events corresponding to the sequence of Extended measurements
+	// to this attestation root
+	GetCEL() *cel.CEL
 	// Attest fetches a technology-specific quote from the root of trust.
 	Attest(nonce []byte) (any, error)
 }
@@ -65,8 +68,8 @@ type AttestAgentOpts struct {
 }
 
 type agent struct {
-	ar               attestRoot
-	cosCel           cel.CEL
+	measuredRots     []attestRoot
+	avRot            attestRoot
 	fetchedAK        *client.Key
 	client           verifier.Client
 	principalFetcher principalIDTokenFetcher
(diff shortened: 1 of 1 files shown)


launcher/agent/agent.go
```diff
@@ -39,6 +40,23 @@ var defaultCELHashAlgo = []crypto.Hash{crypto.SHA256, crypto.SHA1}
 
 type principalIDTokenFetcher func(audience string) ([][]byte, error)
 
+type rotType int
+
+// ordering matters, better root of trust must have a higher index
```
alexmwu, line 45: What constitutes a "better" rot?

launcher/agent/agent.go
```diff
@@ -162,24 +205,30 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
 		},
 	}
 
-	attResult, err := a.ar.Attest(challenge.Nonce)
+	// attest fetch the "best" attest root
+	aaarrr := a.getBestAttestRoot()
```
alexmwu, line 209: nit: ar or aRoot

alexmwu (changes requested):
Not sure I understand this PR. I would have expected us to find all the roots we know about, loop through them, and extend each one with the same event. Potentially have a CEL for each type of root.

jkl73:
> Not sure I understand this PR. I would have expected us to find all the roots we know about, loop through them, and extend each one with the same event. Potentially have a CEL for each type of root.

Currently it will actively try to add `TPM` and `TDXRTMR`.
And yes, each root has it's own CEL, and will extend and record event to all roots.

launcher/agent/agent.go
```diff
@@ -100,27 +120,32 @@ func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher
...
+
 	// check if is a TDX machine
 	qp, err := tg.GetQuoteProvider()
-	if err != nil || qp.IsSupported() != nil {
-		logger.Println("Using TPM PCRs for measurement.")
+	if err != nil {
+		return nil, err
+	}
+	// Use qp.IsSupported to check the TDX RTMR interface is enabled
+	if qp.IsSupported() == nil {
```
alexmwu, line 136: Here, we should create a list of roots of trust to measure to (e.g., `var measuredRots []attestRoot`) and then select the RTMR for the actual attestation verification root of trust (e.g., `avRot`).

```
var tpmAR := &tpmAttestRoot{...}
var measuredRots []*attestRoot
var avRot *attestRoot
// Check if VM is running TDX
...
if qp.IsSupported() == nil {
  tdxAR := &tdxAttestRoot
  measuredRots = append(measuredRots, tdxAR)
  avRot = tdxAr
} else {
  avRot = tpmAR
}
// Unconditionally add TPM.
measuredRots = append(measuredRots, tpmAR)
```

launcher/agent/agent.go
```diff
@@ -39,6 +39,22 @@ var defaultCELHashAlgo = []crypto.Hash{crypto.SHA256, crypto.SHA1}
...
+)
+
+func (r rotType) String() string {
+	return [...]string{"TPMPCR", "TDXRTMR"}[r-1]
+}
+
+func (r rotType) EnumIndex() int {
+	return int(r)
+}
+
```
alexmwu, line 57: All of this is unused. Please remove.

launcher/agent/agent.go
```diff
@@ -51,9 +67,13 @@ type AttestationAgent interface {
...
 type attestRoot interface {
 	// Extend measures the cel content into a measurement register and appends to the CEL.
-	Extend(cel.Content, *cel.CEL) error
+	Extend(cel.Content) error
+	// Get Canonical Event Log, The CEL should be consistent
+	GetCEL() *cel.CEL
 	// Attest fetches a technology-specific quote from the root of trust.
 	Attest(nonce []byte) (any, error)
+	// GetType returns the Root of Trust type.
+	GetType() rotType
```
alexmwu, line 76: This is also unused

launcher/agent/agent.go
```diff
@@ -134,13 +163,20 @@ func (a *agent) Close() error {
...
+			return err
+		}
+	}
+	return nil
 }
 
 // Attest fetches the nonce and connection ID from the Attestation Service,
 // creates an attestation message, and returns the resultant
 // principalIDTokens and Metadata Server-generated ID tokens for the instance.
+// Attest only attests to the "best" Attest Root.
```
alexmwu, line 179: nit: we should define how we select the attest root.
// When possible, Attest uses the technology-specific attestation root-of-trust and otherwise falls back to the vTPM.

jkl73: done

launcher/agent/agent.go
jkl73: deleted

launcher/agent/agent.go
jkl73: deleted

launcher/agent/agent.go
jkl73: Done

launcher/agent/agent.go
```diff
@@ -51,7 +51,9 @@ type AttestationAgent interface {
 
 type attestRoot interface {
 	// Extend measures the cel content into a measurement register and appends to the CEL.
-	Extend(cel.Content, *cel.CEL) error
+	Extend(cel.Content) error
+	// Get Canonical Event Log, The CEL should be consistent
```
alexmwu, line 55: ", The CEL should be consistent"?

Maybe "GetCEL fetches the CEL with events corresponding to the sequence of `Extend`ed measurements attestation root."

alexmwu (approved):

launcher/agent/agent.go
jkl73: yes, updated

PR #518, opened 2024-12-23
Set cos_gpu_installer image reference as part image build steps

meetrajvala:


launcher/image/preload.sh
```diff
@@ -19,6 +20,10 @@ setup_launcher_systemd_unit() {
   cp exit_script.sh "${CS_PATH}/exit_script.sh"
 }
 
+set_cos_gpu_installer_image_reference() {
+  cos-extensions list -- --gpu-installer >> "${CS_PATH}"/"${GPU_INSTALLER_IMAGE_REF}"
+}
```
yawangwang, line 25: I'm thinking about an alternative approach:
You're basically sharing a config to the launcher process, is it possible to pass it as a kernel cmd arg instead of writing to a file? This also makes gpu installation more measurable in some sense. Awaiting for @alexmwu and @jkl73 's inputs on this.

jkl73: >> is it possible to pass it as a kernel cmd arg instead
This means we need to decide the version of gpu installer during the image building. This is my original thought as well, but I remember @meetrajvala said there was a reason to do this during the runtime, because the version may change? @meetrajvala could you elaborate more here?

meetrajvala: @yawangwang Command line will have length constraint of 4096 chars, but currently we are at less than 50% (~1900 chars) so it should not be a big concern. But is it ok to add config required only for userspace program (and not for the customizing the kernel behavior) to kernel command line ?  I agree that it would make the gpu driver installation bit measurable in some sense, though current implementation also addresses it by putting file under `oem` partition which will be sealed in the later step. 

@jkl73 earlier we were discussing about hardcoding the installer version, which would not be very appropriate as we need to keep updating installer version when base COS image changes (via image-family flag). But now we are deriving it using the cos-extensions command (which would always return the specific supported installer image reference corresponds to current base COS image), so we are good.

yawangwang: We have existing examples of adding userspace program configs to kernel command line (e.g., systemd, launcher), so it should be ok to add gpu installer config as well. Also the main purpose of storing files under `oem` partition is to encrypt/protect it using dm-crypt from tampering by the operator. IMO it doesn't provide a very explicit way of measurement.

jkl73: The whole oem partition is sealed and measured, and its hash is part of the cmdline. So I think technically the gpu ref file is still measured and bind to the specific CS image

yawangwang: thanks for your info. My original thought was to provide an explicit way of measurement (the presence of gpu installer ref in kernel cmd), but as you pointed out, since the hash of oem partition is also measured into kernel cmd, I'm also okay with the current approach.

yawangwang (approved):

launcher/image/preload.sh
alexmwu: The goal here is to know exactly what installer we run for a given version of CS. Explicit measurement is not required, since we won't surface what version of the installer we are using to customers. It's mostly nice to have for debugging purposes, but I don't see a huge reason to change this implementation.

alexmwu (changes requested):
We should also pin the GPU installer image digest per-image (similar to how you pin the ref in the OEM partition), since we need to verify the image hasn't changed since we don't own the installer repo. Then we should check the digest matches when we pull it down at boot time.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -138,20 +138,20 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
 	code, _, _ := status.Result()
 
 	if code != 0 {
-		di.logger.Printf("Gpu driver installation task ended and returned non-zero status code %d", code)
+		di.logger.Info(fmt.Sprintf("Gpu driver installation task ended and returned non-zero status code %d", code))
 		return fmt.Errorf("gpu driver installation task ended with non-zero status code %d", code)
 	}
 
-	di.logger.Println("Gpu driver installation task exited with status: 0")
+	di.logger.Info("Gpu driver installation task exited with status: 0")
```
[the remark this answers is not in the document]
meetrajvala: Updated it.

PR #520, opened 2024-12-28
Reduce NPD full config

jessieqliu:
Based on security review recommendations, we should be conservative about which metrics we enable in NPD. This reduces the full list of supported metrics to a smaller one.

### DIFF
--- a/launcher/internal/healthmonitoring/nodeproblemdetector/systemstats_config.go
+++ b/launcher/internal/healthmonitoring/nodeproblemdetector/systemstats_config.go
@@ -51,41 +51,20 @@ func NewSystemStatsConfig() SystemStatsConfig {
 
 var allConfig = &SystemStatsConfig{
 	CPU: &statsConfig{map[string]metricConfig{
-		"cpu/runnable_task_count": {"cpu/runnable_task_count"},
-		"cpu/usage_time":          {"cpu/usage_time"},
-		"cpu/load_1m":             {"cpu/load_1m"},
-		"cpu/load_5m":             {"cpu/load_5m"},
-		"cpu/load_15m":            {"cpu/load_15m"},
-		"system/cpu_stat":         {"system/cpu_stat"},
-		"system/interrupts_total": {"system/interrupts_total"},
-		"system/processes_total":  {"system/processes_total"},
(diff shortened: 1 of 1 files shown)


jkl73 (approved):

launcher/internal/healthmonitoring/nodeproblemdetector/systemstats_config.go
```diff
@@ -51,41 +51,20 @@ func NewSystemStatsConfig() SystemStatsConfig {
...
 	Host: &statsConfig{map[string]metricConfig{
 		"host/uptime": {"host/uptime"},
 	}},
 	Memory: &statsConfig{map[string]metricConfig{
-		"memory/anonymous_used":   {"memory/anonymous_used"},
-		"memory/bytes_used":       {"memory/bytes_used"},
-		"memory/dirty_used":       {"memory/dirty_used"},
-		"memory/page_cache_used":  {"memory/page_cache_used"},
-		"memory/unevictable_used": {"memory/unevictable_used"},
-		"memory/percent_used":     {"memory/percent_used"},
```
alexmwu, line 88: These also seem useful, except perhaps unevictable?

alexmwu (commented):
Should we change the config from "all" then? What if we want to enable the default, full NPD list in the future? https://github.com/google/go-tpm-tools/blob/ef8a29b197bedef56d424c15c2e1e85fafa9f661/launcher/spec/launch_policy.go#L63

jessieqliu:
> Should we change the config from "all" then? What if we want to enable the default, full NPD list in the future?
> 
> https://github.com/google/go-tpm-tools/blob/ef8a29b197bedef56d424c15c2e1e85fafa9f661/launcher/spec/launch_policy.go#L63

My intention was for the "all" term to describe the full set of metrics that CS supports, not necessarily everything supported by NPD. So if in the future we add support for more metrics, the definition would naturally change with it (and we would clearly document this change for the new image). IMO creating a dedicated label for this specific set of metrics going forward is harder to maintain, and it's difficult to think of an appropriate name since it's just what we think is useful at the moment. We could change it to something like "all-supported" for more clarity/flexibility, though. WDYT?

alexmwu (approved):

PR #521, opened 2025-01-02
Add ContainerImageSignature type to verifier client

jessieqliu:
OCI signatures are converted to a different structure in the Rest client for GCA. With the addition of ITA, we want to send a similar structure, so the conversion should be done at a higher level - the agent - rather than being verifier-specific.

The verifier client package will have its own `ContainerSignature` type - which contains only the fields that will be sent to verifier services - rather than using the full `oci.Signature` object.

`fakesignature` is also updated to remove implementations for `PublicKey` and `SigningAlgorthm` since the corresponding `fakeverifier` will use the new `ContainerSignature` type that lacks these methods. Instead, `fakeverifier` will parse the pubkey and sigalg from the payload - similar to how this is done in the actual verification process.

- In `verifier/client`, `VerifyAttestationRequest.ContainerImageSignatures` type is changed from `[]ociSignature` to `[]*ContainerSignature`
- In `verifier/fake`, `verifyContainerImageSignature()` is changed to `extractClaims()`
- In `verifier/oci/cosign/fakesignature`, `fakeSig.PublicKey()` and `fakeSig.SigningAlgorithm` now return "not implemented" errors. Additionally, `fakeSig.Payload()` returns "f.data,f.sigAlg" rather than the data alone.
- In `verifier/rest`, `convertOCISignatureToREST` is removed

### DIFF
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -252,20 +252,9 @@ func convertToContainerSignatures(ociSigs []oci.Signature) ([]*verifier.Containe
 		if err != nil {
 			return nil, fmt.Errorf("failed to decode signature for signature [%v]: %v", ociSig, err)
 		}
-		pubKey, err := ociSig.PublicKey()
-		if err != nil {
-			return nil, fmt.Errorf("failed to get public key from signature [%v]: %v", ociSig, err)
-		}
-		sigAlg, err := ociSig.SigningAlgorithm()
-		if err != nil {
-			return nil, fmt.Errorf("failed to get signing algorithm from signature [%v]: %v", ociSig, err)
-		}
-
 		sigs[i] = &verifier.ContainerSignature{
 			Payload:   payload,
 			Signature: sigBytes,
-			PubKey:    pubKey,
-			SigAlg:    sigAlg,
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


launcher/agent/agent.go
```diff
@@ -210,6 +215,30 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
 	return resp.ClaimsToken, nil
 }
 
+func convertToContainerSignatures(ociSigs []oci.Signature) ([]*verifier.ContainerSignature, error) {
```
yawangwang, line 218: nit: `convertOCIToContainerSignatures`

launcher/agent/agent.go
```diff
@@ -196,8 +197,12 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
 
 	signatures := a.sigsCache.get()
 	if len(signatures) > 0 {
-		req.ContainerImageSignatures = signatures
-		a.logger.Info("Found container image signatures: %v\n", "signatures", signatures)
+		verifierSigs, err := convertToContainerSignatures(signatures)
+		if err != nil {
+			return nil, fmt.Errorf("error converting container signatures: %v", err)
+		}
```
yawangwang, line 203: The current behavior of converting sigs seems to return on error. However, this violates the previous behavior which is logging on error and sending any valid signature found to verifier. We don't want a single failure of conversion to break the entire signature verification process.

jessieqliu: Done. Changed the conversion function to be for one signature (not the full list) and log/continue on error instead of returning.

yawangwang (approved):

jkl73 (approved):

PR #528, opened 2025-01-17
Add Clients struct to attestation agent

jessieqliu:


jessieqliu:
Changed our approach for the clients, this is no longer applicable.

PR #532, opened 2025-01-28
Add ITA Attest to teeserver

jessieqliu:
Add ITA Attestation to the launcher. Adds ITA endpoint to the teeserver and ITA fields to the launchspec.

- Add an `AttestWithClient` method to `AttestationAgent`. Add the same to `fakeAttestationAgent` in `container_runner_test`
- Add `LogFriendly` method to launchspec that returns the launchspec with the ITA Key censored.
- Change `teeserver.New` signature to `New(ctx context.Context, unixSock string, a agent.AttestationAgent, logger logging.Logger, launchSpec spec.LaunchSpec, clients *AttestClients)`
- Add `AttestWithClient` method to teeserver `attestHandler`


launcher/agent/agent.go
```diff
@@ -165,7 +166,20 @@ func (a *agent) MeasureEvent(event cel.Content) error {
...
 func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error) {
-	challenge, err := a.client.CreateChallenge(ctx)
+	if a.client == nil {
+		return nil, fmt.Errorf("attest agent does not have initialized verifier client")
+	}
+
+	return a.AttestWithClient(ctx, opts, a.client)
+}
+
+// Attest fetches the nonce and connection ID from the Attestation Service via the provided client,
```
JoshuaKrstic, line 176: nit: Clean up comment for new method

launcher/container_runner.go
```diff
@@ -552,13 +560,34 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to measure CEL events: %v", err)
 	}
 
-	if err := r.fetchAndWriteToken(ctx); err != nil {
-		return fmt.Errorf("failed to fetch and write OIDC token: %v", err)
+	// Only refresh token if agent has a default GCA client (not ITA use case).
+	if r.launchSpec.ITARegion == "" {
```
JoshuaKrstic, line 564: This feels odd. Maybe instead we give each verifier a "auto refresh" method that we call into here? For ITA we can pass a method that does nothing and give this one to the GCA method. WDYT?

launcher/container_runner.go
```diff
@@ -552,13 +560,34 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
+			return fmt.Errorf("failed to fetch and write OIDC token: %v", err)
+		}
 	}
 
 	// create and start the TEE server
 	r.logger.Info("EnableOnDemandAttestation is enabled: initializing TEE server.")
-	teeServer, err := teeserver.New(ctx, path.Join(launcherfile.HostTmpPath, teeServerSocket), r.attestAgent, r.logger)
+
+	attestClients := &teeserver.AttestClients{}
+	if r.launchSpec.ITARegion != "" {
```
JoshuaKrstic, line 574: This is the third time we've checked this.

Can we refactor this to create the clients at the top, and then change our behavior given which clients we were able to make?

launcher/container_runner.go
```diff
@@ -552,13 +560,34 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
+		itaClient, err := ita.NewClient(r.launchSpec.ITARegion, r.launchSpec.ITAKey)
+		if err != nil {
+			return fmt.Errorf("failed to create ITA client: %v", err)
+		}
+
+		attestClients.ITA = itaClient
+	} else {
```
JoshuaKrstic, line 581: Some customers need both to help transition or just because some of their end users want to root in intel and some want to root in Google.

I think we can just remove the if-else and create both

launcher/spec/launch_spec.go
```diff
@@ -240,9 +244,34 @@ func (s *LaunchSpec) UnmarshalJSON(b []byte) error {
...
 
+	if s.Experiments.EnableItaVerifier {
+		itaRegionVal, itaRegionOK := unmarshaledMap[itaRegion]
+		itaKeyVal, itaKeyOK := unmarshaledMap[itaKey]
+
+		if itaRegionOK != itaKeyOK {
+			return fmt.Errorf("ITA fields %s and %s must both be provided", itaRegion, itaKey)
+		}
+
+		if itaRegionOK && itaRegionVal != "" {
```
JoshuaKrstic, line 255: We should check the empty strings above when we emit an error

launcher/teeserver/tee_server.go
```diff
@@ -67,28 +78,69 @@ func (a *attestHandler) Handler() http.Handler {
...
+}
+
 // getDefaultToken handles the request to get the default OIDC token.
 // For now this function will just read the content of the file and return.
 // Later, this function can use attestation agent to get a token directly.
 func (a *attestHandler) getToken(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "text/html")
 
+	// If the handler does not have a GCA client, create one.
+	if a.clients.GCA == nil {
```
JoshuaKrstic, line 98: From my other comments we should just do this in the container_runner

JoshuaKrstic (changes requested):

launcher/container_runner.go
jessieqliu: Hmm IMO that seems complicated. If we're not refreshing the ITA token it's simpler to not deal with a coroutine at all and disable all that logic at the top level (container runner). The code is also more readable in this case (ITA = no refresh), despite the awkward conditional check.

Would it be better if we change the conditional to a nil check for the GCA client? Or any suitable check related to GCA rather than ITA?

launcher/container_runner.go
jessieqliu: I think it'd be better to create each client as-needed. The GCA client in particular requires RPCs (Get/List Location) to be initialized, and we shouldn't introduce those potential failure points unless the customer actually wants a GCA token (meaning they request it ODA). 

If the GCA is having problems - for example - we wouldn't want ITA-only customers to be blocked by GCA client initialization. A way around this could be to fail-open on creating the GCA client in that case, but I think that only further complicates things.

launcher/container_runner.go
jessieqliu: Resolved offline

JoshuaKrstic (approved):
Approved with my changes coming in a follow up CL

launcher/agent/agent.go
```diff
@@ -164,7 +165,20 @@ func (a *agent) MeasureEvent(event cel.Content) error {
...
 func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error) {
-	challenge, err := a.client.CreateChallenge(ctx)
+	if a.client == nil {
+		return nil, fmt.Errorf("attest agent does not have initialized verifier client")
+	}
+
+	return a.AttestWithClient(ctx, opts, a.client)
+}
+
+// AttestWihtClient fetches the nonce and connection ID from the Attestation Service via the provided client,
```
alexmwu, line 175: nit: AttestWithClient

launcher/teeserver/tee_server.go
```diff
@@ -43,10 +52,12 @@ func New(ctx context.Context, unixSock string, a agent.AttestationAgent, logger
 		netListener: nl,
 		server: &http.Server{
 			Handler: (&attestHandler{
-				ctx:              ctx,
-				attestAgent:      a,
-				defaultTokenFile: filepath.Join(launcherfile.HostTmpPath, launcherfile.AttestationVerifierTokenFilename),
-				logger:           logger,
+				ctx:         ctx,
+				attestAgent: a,
+				// defaultTokenFile: filepath.Join(launcherfile.HostTmpPath, launcherfile.AttestationVerifierTokenFilename),
```
alexmwu, line 57: Why is this commented out?

alexmwu (approved):

launcher/teeserver/tee_server.go
jessieqliu: Deleted

PR #540, opened 2025-02-06
Fix a grub eventlog parsing issue with null terminator

jkl73:
Current grub parsing logic considers both GRUB null terminator cases (we check the hash digest on both raw data with or without the "\x00" byte)

There is a possibility for someone to append a random byte (not \x00) at end of the raw data in the eventlog. The random byte won't be included in the digest verification, but will show up as a piece of verified raw data.

When we add the raw data to the struct, we always include "\x00" byte if it exists in raw data. This might causing confusion since the raw data in our struct might not exactly match the hash digest. User needs to remove the null terminator from the raw data manually if they want to use it.

### DIFF
--- a/server/eventlog.go
+++ b/server/eventlog.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"hash"
 
 	"github.com/google/go-attestation/attest"
 	"github.com/google/go-eventlog/register"
@@ -445,6 +446,7 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
 	var files []*pb.GrubFile
 	var commands []string
 	for idx, event := range events {
+		hasher := hash.New()
 		index := event.GetPcrIndex()
 		if index != 8 && index != 9 {
 			continue
@@ -463,7 +465,6 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
 			files = append(files, &pb.GrubFile{Digest: event.GetDigest(),
 				UntrustedFilename: event.GetData()})
 		} else if index == 8 {
-			hasher := hash.New()
 			suffixAt := -1
 			rawData := event.GetData()
 			for _, prefix := range validPrefixes {
@@ -475,14 +476,16 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
 			if suffixAt == -1 {
 				return nil, fmt.Errorf("invalid prefix seen for PCR%d event: %s", index, rawData)
 			}
-			hasher.Write(rawData[suffixAt : len(rawData)-1])
-			if !bytes.Equal(event.Digest, hasher.Sum(nil)) {
-				// Older GRUBs measure "grub_cmd " with the null terminator.
-				// However, "grub_kernel_cmdline " measurements also ignore the null terminator.
-				hasher.Reset()
-				hasher.Write(rawData[suffixAt:])
-				if !bytes.Equal(event.Digest, hasher.Sum(nil)) {
-					return nil, fmt.Errorf("invalid digest seen for GRUB event log in event %d: %s", idx, hex.EncodeToString(event.Digest))
+
+			// Check the slice is not empty after the suffix, which ensures rawData[len(rawData)-1] is not part
+			// of the suffix.
+			if len(rawData[suffixAt:]) > 0 && rawData[len(rawData)-1] == '\x00' {
+				if err := verifyNullTerminatedDataDigest(hasher, rawData[suffixAt:], event.Digest); err != nil {
+					return nil, fmt.Errorf("invalid GRUB event (null-terminated) #%d: %v", idx, err)
+				}
+			} else {
+				if err := verifyDataDigest(hasher, rawData[suffixAt:], event.Digest); err != nil {
+					return nil, fmt.Errorf("invalid GRUB event #%d: %v", idx, err)
 				}
 			}
 			hasher.Reset()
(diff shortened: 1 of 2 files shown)


alexmwu:
I don't see this as an issue. We should have the output be the exact event data that was matched rather than mutate it. We can vend a normalize string function, or someone can implement their own.

jkl73:
> I don't see this as an issue. We should have the output be the exact event data that was matched rather than mutate it. We can vend a normalize string function, or someone can implement their own.

I'm worrying about the possibility for someone to append an random byte (not \x00) at end of the raw data in the eventlog. The random byte won't be included in the digest verification.

Another way might check the last byte to make sure it is a null byte (thus will be normalized)?

alexmwu:
> > I don't see this as an issue. We should have the output be the exact event data that was matched rather than mutate it. We can vend a normalize string function, or someone can implement their own.
> 
> I'm worrying about the possibility for someone to append an random byte (not \x00) at end of the raw data in the eventlog. The random byte won't be included in the digest verification.
> 
> Another way might check the last byte to make sure it is a null byte (thus will be normalized)?

I see. So then, we should only allow stripping the last byte if it's 0x00. I don't think the time to remove the byte is after we confirm the digest but before.
```
                                 // Older GRUBs measure "grub_cmd " with the null terminator.
				// However, "grub_kernel_cmdline " measurements also ignore the null terminator.
       if rawData[len(rawData)-1] == 0 {
         if err := verifyNullTerminatedData(rawData); err != nil {
           return nil, fmt.Errorf("invalid GRUB event log in event %d", idx)
        }
        else {
          if err := verifyData(rawData); err != nil {
           return nil, fmt.Errorf("invalid GRUB event log in event %d", idx)
          }
         			hasher.Reset()
			commands = append(commands, string(rawData))
  }
}

verifyNullTerminatedData(hasher hash.Hash, rawData []byte) error {

  hasher.Write(rawData[suffixAt : len(rawData)-1])
  if !bytes.Equal(event.Digest, hasher.Sum(nil)) {
    hasher.Reset()
    hasher.Write(rawData[suffixAt:])
    if err := verifyData(hasher, rawData); err != nil {
      return err
    }
  }
  return nil
}

verifyData(hasher hash.Hash, rawData []byte) error {
 if !bytes.Equal(event.Digest, hasher.Sum(nil)) {
      return fmt.Errorf("invalid digest:, hex.EncodeToString(event.Digest))
    }
}
```

This also needs to be backported to go-eventlog

yawangwang:
> > I don't see this as an issue. We should have the output be the exact event data that was matched rather than mutate it. We can vend a normalize string function, or someone can implement their own.
> 
> I'm worrying about the possibility for someone to append an random byte (not \x00) at end of the raw data in the eventlog. The random byte won't be included in the digest verification.
> 
> Another way might check the last byte to make sure it is a null byte (thus will be normalized)?

I would say this is a more valuable point because this fix will avoid the last byte being potentially exploited, could you add it to the PR description as well?

jkl73:
> This also needs to be backported to go-eventlog
track here https://github.com/google/go-eventlog/issues/11

server/eventlog.go
```diff
@@ -475,17 +474,17 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
...
-			hasher.Write(rawData[suffixAt : len(rawData)-1])
-			if !bytes.Equal(event.Digest, hasher.Sum(nil)) {
-				// Older GRUBs measure "grub_cmd " with the null terminator.
-				// However, "grub_kernel_cmdline " measurements also ignore the null terminator.
-				hasher.Reset()
-				hasher.Write(rawData[suffixAt:])
-				if !bytes.Equal(event.Digest, hasher.Sum(nil)) {
-					return nil, fmt.Errorf("invalid digest seen for GRUB event log in event %d: %s", idx, hex.EncodeToString(event.Digest))
+
+			if len(rawData) > suffixAt && rawData[len(rawData)-1] == '\x00' {
```
alexmwu, line 478: why is the len(rawData) > suffixAt check needed? you just need to check it's null-terminated.

server/eventlog.go
```diff
@@ -495,6 +494,32 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
 	return &pb.GrubState{Files: files, Commands: commands}, nil
 }
 
+// verifyNullTerminatedRawData checks the digest of the data.
+// Returns nil if digest match the hash of the data or the data without the last bytes (\x00).
+// The caller needs to make sure len(data) is at least 1, and data is ended with '\x00',
+// otherwise this function will return an error.
+func verifyNullTerminatedDataDigest(hash crypto.Hash, data []byte, digest []byte) error {
+	if len(data) < 1 || data[len(data)-1] != '\x00' {
```
alexmwu, line 502: nit: len(data) == 0 reads better

server/eventlog.go
```diff
@@ -495,6 +494,32 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
...
+		if err := verifyDataDigest(hash, data[:len(data)-1], digest); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// verifyDataDigest checks the digest of the data.
+func verifyDataDigest(hash crypto.Hash, data []byte, digest []byte) error {
+	hasher := hash.New()
```
alexmwu, line 515: why are we creating a new hasher everytime? It's faster to reuse the same hash.Hash object.

server/eventlog.go
jkl73: I want to make sure after the suffix, rawData has at least one byte.

I was thinking if we ever add an empty string as an acceptable suffix, we need to check the length of the rawData to make sure the check is not out of bound.

server/eventlog.go
jkl73: changed

server/eventlog.go
jkl73: Did that just for simplicity. But ya I will add reset() and reuse the hasher for performance.

server/eventlog.go
alexmwu: We can cover that with a test. I would prefer to do that over making already complex code harder to read

server/eventlog.go
```diff
@@ -495,6 +496,32 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
...
+		if err := verifyDataDigest(hasher, data[:len(data)-1], digest); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// verifyDataDigest checks the digest of the data.
+func verifyDataDigest(hasher hash.Hash, data []byte, digest []byte) error {
+	hasher.Reset()
```
alexmwu, line 517: nit: reset at the end

alexmwu (approved):

server/eventlog.go
jkl73: done

jkl73:
>We can cover that with a test. I would prefer to do that over making already complex code harder to read
I changed the it to `len(rawData[suffixAt:]) > 0` and add some comment, hopefully it's more clear right now (I'm not sure how to cover this in a test..)

server/eventlog.go
```diff
@@ -445,6 +446,7 @@ func getGrubState(hash crypto.Hash, events []*pb.Event) (*pb.GrubState, error) {
 	var files []*pb.GrubFile
 	var commands []string
 	for idx, event := range events {
+		hasher := hash.New()
```
yawangwang, line 449: nit: move to hasher to its original place to keep it with `hasher.Reset` in a same scope.

yawangwang (approved):

server/eventlog.go
alexmwu: > I changed the it to len(rawData[suffixAt:]) > 0 and add some comment, hopefully it's more clear right now (I'm not sure how to cover this in a test..)

A simple comment or test that validPrefixes are all not "" if your concern is we ever make "" a valid prefix. I really don't think that would ever happen is why I think this check is unnecessary, but I'll leave this up to you.

alexmwu (changes requested):
Sorry, I realize you don't have tests that cover your concern: that the last byte can be manipulated to be a non-null terminator. Please add a test to cover this case by modifying a test with another byte and checking the digest fails to validate

alexmwu (approved):

PR #542, opened 2025-02-18
Update GPU driver installation flow for supporting confidential-gpus

meetrajvala:
This PR updates the GPU driver installation flow for supporting the confidential GPUs. 

**Existing GPU driver installation flow:**

- Install the GPU drivers vis cos_gpu_installer

  - Downloads the GPU driver and required artifacts from GCS bucket
  - Loads kernel modules
  - Verify the GPU driver installation
 
**Updated GPU driver installation flow for supporting the confidential GPUs:**

- Install the GPU drivers via cos_gpu_installer with `--no-verify` flag

  - Downloads the GPU driver and required artifacts from GCS bucket (Skips the kernel module loading and verification step)

- Load nvidia kernel modules
- Start `nvidia-persistenced process`
- Verify GPU driver installation
- Set GPU state to READY (explicitly)

Following steps we perform manually for installing the GPU drivers for confidential GPUs which are accommodated in launcher as part of this PR.
```
1. /cos-gpu-installer install -version=default -host-dir=/var/lib/nvidia --no-verify
2. sudo /sbin/insmod /var/lib/nvidia/drivers/nvidia.ko
3. sudo /sbin/insmod /var/lib/nvidia/drivers/nvidia-uvm.ko
4. sudo /sbin/insmod /var/lib/nvidia/drivers/nvidia-modeset.ko
5. sudo /sbin/insmod /var/lib/nvidia/drivers/nvidia-drm.ko
6. sudo mount --bind /var/lib/nvidia /var/lib/nvidia
7. sudo mount -o remount,exec /var/lib/nvidia
8. sudo /var/lib/nvidia/bin/nvidia-persistenced
9. sudo /var/lib/nvidia/bin/nvidia-smi
```

**Testing:**
Image test will be added separately once confidential GPU quota is setup for project. Currently, it is only being tested with regression tests using non-confidential GPUs.


launcher/image/preload.sh
```diff
@@ -19,6 +19,10 @@ setup_launcher_systemd_unit() {
   cp exit_script.sh "${CS_PATH}/exit_script.sh"
 }
 
+setup_nvidia_persistenced_systemd_unit() {
+  cp nvidia-persistenced.service "${CS_PATH}/nvidia-persistenced.service"
```
yawangwang, line 23: Does nvidia-persistenced.service get pre-installed on cos image when building CS images?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -105,7 +108,7 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
 		containerd.WithNewSnapshot(installerSnapshotID, image),
 		containerd.WithNewSpec(oci.WithImageConfig(image),
 			oci.WithPrivileged,
-			oci.WithProcessArgs("/cos-gpu-installer", "install", "-version=default", fmt.Sprintf("-host-dir=%s", InstallationHostDir)),
+			oci.WithProcessArgs("/cos-gpu-installer", "install", "-version=default", fmt.Sprintf("-host-dir=%s", InstallationHostDir), "--no-verify"),
```
yawangwang, line 111: nit: could you add comments to explain why --no-verify is needed?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -167,3 +190,71 @@ func remountAsExecutable(dir string) error {
...
+	s, err := systemctl.New()
+	if err != nil {
+		return fmt.Errorf("failed to create systemctl client: %v", err)
+	}
+	defer s.Close()
+
+	logger.Info("Starting nvidia-persistenced.service")
+	if err := s.Start("nvidia-persistenced.service"); err != nil {
+		return fmt.Errorf("failed to start nvidia-persistenced.service: %v", err)
+	}
```
yawangwang, line 228: From https://download.nvidia.com/XFree86/Linux-x86_64/396.51/README/nvidia-persistenced.html, the daemon utility nvidia-persistenced is installed by the NVIDIA Linux GPU driver installer, but it is not installed to run on system startup. So I guess the execution binary `/var/lib/nvidia/bin/nvidia-persistenced` is downloaded after the containterd launches `cos-gpu-driver-installer` workload. 

Could you confirm with the COS team about whether this `/var/lib/nvidia/bin/nvidia-persistenced` is pre-installed on COS? If not, then you can probably remove `nvidia-persistenced.service` from entrypoint.sh.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -167,3 +190,71 @@ func remountAsExecutable(dir string) error {
...
+	// Run nvidia-smi conf-compute command to set GPU state to READY.
+	if err := exec.Command("nvidia-smi", "conf-compute", "-srs", "1").Run(); err != nil {
+		return fmt.Errorf("failed to set the gpu state to ready: %v", err)
+	}
+	return nil
+}
+
+func isGPUCCModeEnabled() bool {
+	// Run nvidia-smi conf-compute command to check if confidential compute mode is ON.
+	ccModeOutput, _ := exec.Command("nvidia-smi", "conf-compute", "-f").Output()
```
yawangwang, line 214: nit: consider returning error as well.

yawangwang:
Could you add more context to the description around what steps are required to enable cGPU driver installation for cos-tdx image?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -167,3 +190,71 @@ func remountAsExecutable(dir string) error {
...
+	nvidiaModeset := &modules.Module{
+		Name: "nvidia_modeset",
+		Path: filepath.Join(kernelModulePath, "nvidia-modeset.ko"),
+	}
+	nvidiaDrm := &modules.Module{
+		Name: "nvidia_drm",
+		Path: filepath.Join(kernelModulePath, "nvidia-drm.ko"),
+	}
+	// Need to load modules in order due to module dependency.
+	gpuModules := []*modules.Module{nvidia, nvidiaUvm, nvidiaModeset, nvidiaDrm}
```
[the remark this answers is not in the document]
meetrajvala: Yes.

launcher/image/preload.sh
meetrajvala: nvidia-persistenced.service is not part of base COS image, we are adding it while building CS image.

launcher/internal/gpu/driverinstaller.go
meetrajvala: `/var/lib/nvidia/bin/nvidia-persistenced` is not pre-installed on base COS image and it get installed by `cos-gpu-installer`. As part of entrypoint, we are copying the systemd unit file (as installation path is fixed and known), and nvidia-persistenced systemd process will started only after binary is downloaded by `cos-gpu-driver-installer`.

meetrajvala: In fact we do not need to launch it as systemd process because we would run the launcher on each reboot. Let me update it to just run the binary via launcher instead of systemd process.

yawangwang: Yeah my point is there's no need to execute persistenced as a systemd process unless it's pre-installed on the base COS. I think running the binary should be fine.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -167,3 +198,82 @@ func remountAsExecutable(dir string) error {
...
+	os.Setenv("PATH", newPathEnv)
+	logger.Info("Starting nvidia-persistenced process")
+	if err := exec.Command("nvidia-persistenced").Run(); err != nil {
+		return fmt.Errorf("failed to launch nvidia-persistenced daemon: %v", err)
+	}
+	logger.Info("nvidia-persistenced daemon successfully started")
+	return nil
+}
+
+func loadGPUDrivers(moduleParams modules.ModuleParameters) error {
```
jkl73, line 253: loadNvidiaKO sounds more accurate

Also why do we need to manually load kernel modules, is that because the --no-verify flag?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -167,3 +198,82 @@ func remountAsExecutable(dir string) error {
...
+	default:
+		return false
+	}
+}
+
+func launchNvidiaPersistencedProcess(logger logging.Logger) error {
+	newPathEnv := fmt.Sprintf("%s/bin:%s", InstallationHostDir, os.Getenv("PATH"))
+	os.Setenv("PATH", newPathEnv)
+	logger.Info("Starting nvidia-persistenced process")
+	if err := exec.Command("nvidia-persistenced").Run(); err != nil {
```
jkl73, line 246: Ideally we should use the systemd to manage the lifecycle of this process, but I guess for now it is fine

jkl73 (commented):
Mostly LGTM, are you waiting on me to add h100 quota to add test?

launcher/internal/gpu/driverinstaller.go
meetrajvala: Yes, it is because of `--no-verify` flag. Use of `--no-verify` flag also skips the loading of kernel modules. COS  is working on adding one more flag specific to skip `nvidia-smi` step. Once that is available, we can remove this function and would no longer require to load modules manually.

meetrajvala:
> Mostly LGTM, are you waiting on me to add h100 quota to add test?

It is not blocker as we are merging it is feature branch. I am planning to add image tests in follow-up PR once h100 quota is setup.

jkl73 (commented):
LGTM, make sure to get LGTM from yawang and/or alex before merging.

I feel like the whole GPU installation process is adding a lot of complexity to the launcher binary, this might be a good candidate to start as a new component/binary/process parallel to the launcher binary. We can discuss more later, for now it is okay to have it in this branch.

meetrajvala:
> LGTM, make sure to get LGTM from yawang and/or alex before merging.
> 
> I feel like the whole GPU installation process is adding a lot of complexity to the launcher binary, this might be a good candidate to start as a new component/binary/process parallel to the launcher binary. We can discuss more later, for now it is okay to have it in this branch.

I am happy to discuss this. We also have feature request open with COS team to simplify this, mainly for confidential GPU support. which should reduce some the steps in future, which are added as part of this PR.

jkl73 (approved):

yawangwang:
> > LGTM, make sure to get LGTM from yawang and/or alex before merging.
> > I feel like the whole GPU installation process is adding a lot of complexity to the launcher binary, this might be a good candidate to start as a new component/binary/process parallel to the launcher binary. We can discuss more later, for now it is okay to have it in this branch.
> 
> I am happy to discuss this. We also have feature request open with COS team to simplify this, mainly for confidential GPU support. which should reduce some the steps in future, which are added as part of this PR.

yawangwang (approved):

launcher/internal/gpu/driverinstaller.go
```diff
@@ -167,3 +198,82 @@ func remountAsExecutable(dir string) error {
...
+		return true
+	default:
+		return false
+	}
+}
+
+func launchNvidiaPersistencedProcess(logger logging.Logger) error {
+	newPathEnv := fmt.Sprintf("%s/bin:%s", InstallationHostDir, os.Getenv("PATH"))
+	os.Setenv("PATH", newPathEnv)
+	logger.Info("Starting nvidia-persistenced process")
```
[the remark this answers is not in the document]
meetrajvala: I added it to path variable as we were running it multiple times later. I am ok running it directly.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -138,11 +141,31 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
...
 		return fmt.Errorf("gpu driver installation task ended with non-zero status code %d", code)
 	}
 
-	di.logger.Println("Gpu driver installation task exited with status: 0")
+	moduleParams := modules.NewModuleParameters()
+	if err = loadGPUDrivers(moduleParams); err != nil {
+		return fmt.Errorf("failed load GPU drivers: %v", err)
+	}
+
+	if err = startNvidiaPersistencedService(di.logger); err != nil {
```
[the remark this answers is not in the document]
meetrajvala: `nvidia-persistenced` works fine for non-confidential GPUs too as it simply sleeps while maintaining a reference to the NVIDIA device state and prevents the NVIDIA kernel driver from exiting. If we install drivers using `cos-extensions`, it  also runs this daemon as part of `post_install` step for non confidential GPUs. Considering this, we do not required to gate this with `isGPUCCModeEnabled()`.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -167,3 +198,82 @@ func remountAsExecutable(dir string) error {
...
+	if err := exec.Command("nvidia-smi", "conf-compute", "-srs", "1").Run(); err != nil {
+		return fmt.Errorf("failed to set the gpu state to ready: %v", err)
+	}
+	return nil
+}
+
+func isGPUCCModeEnabled(logger logging.Logger, gpuType deviceinfo.GPUType) (bool, error) {
+	// The nvidia-smi conf-compute command fails for GPU which doesn't support confidential computing.
+	// This check would bypass nvidia-smi conf-compute command for GPU not having confidential compute support.
+	if !isCCSupportedGpu(gpuType) {
```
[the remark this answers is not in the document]
meetrajvala: Agree that this would cause launcher change whenever there is new GPU type.

Reason I updated to GPU type check was because of the different errors I was getting for different non-confidential GPUs. For A100 and H100 (with CC mode on), I was getting output "CC Mode : Off" and "CC Mode: On" respectively with exit code 0. But for  V100 and T4 I was getting output "No CC capable devices found" with exit code 6 while running `nvidia-smi conf-compute` command.

I have updated `isGPUCCModeEnabled()` function to not to use the hardcode values, instead it would check the output of `nvidia-smi conf-compute -f`:

If output contains "No CC capable devices found" irrespective of error-code 
     =>  We can say that CC is not supported.
Then check if error not nil
     => if err is not nil then there is actual some error while running `nvidia-smi conf-compute`
Finally
     => Check if CC mode is On or Off in `nvidia-smi conf-compute` output and return nil value for error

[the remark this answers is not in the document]
meetrajvala: I have updated `isGPUCCModeEnabled()` function to not to use the hardcode values, instead it would check the output of `nvidia-smi conf-compute -f`:

If output contains "No CC capable devices found" irrespective of error-code 
     =>  We can say that CC is not supported.
Then check if error not nil
     => if err is not nil then there is actual some error while running `nvidia-smi conf-compute`
Finally
     => Check if CC mode is On or Off in `nvidia-smi conf-compute` output and return nil value for error

PR #543, opened 2025-02-19
merge tdx_rtmr to main

jkl73:
merging rtmr branch to main.

cs main image starts to based on cos-tdx-113-lts which will be used for both sev and tdx.

### DIFF
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -12,12 +12,20 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
+	"github.com/google/go-configfs-tsm/configfs/configfsi"
+
+	"github.com/google/go-configfs-tsm/configfs/linuxtsm"
+	tg "github.com/google/go-tdx-guest/client"
+	tlabi "github.com/google/go-tdx-guest/client/linuxabi"
+
 	"github.com/google/go-tpm-tools/cel"
 	"github.com/google/go-tpm-tools/client"
+	"github.com/google/go-tpm-tools/internal"
 	"github.com/google/go-tpm-tools/launcher/internal/logging"
 	"github.com/google/go-tpm-tools/launcher/internal/signaturediscovery"
 	"github.com/google/go-tpm-tools/launcher/spec"
@@ -42,6 +50,16 @@ type AttestationAgent interface {
 	Close() error
 }
 
+type attestRoot interface {
+	// Extend measures the cel content into a measurement register and appends to the CEL.
+	Extend(cel.Content) error
+	// GetCEL fetches the CEL with events corresponding to the sequence of Extended measurements
+	// to this attestation root
+	GetCEL() *cel.CEL
+	// Attest fetches a technology-specific quote from the root of trust.
+	Attest(nonce []byte) (any, error)
+}
+
 // AttestAgentOpts contains user generated options when calling the
 // VerifyAttestation API
 type AttestAgentOpts struct {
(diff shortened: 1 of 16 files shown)


launcher/agent/agent.go
```diff
@@ -94,15 +147,21 @@ func (a *agent) Close() error {
...
+			return err
+		}
+	}
+	return nil
 }
 
 // Attest fetches the nonce and connection ID from the Attestation Service,
 // creates an attestation message, and returns the resultant
 // principalIDTokens and Metadata Server-generated ID tokens for the instance.
+// When possible, Attest uses the technology-specific attestation root-of-trust
```
alexmwu, line 163: I don't think this is accurate. If it is, we shouldn't fall back and just fail.

launcher/agent/agent.go
```diff
@@ -114,27 +173,49 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
...
 	req := verifier.VerifyAttestationRequest{
 		Challenge:      challenge,
 		GcpCredentials: principalTokens,
-		Attestation:    attestation,
 		TokenOptions:   opts.TokenOptions,
 	}
 
+	switch v := attResult.(type) {
+	case *pb.Attestation:
+		a.logger.Info("attestation through TPM quote")
```
alexmwu, line 195: I think this log line is extraneous. We already have a log line when we select the avRot.

launcher/agent/agent.go
```diff
@@ -114,27 +173,49 @@ func (a *agent) Attest(ctx context.Context, opts AttestAgentOpts) ([]byte, error
...
+
+		v.CanonicalEventLog = cosCel.Bytes()
+		req.Attestation = v
+	case *verifier.TDCCELAttestation:
+		a.logger.Info("attestation through TDX quote")
```
alexmwu, line 200: Same here.

verifier/rest/rest.go
```diff
@@ -230,20 +237,67 @@ func convertRequestToREST(request verifier.VerifyAttestationRequest) *ccpb.Verif
...
+
+		verifyReq.TpmAttestation = &confidentialcomputingpb.TpmAttestation{
+			Quotes:            quotes,
+			TcgEventLog:       request.Attestation.GetEventLog(),
+			CanonicalEventLog: request.Attestation.GetCanonicalEventLog(),
+			AkCert:            request.Attestation.GetAkCert(),
+			CertChain:         certs,
+		}
+
+		if request.Attestation.GetSevSnpAttestation() != nil {
```
alexmwu, line 270: Why are we handling this? If this is for the cmdline, let's go and make those changes to force the cmdline to use the API correctly rather than have this complex logic

verifier/rest/rest.go
```diff
@@ -230,20 +237,67 @@ func convertRequestToREST(request verifier.VerifyAttestationRequest) *ccpb.Verif
...
+			sevsnp, err := convertSEVSNPProtoToREST(request.Attestation.GetSevSnpAttestation())
+			if err != nil {
+				log.Fatalf("Failed to convert SEVSNP proto to API proto: %v", err)
+			}
+			verifyReq.TeeAttestation = sevsnp
+		}
+
+		if request.Attestation.GetTdxAttestation() != nil {
```
alexmwu, line 278: Same comment here for TDX: make the cmdline tool use go-tdx-guest directly

alexmwu (changes requested):

launcher/agent/agent.go
jkl73: Corrected

launcher/agent/agent.go
jkl73: right, but I'd prefer to have some extra logging here, might be helpful for debugging later once there are more attestation type.

verifier/rest/rest.go
jkl73: The logic is added here https://github.com/google/go-tpm-tools/commit/b30d3733c5835cb35cfeabc68a194bbf98f18fd7
I think it's for the cmd tool.

I think we currently don't have a seperate API for SNP yet.

We probably need to create seperate PR for this.

alexmwu (approved):

PR #544, opened 2025-02-26
Refactor multiple client setup and clean up surrounding code

JoshuaKrstic:
- Updated where verifier clients are managed to be flexible to adding new verifiers in the future
- Cleaned up some surrounding code (like changing the token refreshing to only happen within a loop)
- Removed "ConnID" from the challenge as it seems to be unused.
- Updated the GCA rest client to lazily initialize its location to prevent requests from hitting the GCA when the workload doesn't use it as its verifier


launcher/models/AttestClients.go
jessieqliu, line 1: nit: snake case instead of camel case file names, for consistency

PR #556, opened 2025-03-07
Add image tests to validate Confidential GPU support

meetrajvala:
This PR adds required image test for confidential GPU support in CS. It also updates the `create_gpu_vm.sh` script to support confidential VM as we need launch TDX VM for confidential GPU. Note that we got the quota for `preemptible` H100 GPUs (which are sufficient for image test runs), so `create_gpu_vm.sh` is updated accordingly.

**Testing:**

- Image test with confidential GPU completed successfully.


launcher/image/test/test_gpu_driver_installation_cloudbuild.yaml
```diff
@@ -75,6 +90,10 @@ steps:
   id: NoGpuWorkloadTest
   entrypoint: 'bash'
   args: ['scripts/gpu/test_gpu_nogpu.sh', '${_VM_NAME_PREFIX}-${BUILD_ID}-nogpu', '${_ZONE}']
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: ConfidentialGpuWorkloadTest
+  entrypoint: 'bash'
+  args: ['scripts/gpu/test_gpu_workload.sh', '${_VM_NAME_PREFIX}-${BUILD_ID}-tdxcvm-cgpu', 'us-east5-a']
```
yawangwang, line 96: Just want to confirm, the script `test_gpu_workload.sh` checks the existence of 'Test PASSED' from logs to verify if the sample workload gets executed using GPU. What's the source of the log 'Test PASSED'? Does it come from `cuda-vector-add:latest`?

jkl73: +1

jkl73 (approved):

launcher/image/test/test_gpu_driver_installation_cloudbuild.yaml
meetrajvala: Yes, it comes from `cuda-vector-add` container. Source for the test GPU workload container: https://github.com/NVIDIA/cuda-samples/blob/master/Samples/0_Introduction/vectorAdd/vectorAdd.cu

I have updated test scripts and docker file with reference of source code.

PR #557, opened 2025-03-10
Remove filesystem and timer dependencies from container_runner_test.go

JoshuaKrstic:
This PR aims to remove time.Sleep() from container_runner_test.go to reduce flakiness and reduce test runtime.

To do this I added:

- A timer interface
- A real and fake implementation of the timer interface.
- A data writer interface
- A real and fake implementation of a token writer

I put these in "models" to separate them from the core "container_runner" logic.

In addition, the test TestTokenIsNotChangedIfRefreshFails was removed as it is redundant with TestTokenRefreshRetryPolicyFail.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -32,6 +32,7 @@ import (
 	"github.com/google/go-tpm-tools/launcher/internal/logging"
 	"github.com/google/go-tpm-tools/launcher/internal/signaturediscovery"
 	"github.com/google/go-tpm-tools/launcher/launcherfile"
+	"github.com/google/go-tpm-tools/launcher/models"
 	"github.com/google/go-tpm-tools/launcher/registryauth"
 	"github.com/google/go-tpm-tools/launcher/spec"
 	"github.com/google/go-tpm-tools/launcher/teeserver"
@@ -48,9 +49,9 @@ type ContainerRunner struct {
 	attestAgent   agent.AttestationAgent
 	logger        logging.Logger
 	serialConsole *os.File
+	tokenWriter   models.DataWriter
+	timer         models.Timer
 }
-
-const tokenFileTmp = ".token.tmp"
 
 const teeServerSocket = "teeserver.sock"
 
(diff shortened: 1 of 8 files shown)


launcher/models/data_writer.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// FileWriter is a tokenWriter that writes the token to a file.
+type FileWriter struct {
+	directory string
+	filename  string
+}
+
+// Write writes the data to a tmp file before copying it over to the desired location.
+func (t *FileWriter) Write(token []byte) error {
```
jkl73, line 26: This write function is specific to write a token file, but the file name (data_writer) is very generic, might want to change the name

launcher/container_runner.go
```diff
@@ -240,12 +241,24 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
...
+	<-timer.C()
+
 	return &ContainerRunner{
 		container,
 		launchSpec,
 		attestAgent,
 		logger,
 		serialConsole,
+		tokenWriter,
+		timer,
```
jkl73, line 261: I feel like adding a timer in ContainerRunner just for the ease of test might be a little bit too much.

If we have to use a timer, maybe offloading the token writing step to the attestAgent might be better.

launcher/container_runner_test.go
```diff
@@ -408,54 +354,73 @@ func testRetryPolicyWithNTries(t *testing.T, numTries int, expectRefresh bool) {
 		}
 		return nil, errors.New("attest unsuccessful")
 	}
+
+	tokenWriter := models.NewFakeTokenWriter()
+	fakeTimer := models.NewFakeTimer()
 	runner := ContainerRunner{
 		attestAgent: &fakeAttestationAgent{attestFunc: attestFunc},
 		logger:      logging.SimpleLogger(),
+		tokenWriter: tokenWriter,
```
alexmwu, line 363: I think the implementation could just be the function signature
```
fetchAndWriteTokenWithRetry(ctx context.Context,
	retry func() *backoff.ExponentialBackOff, refresh func(ctx context.Context) (time.Duration, error)) error
```

That way you can just pass in a `FakeRefresher` with a `Refresh` and a `GetNextToken` method.

alexmwu (changes requested):
"reduce flakiness" can you explain the flakiness?

In general, I agree that this is quite complex just for enabling a test. We should at least get ride of the FS changes.

PR #560, opened 2025-03-25
Test ITA image

JoshuaKrstic:


launcher/container_runner.go
```diff
@@ -467,56 +481,65 @@ func (r *ContainerRunner) refreshToken(ctx context.Context) (time.Duration, erro
...
+			if !isInitialized {
+				initComplete <- err // Signal to the calling function that the first refresh is done.
+				isInitialized = true
+			}
+		}
+
+		r.logger.Info("token refresher goroutine started")
+		// Start with a timer that fires immediately to get the first token.
+		timer := newTimer(0)
+		defer timer.Stop()
```
yawangwang, line 506: it would be safe to add `defer close(initComplete)` to signal the receiver of this channel when the goroutine completes.

launcher/container_runner.go
```diff
@@ -581,33 +637,33 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
+		// Create the well known token file location.
+		if err := os.MkdirAll(launcherfile.HostTmpPath, 0755); err != nil {
+			return err
 		}
+		r.logger.Info("Created directory", "path", launcherfile.HostTmpPath)
 
-		attestClients.ITA = itaClient
-	} else {
-		gcaClient, err := util.NewRESTClient(ctx, r.launchSpec.AttestationServiceAddr, r.launchSpec.ProjectID, r.launchSpec.Region)
+		errchan := r.startTokenRefresher(ctx, defaultRetryPolicy, clock.NewRealTimer, wellKnownFileLocationWriter{})
```
yawangwang: this achieves the same performance efficiency as `r.fetchAndWriteToken`. 
Before refactor, `r.fetchAndWriteToken` will hang there till an initialization error returns.
After refactor, `r.startTokenRefresher` will return immediately with the channel, but `err := <-errchan` will hang there till an initialization error is stuffed into that channel.

Refactor seems to make code more complicated as well, so what other benefits do you think it brings?

JoshuaKrstic:
> Can we move the timer refactoring and container runner testing out of this PR? It makes this harder to review and is not related to ITA (we don't refresh the ITA token).
> 
> I would also like to see a Cloud Build image test that shows us hitting ITA if possible. It can be an error, but enough to see we are able to hit their service on a new image creation

Yeah absolutely, good call. Ill split it out now.

launcher/container_runner.go
```diff
@@ -581,33 +637,33 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
+		err := <-errchan
```
yawangwang, line 660: this achieves the same performance efficiency as `r.fetchAndWriteToken`. 
Before refactor, `r.fetchAndWriteToken` will hang there till an initialization error returns.
After refactor, `r.startTokenRefresher` will return immediately with the channel, but `err := <-errchan` will hang there till an initialization error is stuffed into that channel.

Refactor seems to make code more complicated as well, so what other benefits do you think it brings?

verifier/ita/client.go
```diff
@@ -32,7 +32,7 @@ var regionalURLs map[string]string = map[string]string{
 	"EU": "https://api.eu.trustauthority.intel.com",
 }
 
-type client struct {
+type itaClient struct {
```
yawangwang: nit: since the package name is `ita`, it would be fine to keep the original naming to avoid repetition.

launcher/container_runner.go
JoshuaKrstic: Good point

launcher/container_runner.go
JoshuaKrstic: I'm surprised to hear you think this code is more complicated. I think having tools to not write files to disk and to manage time deterministically is less complicated. To answer your question, I think there are several reasons that this refactor is better than what we had before:


1. Token writer
    - Tests no longer rely on the filesystem, making testing + reasoning about the code easier. 
    - Before, we had an entire test/test cases to make sure that the file at the known endpoint changes when a sync happens. This is complicated to test.
    - with my change to the writer, we are able to see if the mock writer was requested to write a token, a very direct and easily testable signal. We can even get the exact data that we intended to write. 
2. Timer logic
    - We no longer need to wait for things to happen, we can trigger the timer on demand. Timing in tests is difficult to manage and results in sleeps or race conditions. With a test timer we have neither
3. Code flow
    - ContainerRunner now calls RefreshToken exactly once. In the other implementation we had two calls to RefreshToken
    - fetchAndWriteTokenWithRetry was doing too many things - it created a directory in the file system, called the GCA through refreshToken, and then started a goroutine to keep doing that. Now the function only starts a goroutine. We delegate the setup of the filesystem to the calling function. 
    - Since we now have a direct signal into the goroutine, we can confirm it starts correctly. Before, we only were able to confirm that we started it since we always returned nil after starting the goroutine.
4. Functionality choices
    - you're right that the functionality is the same, and it should be for backwards compatibility reasons. We do commit to writing a token to the filesystem before the workload container starts.
    - With my change, we now have a choice to NOT do this (for instance, when we fix the issue where the refresher stops and never comes back if we exceed the max number of retries)

Let me know if you have thoughts on this. I am very open to improving my solution (like adding the deferred channel close as you suggested)

yawangwang: Apologies for the confusion. I'm good with your refactors for items 1, 2, and 4.

For item 3, I agree with your first two points. Regarding the third point, since create a goroutine in Golang is a non-blocking, async operation, the call itself completes almost instantly. As a result, the previous pattern will return nil immediately after the first refreshToken invocation.

Your change by adding `err := <-errchan` is a sync operation, which basically waits for the first refreshToken invocation to complete. To me the code flows seems the same as the previous pattern.

My advice is to keep the previous pattern, as the value of using a channel here seems limited.

JoshuaKrstic: Thanks for clarifying, I'm glad you're on board with most things!

I'm not sure I know what you mean by "previous pattern" in the context of item 3 point 3
Do you suggest I return an error directly from the startup method? So instead of return errchan I do return <-errchan?
I did try this and the issue is testing becomes irritating (and removes choice from the caller) since we need to we need to see if an error has completed, but we also need to orchestrate the behavior in the test. I couldn't figure that out and took that as a signal that I shouldn't do it.

yawangwang: Discussed offline, I'm now onboard with your approach.

yawangwang (approved):
Contingent approval on ITA-related changes. It would be safer to have cloud build image tests against ITA in place (as Alex suggested https://github.com/google/go-tpm-tools/pull/560#issuecomment-3408561134) before merging in.

alexmwu:
Closing in favor of #605

PR #561, opened 2025-03-26
Add eventlog parsing logic for GPU device CC Mode event types

meetrajvala:
#### Changes ####

- Adds the eventlog parsing logic for GPU device state. we will have `cc_mode` field under GPU device state which would indicate whether Confidential Computing is enabled or not for the GPU. Default value of `cc_mode` is set to `UNSET` (first value of proto enum) and that would also be the value for the images which doesn't support GPU driver installation.  

#### Why ? ####

- GCA will use this parsing logic while verifying the attestation and including GPU specific claims in OIDC token.
 
#### Testing ####

- Updated relevant unit tests and test run was successful.


yawangwang (approved):

proto/attest.proto
```diff
@@ -191,11 +191,28 @@ message HealthMonitoringState {
...
 
+enum GPUDeviceCCMode {
+  UNSET = 0;
+  // The NVIDIA H100 hardware, firmware, and software have fully activated all the confidential computing features.
+  ON = 1;
+  // None of the confidential computing-specific features are active.
+  OFF = 2;
+  // The GPU is in a partial CC mode that will match the workflows of CC mode ON, but with security protections disabled.
+  DEVTOOLS = 3;
+}
```
alexmwu, line 202: Please add a reference for these values

alexmwu (approved):

PR #565, opened 2025-04-10
Add GPU driver installer digest verification as part of confidential GPU driver installation flow

meetrajvala:
**Changes:**

- Adds `cos_gpu_installer` image digest verification check before launching the installer container. For image reference, it now refers to the file stored in OEM partition instead of finding the image reference at runtime.
- Changes in `preload.sh` file to add the `cos_gpu_installer` image digest and image reference files under OEM partition. Here manifests API is used to get the image digest (for given image reference) because we do not have `docker` and `gcloud` available in the build container.
- Update the experiment flag to new one (named `EnableConfidentialGPUSupport`) as `EnableGpuDriverInstallation` (existing flag) is being used by trusted space and it would be good to maintain these under different feature flags.
- Update existing image tests to be more specific for confidential GPU.

**Testing:**

- Manual testing
- Unit tests for relevant helper methods

**Changes which would be part of follow-up PR:**

- Measure GPU CC mode status.
- Add driver hash file under OEM partition as part of image build process and use it to verify the installed drivers at runtime.
- Update/Add relevant image tests


yawangwang (commented):
Could you also add what changes you plan to make in the PR description?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -79,6 +90,11 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
 		return fmt.Errorf("failed to pull installer image: %v", err)
 	}
 
+	installerDigest := image.Target().Digest.String()
+	if err := verifyInstallerImageDigest(installerDigest); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
meetrajvala: > Consolidate them into one function with image as input.

Consolidated the `installerDigest` in `verifyInstallerImageDigest` function.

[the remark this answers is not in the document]
meetrajvala: > Also I'd recommend creating a new function `verifyCGPUDriverAttestation` consisting of multiple runtime verification steps according to the DD. `verifyInstallerImageDigest` is just step#1 of of the verification process.

I have not added it under one `verifyCGPUDriverAttestation` function because installer digest verification is pre-installation check but driver digest verification and driver installation verification are post installation checks.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -208,20 +238,42 @@ func setGPUStateToReady() error {
...
+	ccMode, err := GetGPUCCMode()
+	if err != nil {
+		return false, err
+	}
+	return ccMode == CCModeON, nil
+}
+
+// GetGPUCCMode executes nvidia-smi to determine the current Confidential Computing (CC) mode status of the GPU.
+// It returns the CC mode ("ON" or "OFF") and an error if the command fails or if the output cannot be parsed.
+func GetGPUCCMode() (CCMode, error) {
```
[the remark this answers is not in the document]
meetrajvala: I have updated it to `QueryCCMode`. If I drop the `Get` prefix, it would give an lint warning because of other packages would call it gpu.GPUCCMode() and I can not keep just `CCMode` as it would clash with enum.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -18,19 +19,29 @@ import (
...
+var supportedCGPUTypes = []deviceinfo.GPUType{
 	deviceinfo.H100,
 }
 
+// CCMode represents the status confidential computing mode of the GPU.
+type CCMode string
+
+func (ccm CCMode) isValid() error {
+	switch ccm {
+	case CCModeOFF, CCModeON:
```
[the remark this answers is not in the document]
meetrajvala: I just updated this PR with DEVTOOLS mode specific changes as well. now we do not have to make it in follow-up PR (of measurement related changes).

launcher/image/preload.sh
```diff
@@ -100,6 +103,41 @@ configure_systemd_units_for_hardened() {
...
+set_gpu_driver_ref_values() {
+  local cos_gpu_installer_image_ref
+  local cos_gpu_installer_image_digest
+
+  mkdir ${GPU_REF_VALUES_PATH}
+  cos_gpu_installer_image_ref=$(cos-extensions list -- --gpu-installer)
+  cos_gpu_installer_image_digest=$(get_cos_gpu_installer_image_digest ${cos_gpu_installer_image_ref})
+  
+  echo ${cos_gpu_installer_image_ref} >> ${COS_GPU_INSTALLER_IMAGE_REF}
+  echo ${cos_gpu_installer_image_digest} >> ${COS_GPU_INSTALLER_IMAGE_DIGEST}
```
jkl73, line 138: Just be extra careful, make sure the files are not empty and make the digest has the expected length

jkl73 (approved):

jkl73:
Please wait for lgtm from all reviewers to submit

launcher/internal/gpu/driverinstaller.go
```diff
@@ -162,21 +178,36 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
 		if err = setGPUStateToReady(); err != nil {
 			return fmt.Errorf("failed to set the GPU state to ready: %v", err)
 		}
+	} else {
+		return fmt.Errorf("confidential compute is not enabled for the gpu type %s", gpuType)
```
yawangwang, line 182: This is the fail-close pattern: either cc_mode is OFF or DEVTOOLS will lead to launcher exit. My understanding is launcher should measure whatever value of the CC_mode, send them to GCA, and let the relying party to decide what CC_mode is acceptable in their appraisal policy.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -208,20 +239,67 @@ func setGPUStateToReady() error {
...
+	ccMode, err := QueryCCMode()
+	if err != nil {
+		return false, err
+	}
+	return ccMode == CCModeON, nil
+}
+
+// QueryCCMode executes nvidia-smi to determine the current Confidential Computing (CC) mode status of the GPU.
+// If DEVTOOLS mode is enabled, it would override CC mode as DEVTOOLS. DEVTOOLS mode would be enabled only when CC mode is ON.
+func QueryCCMode() (CCMode, error) {
```
yawangwang, line 252: Could you add unit tests for this method as it's now growing more complicated. Also it would be a good practice to add unit tests for other helper methods.

launcher/internal/gpu/driverinstaller.go
meetrajvala: I added this lines while removing the non confidential GPU related tests (which were added for trusted space) but I didn't realize that we need to measure the cc_mode even in the case of OFF or DEVTOOLS for reflecting it into the token. Thanks for catching it !

launcher/image/preload.sh
meetrajvala: Added the digest length check for `cos_gpu_installer_image_digest` and the empty value checks for `cos_gpu_installer_image_ref` and `cos_gpu_installer_image_digest`.

launcher/internal/gpu/driverinstaller.go
meetrajvala: I was planning to add the unit tests for these methods but we decided that because most of the helper methods use `nvidia-smi` commands under the hood, it would be good to cover it via integration tests instead of mocking most of the codelines (related to containerd and nvidia-smi output). For other helper functions (which do not rely on `nvidia-smi` or containerd), I have added the unit tests.

launcher/internal/gpu/driverinstaller_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.filePath != "/tmp/non_existent_file.txt" {
+				err := os.WriteFile(tt.filePath, []byte(tt.fileContent), 0644)
+				if err != nil {
+					t.Errorf("failed to write to the testfile %s", tt.filePath)
+				}
+				defer os.Remove(tt.filePath)
```
yawangwang, line 57: You can use `t.TempDir` https://pkg.go.dev/testing#B.TempDir to create a temporary directory for testing purpose. And the temp dir in return will get auto-deleted after tests complete.

launcher/internal/gpu/driverinstaller.go
yawangwang: You seemed to combine the original `parseCCMode` back to this method. Why?

Ofc integration tests can cover `nvidia-smi` usages, but unit tests are faster and can help early detect any issues. Plus, mocking the result of `nvidia-smi` command shouldn't take too much work, all it takes is change the method signature a bit.

e.g.,
```
type nvidiaSmiCmdOutput func() ([]byte, error)

func QueryCCMode(fn nvidiaSmiCmdOutput) (CCMode, error) {
    output, err := fn()
    // your implementations...
}
```

And the caller can invoke `QueryCCMode` this way:
```
ccMode, err := QueryCCMode(func() ([]byte, error) { return nvidiaSmiCmd.Output() })
if err != nil {
...
}
```
This will make writing unit tests much easier, WDYT?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -15,22 +15,30 @@ import (
...
 )
 
-var supportedGpuTypes = []deviceinfo.GPUType{
+var supportedGPUTypes = []deviceinfo.GPUType{
 	deviceinfo.L4,
 	deviceinfo.T4,
 	deviceinfo.A100_40GB,
 	deviceinfo.A100_80GB,
 	deviceinfo.H100,
 }
```
yawangwang, line 37: I thought we would only support confidential GPU type, why now supporting other non-confidential GPU types?

launcher/image/preload.sh
```diff
@@ -100,6 +103,57 @@ configure_systemd_units_for_hardened() {
...
 
+get_cos_gpu_installer_image_digest() {
+  local image_ref="${1}"
+  local registry
+  local repo_with_image_name
+  local tag
+  local manifest_url
+  local image_digest
+
+  if [[ "$image_ref" =~ ^([^/]+)/([^:]+):([^:]+)$ ]]; then
```
alexmwu, line 114: This regex seems like it's a bit overly broad. What about matching the host?

launcher/internal/experiments/experiments.go
```diff
@@ -11,9 +11,10 @@ import (
 // Failure to unmarshal the experiment JSON data will result in an empty object being returned
 // to treat experiment flags as their default value. The error should still be checked.
 type Experiments struct {
-	EnableTestFeatureForImage   bool
-	EnableTempFSMount           bool
-	EnableGpuDriverInstallation bool
+	EnableTestFeatureForImage    bool
+	EnableTempFSMount            bool
```
alexmwu, line 15: Please remove this experiment.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -168,15 +180,31 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
...
-func getInstallerImageReference() (string, error) {
-	installerImageRefBytes, err := exec.Command("cos-extensions", "list", "--", "--gpu-installer").Output()
+func getInstallerImageReference(installerImageRefFile string) (string, error) {
+	imageRefBytes, err := os.ReadFile(installerImageRefFile)
 	if err != nil {
 		return "", fmt.Errorf("failed to get the cos-gpu-installer version: %v", err)
 	}
-	installerImageRef := strings.TrimSpace(string(installerImageRefBytes))
+	installerImageRef := strings.TrimSpace(string(imageRefBytes))
+	if len(installerImageRef) < 1 {
```
alexmwu, line 189: == 0

launcher/internal/gpu/driverinstaller.go
```diff
@@ -168,15 +180,31 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
...
+	installerDigest := image.Target().Digest.String()
+	imageDigestBytes, err := os.ReadFile(referenceDigestFile)
+	if err != nil {
+		return fmt.Errorf("failed to get the cos-gpu-installer image digest: %v", err)
+	}
+	expectedInstallerDigest := strings.TrimSpace(string(imageDigestBytes))
+	if installerDigest == expectedInstallerDigest {
+		return nil
+	}
+	return fmt.Errorf("cos_gpu_installer image digest verification failed - expected : %s, actual : %s", expectedInstallerDigest, installerDigest)
```
alexmwu, line 205: nit: it's more typical to invert the error check (installerDigest != expectedInstallerDigest) and return nil at the end.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -15,22 +15,30 @@ import (
...
 
+// CCMode enums
 const (
-	installerContainerID = "tee-gpu-driver-installer-container"
-	installerSnapshotID  = "tee-gpu-driver-installer-snapshot"
+	CCModeON             CCMode = "ON"
+	CCModeOFF            CCMode = "OFF"
+	CCModeDevTools       CCMode = "DEVTOOLS"
+	installerContainerID        = "tee-gpu-driver-installer-container"
+	installerSnapshotID         = "tee-gpu-driver-installer-snapshot"
```
alexmwu, line 28: this should be in a separate const block

launcher/internal/gpu/driverinstaller.go
```diff
@@ -15,22 +15,30 @@ import (
 	"github.com/containerd/containerd/oci"
 	"github.com/google/go-tpm-tools/launcher/internal/logging"
 	"github.com/google/go-tpm-tools/launcher/spec"
+	"github.com/google/logger"
 	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
+// CCMode enums
```
alexmwu, line 22: Why aren't you using this enum? https://github.com/google/go-tpm-tools/blob/cs_cgpu_h100/proto/attest.proto#L195

launcher/internal/gpu/driverinstaller.go
```diff
@@ -15,22 +15,30 @@ import (
...
 const (
-	installerContainerID = "tee-gpu-driver-installer-container"
-	installerSnapshotID  = "tee-gpu-driver-installer-snapshot"
+	CCModeON             CCMode = "ON"
```
alexmwu, line 24: If you decide not to use the enum above, it's generally good practice to ahve an unspecified value. You return an empty string below, but should return the unspecified enum val.

launcher/internal/gpu/driverinstaller.go
alexmwu: +1

launcher/internal/gpu/driverinstaller.go
```diff
@@ -208,20 +236,55 @@ func setGPUStateToReady() error {
...
+		return "", err
+	}
+
 	// The nvidia-smi conf-compute command fails for GPU which doesn't support confidential computing.
 	// This check would bypass nvidia-smi conf-compute command for GPU not having confidential compute support.
 	if strings.Contains(string(ccModeOutput), "No CC capable devices found") {
-		logger.Info(fmt.Sprintf("Confidential Computing is not supported for GPU type : %s", gpuType.String()))
-		return false, nil
+		logger.Info("Confidential Computing is not supported for GPU")
+		return ccMode, nil
```
alexmwu, line 262: Should there be an UNSUPPORTED value?

launcher/internal/experiments/experiments.go
meetrajvala: Removing this would require to remove its usage references. I am planning to address this in follow-up PR which updates this branch with main branch.

launcher/internal/gpu/driverinstaller.go
meetrajvala: Ah, thanks for pointing out. I can use this one.

launcher/internal/gpu/driverinstaller.go
meetrajvala: This is addressed as I updated it to use [this enum](https://github.com/google/go-tpm-tools/blob/cs_cgpu_h100/proto/attest.proto#L195).

launcher/internal/gpu/driverinstaller.go
meetrajvala: As discussed, I updated to only supported confidential GPU list for which kernel modules are open sourced.

launcher/image/preload.sh
meetrajvala: Yes, it is bit broader. Reason I didn't match host or other fields is for flexibility against the installer reference updates (e.g. if cos updates moves the installer image to some other registry).

launcher/internal/gpu/driverinstaller.go
meetrajvala: I have updated it to fail early for non confidential GPUs and so this would be no longer required. Additionally, I updated it to use [this enum](https://github.com/google/go-tpm-tools/blob/cs_cgpu_h100/proto/attest.proto#L195) which would provide UNSET status for such case.

launcher/internal/gpu/driverinstaller.go
meetrajvala: > You seemed to combine the original `parseCCMode` back to this method. Why?

That's right. The previous `parseCCMode` function was primarily using regex to extract the CC mode based on the 'ON' or 'OFF' values from the nvidia-smi output. I have since updated the logic to use strings.Contains() directly within the QueryCCMode function. This simplifies the code and removes the need for a separate parsing function.

launcher/internal/gpu/driverinstaller_test.go
```diff
@@ -0,0 +1,282 @@
...
+					t.Fatalf("failed to write to the testfile %s: %v", filePath, err)
+				}
+
+			}
+
+			ref, err := getInstallerImageReference(filePath)
+
+			if (err != nil) != tt.wantErr {
+				t.Errorf("getInstallerImageReference() error = %v, wantErr %v", err, tt.wantErr)
+				return
```
yawangwang, line 66: Golang tests should generally not abort at the first encountered problem. go/go-style/best-practices#t-fatal

Remove `return` and please apply to everywhere else in the test.

launcher/internal/gpu/driverinstaller.go
```diff
@@ -190,38 +216,45 @@ func remountAsExecutable(dir string) error {
...
-	nvidiaSmiCmd := fmt.Sprintf("%s/bin/nvidia-smi", InstallationHostDir)
-	ccModeOutput, err := exec.Command(nvidiaSmiCmd, "conf-compute", "-f").Output()
-	// The nvidia-smi conf-compute command fails for GPU which doesn't support confidential computing.
-	// This check would bypass nvidia-smi conf-compute command for GPU not having confidential compute support.
-	if strings.Contains(string(ccModeOutput), "No CC capable devices found") {
-		logger.Info(fmt.Sprintf("Confidential Computing is not supported for GPU type : %s", gpuType.String()))
-		return false, nil
+// QueryCCMode executes nvidia-smi to determine the current Confidential Computing (CC) mode status of the GPU.
+// If DEVTOOLS mode is enabled, it would override CC mode as DEVTOOLS. DEVTOOLS mode would be enabled only when CC mode is ON.
+func QueryCCMode(nvidiaSmiCCModeCmd, nvidiaSmiDevToolsCmd nvidiaSmiCmdOutput) (attest.GPUDeviceCCMode, error) {
```
yawangwang, line 235: nit: drop `nvidiaSmi` prefix to avoid naming repetitions. `ccModeCmd`, `devToolsCmd` should be fine. This applies everywhere else in this file.

yawangwang (approved):
LGTM contingent on successful cloud build tests

PR #567, opened 2025-04-23
Add driver digest verification in confidential GPU driver installation flow

meetrajvala:

- Adds the verification of driver (nvidia .run file) digest in driver installation flow.
- Reads the run file digest from well known public GCS bucket and stores it under OEM partition as reference driver digest.
- Required utility functions for parsing and hash calculation.  

**Testing:**

- Unit tests for newly added functions

**Changes which would be part of follow-up PR:**

- Measure GPU CC mode status.

### DIFF
--- a/launcher/image/preload.sh
+++ b/launcher/image/preload.sh
@@ -6,8 +6,8 @@ readonly EXPERIMENTS_BINARY="confidential_space_experiments"
 readonly GPU_REF_VALUES_PATH="${CS_PATH}/gpu"
 readonly COS_GPU_INSTALLER_IMAGE_REF="${GPU_REF_VALUES_PATH}/cos_gpu_installer_image_ref"
 readonly COS_GPU_INSTALLER_IMAGE_DIGEST="${GPU_REF_VALUES_PATH}/cos_gpu_installer_image_digest"
-readonly DRIVER_DIGEST_SHA256SUM="${GPU_REF_VALUES_PATH}/driver_digest_sha256sum"
-readonly DRIVERS_GCS_BUCKET="cos-nvidia-gpu-drivers"
+readonly DRIVER_DIGEST="${GPU_REF_VALUES_PATH}/driver_digest"
+readonly DRIVER_GCS_BUCKET="cos-nvidia-gpu-drivers"
 
 copy_launcher() {
   cp launcher "${CS_PATH}/cs_container_launcher"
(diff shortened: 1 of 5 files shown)


launcher/image/preload.sh
```diff
@@ -126,6 +128,26 @@ get_cos_gpu_installer_image_digest() {
   echo "${image_digest}"
 }
 
+set_reference_driver_digest() {
```
yawangwang, line 131: nit: set_driver_digest_reference

launcher/image/preload.sh
```diff
@@ -6,6 +6,8 @@ readonly EXPERIMENTS_BINARY="confidential_space_experiments"
 readonly GPU_REF_VALUES_PATH="${CS_PATH}/gpu"
 readonly COS_GPU_INSTALLER_IMAGE_REF="${GPU_REF_VALUES_PATH}/cos_gpu_installer_image_ref"
 readonly COS_GPU_INSTALLER_IMAGE_DIGEST="${GPU_REF_VALUES_PATH}/cos_gpu_installer_image_digest"
+readonly DRIVER_DIGEST_SHA256SUM="${GPU_REF_VALUES_PATH}/driver_digest_sha256sum"
+readonly DRIVERS_GCS_BUCKET="cos-nvidia-gpu-drivers"
```
yawangwang, line 10: nit: `DRIVER_GCS_BUCKET`

launcher/image/preload.sh
```diff
@@ -126,6 +128,26 @@ get_cos_gpu_installer_image_digest() {
...
+    echo "Error: failed to download the driver digest file from ${driver_digest_gcs_url}." >&2
+    return 1
+  fi
+  
+  driver_digest=$(cat ${DRIVER_DIGEST_SHA256SUM} | cut -d " " -f 1)
+  # Check for the expected length of the SHA256 digest (64 hex characters)
+  if [ ${#driver_digest} -ne 64 ]; then
+    echo "Error: driver digest has an unexpected length: ${#driver_digest}, Expected 64." >&2
+    return 1
+  fi
```
yawangwang, line 148: This only checks the string length, could you also check if this is a hex string?

launcher/image/preload.sh
```diff
@@ -145,14 +177,11 @@ set_gpu_driver_ref_values() {
   fi
```
patilatul-goog, line 177: Not related to the current PR.
But it seems like the condition in this existing method is slightly incorrect.

On line [174](https://github.com/google/go-tpm-tools/pull/567/files#diff-89b06e3d3b793a67b04d03492e3c633007a92f9e7c9c9a5a8cc09b5688957f07R174), looks like it was intended to check for _digest_ but it re-checks the _ref_

We can handle this in a separate PR since it isn't related to your change

patilatul-goog (approved):

launcher/image/preload.sh
meetrajvala: Thanks @patilatul-goog for pointing this. Updated it as part of this PR itself as it was small fix.

launcher/image/preload.sh
```diff
@@ -6,6 +6,8 @@ readonly EXPERIMENTS_BINARY="confidential_space_experiments"
 readonly GPU_REF_VALUES_PATH="${CS_PATH}/gpu"
 readonly COS_GPU_INSTALLER_IMAGE_REF="${GPU_REF_VALUES_PATH}/cos_gpu_installer_image_ref"
 readonly COS_GPU_INSTALLER_IMAGE_DIGEST="${GPU_REF_VALUES_PATH}/cos_gpu_installer_image_digest"
+readonly DRIVER_DIGEST_SHA256SUM="${GPU_REF_VALUES_PATH}/driver_digest_sha256sum"
```
alexmwu, line 9: nit: `DRIVER_DIGEST_SHA256SUM`/
for consistency, either make the above `COS_GPU_INSTALLER_IMAGE_DIGEST_SHA256` or the below `DRIVER_DIGEST`

launcher/image/preload.sh
```diff
@@ -126,6 +128,36 @@ get_cos_gpu_installer_image_digest() {
...
+validate_digest() {
+  driver_digest="${1}"
+  # Check for the expected length of the SHA256 digest (64 hex characters)
+  if [ ${#driver_digest} -ne 64 ]; then
+    echo "Error: driver digest has an unexpected length: ${#driver_digest}, Expected 64." >&2
+    return 1
+  fi
+  # Check for valid hexadecimal string
+  if [[ ! ${driver_digest} =~ ^[0-9a-fA-F]+$ ]]; then
+    return "Error: driver digest ${driver_digest} is not valid hexadecimal string." >&2
```
alexmwu, line 140: ...is not a valid...

launcher/image/preload.sh
```diff
@@ -126,6 +128,36 @@ get_cos_gpu_installer_image_digest() {
   echo "${image_digest}"
 }
 
+validate_digest() {
```
alexmwu, line 131: nit: validate_sha256_hex

launcher/image/preload.sh
```diff
@@ -126,6 +128,36 @@ get_cos_gpu_installer_image_digest() {
...
+    return 1
+  fi
+}
+
+set_driver_digest_reference() {
+  local driver_version
+  local driver_digest_gcs_url
+
+  # Fetching the default driver version for H100 GPU.
+  driver_version=$(cos-extensions list -- --target-gpu NVIDIA_H100_80GB | grep DEFAULT | cut -d" " -f 1)
```
alexmwu, line 150: this func should take a parameter for the target gpu rather than hardcoding the H100

launcher/image/preload.sh
```diff
@@ -126,6 +128,36 @@ get_cos_gpu_installer_image_digest() {
...
+    return 1
+  fi
+  # Check for valid hexadecimal string
+  if [[ ! ${driver_digest} =~ ^[0-9a-fA-F]+$ ]]; then
+    return "Error: driver digest ${driver_digest} is not valid hexadecimal string." >&2
+    return 1
+  fi
+}
+
+set_driver_digest_reference() {
```
alexmwu, line 145: nit: store_driver_digest or something similar

launcher/image/preload.sh
alexmwu: Same with the file names

alexmwu, line 9: Should we store by driver version? For example, `${GPU_REF_VALUES_PATH}/${driver_version}/driver_digest_sha256sum`

launcher/internal/gpu/driverinstaller.go
```diff
@@ -203,6 +214,17 @@ func verifyInstallerImageDigest(image containerd.Image, referenceDigestFile stri
 	return nil
 }
 
+func verifyDriverDigest(driverFile, referenceHash string) error {
+	calculatedHash, err := calculateSHA256Hash(driverFile)
+	if err != nil {
+		return err
+	}
+	if calculatedHash != referenceHash {
+		return fmt.Errorf("gpu driver digest verification failed - expected : %s, got : %s", referenceHash, calculatedHash)
```
alexmwu, line 223: nit: GPU

alexmwu (approved):

launcher/image/preload.sh
meetrajvala: Currently, we have only default driver version support (for H100 gpu only) so it might just one more directory in path. I can accomodate this when we extend support for multiple cGPU families requiring different driver versions (There is a tracking ticket for this).

PR #569, opened 2025-05-02
Measure GPU CC mode 

meetrajvala:

- Adds the function for measuring the GPU CC mode and uses it in `container_runner`.
- Moves the GPU driver installation steps later in the `startLauncher` function.
- Added check for scenario when GPU is attached but required metadata flag is not passed for driver installation.
 
**Testing:**

- Unit tests for newly added function


launcher/launcher/main.go
```diff
@@ -189,20 +190,6 @@ func startLauncher(launchSpec spec.LaunchSpec, serialConsole *os.File) error {
...
-			err = installer.InstallGPUDrivers(ctx)
-			if err != nil {
-				return fmt.Errorf("failed to install gpu drivers: %v", err)
-			}
-		} else {
-			logger.Info("Confidential GPU support experiment flag is not enabled for this project. Ensure that it is enabled when tee-install-gpu-driver is set to true")
-			return fmt.Errorf("confidential gpu support experiment flag is not enabled")
-		}
-	}
-
```
yawangwang, line 205: Any reason why moving this block around?

launcher/container_runner_test.go
```diff
@@ -638,6 +639,77 @@ func TestMeasureCELEvents(t *testing.T) {
...
+
+	tests := []struct {
+		name            string
+		mockCCModeCmd   gpu.NvidiaSmiCmdOutput
+		mockDevToolsCmd gpu.NvidiaSmiCmdOutput
+		wantErr         bool
+		wantEvents      []cel.CosType
+	}{
+		{
+			name:            "Successful GPU CC mode measurement",
```
yawangwang, line 664: nit: lowercase of the initial letter for test case name.

launcher/container_runner_test.go
```diff
@@ -638,6 +639,77 @@ func TestMeasureCELEvents(t *testing.T) {
...
+			mockCCModeCmd:   func() ([]byte, error) { return []byte("CC status: ON"), nil },
+			mockDevToolsCmd: func() ([]byte, error) { return []byte("DevTools Mode: OFF"), nil },
+			wantErr:         false,
+			wantEvents: []cel.CosType{
+				cel.GpuCCModeType,
+			},
+		},
+		{
+			name:            "Failed GPU CC mode measurement",
```
yawangwang, line 673: please briefly describe what caused a failure test case in the `name`

launcher/container_runner.go
```diff
@@ -341,6 +341,14 @@ func (r *ContainerRunner) measureCELEvents(ctx context.Context) error {
 		return fmt.Errorf("failed to measure memory monitoring state: %v", err)
 	}
 
+	if r.launchSpec.Experiments.EnableConfidentialGPUSupport && r.launchSpec.InstallGpuDriver {
+		ccModeCmd := gpu.NvidiaSmiOutputFunc("conf-compute", "-f")
+		devToolsCmd := gpu.NvidiaSmiOutputFunc("conf-compute", "-d")
+		if err := r.measureGPUCCMode(ccModeCmd, devToolsCmd); err != nil {
+			return fmt.Errorf("failed to measure GPU CC mode status: %v", err)
+		}
+	}
```
yawangwang, line 350: Please add corresponding tests cases to `TestMeasureCELEvents`

launcher/launcher/main.go
```diff
@@ -245,6 +232,26 @@ func startLauncher(launchSpec spec.LaunchSpec, serialConsole *os.File) error {
...
+		} else {
+			logger.Error("Confidential GPU support experiment flag is not enabled for this project. Ensure that it is enabled when tee-install-gpu-driver is set to true")
+			return fmt.Errorf("confidential gpu support experiment flag is not enabled")
+		}
+	} else {
+		deviceInfo, _ := deviceinfo.GetGPUTypeInfo()
+		if deviceInfo != deviceinfo.NO_GPU {
+			logger.Error("GPU is attached, tee-install-gpu-driver is not set")
+			return fmt.Errorf("tee-install-gpu-driver is expected to set to true when GPU is attached")
+		}
```
yawangwang, line 252: Did you check with Rene on this requirement? This would be a breaking change to CS on H100 because a regular workload can't even get run within with a GPU device attached.

yawangwang (changes requested):

launcher/launcher/main.go
meetrajvala: For trusted space, there was a comment to move this block : https://github.com/google/go-tpm-tools/pull/497#discussion_r1797346566. Followed the same here as it is at the beginning of `startLauncher`

launcher/launcher/main.go
meetrajvala: Agreed. but without this check, if a GPU is present but the driver installation flag is missing, workloads relying on the GPU would likely fail. The nature of that failure would depend on how the workload manages its GPU dependency – it might fall back to CPU usage or fail or terminate entirely. This check ensures predictable behavior when a GPU is attached but installation flag is not set. I assumed that attaching a GPU implies its intended use, thus requiring the driver installation flag to guarantee necessary drivers are present. I can confirm this with Rene.

launcher/image/test/test_gpu_driver_installation_cloudbuild.yaml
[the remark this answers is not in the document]
meetrajvala: I have added the check for CC mode event measurement in `test_gpu_workload.sh` script which tests the successful execution of GPU based workload.

launcher/container_runner.go
meetrajvala: I have moved this block directly under `Run()` as it is reducing the test coverage of `measureCELEvents` function because this block creates the anonymous functions required for `QueryCCMode()` which makes it hard to test  `measureCELEvents` via unit tests. 

Another option was to add these anonymous functions as argument of `measureCELEvent` but that would not be very clean way as it would introduce more experiment flag checks (
`if r.launchSpec.Experiments.EnableConfidentialGPUSupport && r.launchSpec.InstallGpuDriver {...}`
) during the creation of these `ccModeCmd` and `devToolsCmd` functions (in the caller of `measureCELEvents`) and also in `measureCELEvent` function where we will call `measureGPUCCMode`.

launcher/launcher/main.go
meetrajvala: > When would this happen? 

When GPU is attached to the VM but customer didn't set the `tee-install-gpu-driver` flag. In this case, workload is expecting an interaction with GPU device but because we didn't install drivers (as flag was not set), GPU device would not available to the workload container and workload execution might fail. 

> Why would we pass through the GPU device if they don't set the flag?

We do not make the GPU device available to the container if this flag is not set. This check was just added to fail deterministically (when GPU is attached but mistakenly or knowingly installation flag is not set).

launcher/container_runner_test.go
```diff
@@ -638,6 +639,77 @@ func TestMeasureCELEvents(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			gotEvents := []cel.CosType{}
+
+			fakeAgent := &fakeAttestationAgent{
+				measureEventFunc: func(content cel.Content) error {
+					got, _ := content.GetTLV()
```
alexmwu, line 687: nit: check the error. you can t.Fatal

launcher/launcher/main.go
alexmwu: > When GPU is attached to the VM but customer didn't set the tee-install-gpu-driver flag. In this case, workload is expecting an interaction with GPU device but because we didn't install drivers (as flag was not set), GPU device would not available to the workload container and workload execution might fail.

I understand, but we could print a warning log instead of stopping launch. Please double check with Rene.

alexmwu (approved):

jkl73 (approved):

launcher/launcher/main.go
meetrajvala: I checked this with Rene and he suggested to fail upfront instead of getting into a condition where workload might not work.

yawangwang (approved):

PR #580, opened 2025-07-15
Attest now will pass in a default audience instead of passing an error

Sibcgh:


launcher/teeserver/tee_server.go
```diff
@@ -149,9 +153,10 @@ func (a *attestHandler) attest(w http.ResponseWriter, r *http.Request, client ve
 		}
 
 		if tokenOptions.Audience == "" {
-			err := fmt.Errorf("use GET request for the default identity token")
-			a.logAndWriteHTTPError(w, http.StatusBadRequest, err)
-			return
+			// err := fmt.Errorf("use GET request for the default identity token")
+			// a.logAndWriteHTTPError(w, http.StatusBadRequest, err)
+			tokenOptions.Audience = audienceSTS
```
jkl73: Instead of adding the logic, we may want to handle this default audience case in agent itself: https://github.com/google/go-tpm-tools/blob/05abd54ea3ed4ee0520d3d1c668b3fb0f754e6d3/launcher/agent/agent.go#L180

launcher/agent/agent.go
```diff
@@ -183,6 +183,15 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 		return nil, err
 	}
 
+	tokenOpts := opts.TokenOptions
+	if tokenOpts == nil {
+		tokenOpts = &models.TokenOptions{}
+	}
+
+	if tokenOpts.Audience == "" {
+		tokenOpts.Audience = "audienceSTS"
```
jkl73: you mean https://sts.googleapis.com ?

launcher/agent/agent.go
```diff
@@ -183,6 +187,15 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 		return nil, err
 	}
 
+	tokenOpts := opts.TokenOptions
+	if tokenOpts == nil {
+		tokenOpts = &models.TokenOptions{}
+	}
+
+	if tokenOpts.Audience == "" {
```
jessieqliu: I think this conditional should be combined with the above.

```
if tokenOpts == nil {
  tokenOpts = &models.TokenOptions{ Audience: audienceSTS }
}
```

Mostly because the GCA requires the customer to provide an audience if they requested custom nonces (it returns an error if the request has custom nonces but not a customer audience). The current code would autopopulate a custom audience which would render that check useless and probably be misleading to customers (if they think only sending custom nonces is fine).

@JoshuaKrstic is most familiar with the requirements here and should confirm what the intended behavior is.

jessieqliu: Upon further reflection, it is possible a customer would provide TokenOptions with only the token type so it might need to be something like:
```
if tokenOpts == nil {
  tokenOpts = &models.TokenOptions{}
}

// The customer is responsible for providing an audience if they provided nonces.
if tokenOpts.Audience == "" && len(tokenOpts.Nonces) == 0 {
  tokenOpts.Audience = audienceSTS
}
```

launcher/teeserver/tee_server.go
```diff
@@ -148,12 +148,6 @@ func (a *attestHandler) attest(w http.ResponseWriter, r *http.Request, client ve
 			return
 		}
 
-		if tokenOptions.Audience == "" {
```
jessieqliu: @JoshuaKrstic Does this align with our expected usage/behavior for the non-default IPC endpoint? My instinct is this PR should focus on the default case.

JoshuaKrstic: Yes, your instinct is right. The default audience should only exist in the GET request, not the post. Its probably best to do this by having the agent fill in default audience and have the POST request reject for no audience being set.

JoshuaKrstic: Hm actually, now that I think about it, we should just have the GET and the auto-refresh pass their own token options to the agent instead of filling in the audience in the agent.

Thinking through how we want this to work:
1) We want the GCA to ultimately control the request + reject if its bad. We don't want to interpret the request in the client
2) The GET request + sync should be "default".  (aka, have sts as the audience with no nonces)

If we override the request in the agent, we get a weird flow where users can POST and then get a "default" token on accident. We'd then have to explain this behavior. 
IMO sending a POST with no audience should result in a 400.
We can accomplish the default audience by sending token options to the agent from the refresher and the GET endpoint

JoshuaKrstic: After thinking about it more, you made the right call overriding it in the agent. Just keep this condition and we should be good to go.

launcher/agent/agent.go
```diff
@@ -183,6 +187,16 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 		return nil, err
 	}
 
+	tokenOpts := opts.TokenOptions
```
jessieqliu, line 190: Can we add a test that checks for the "default" token options behavior being set here?

PR #585, opened 2025-08-20
Add HashNonce flag to Attest and VerifyAttestation

alexmwu:
Currently, key.Attest takes a slice of data for the attestation nonce. This gets passed directly to the TPM. However, verifiers that are not aware of the TPM's max size for extraData can inadvertently send a challenge that is too large.
This change introduces a HashNonce flag that uses the key's signing scheme hash algorithm to first hash the AttestOpts nonce. The server package also introduces the same logic to hash the nonce based on the Attestation message's AkPub signing scheme.
This allows verifiers to send an arbitrary amount of entropy without knowing the max hash size of the TPM.

Note: this may be a breaking change to users that don't populate AKPub within the attestation proto. This was never optional, and it was always populated by the go-tpm-tools client code.

### DIFF
--- a/client/attest.go
+++ b/client/attest.go
@@ -59,6 +59,11 @@ type AttestOpts struct {
 	// depending on the technology's size. Leaving this nil is not recommended. If
 	// nil, then TEEDevice must be nil.
 	TEENonce []byte
+	// HashNonce will apply the attestation key's signing scheme hash algorithm
+	// to the input Nonce field and use the resulting digest in place of the
+	// original Nonce.
+	// Nonce must still be unique and application-specific.
+	HashNonce bool
 }
 
 // SevSnpQuoteProvider encapsulates the SEV-SNP attestation device to add its attestation report
@@ -286,8 +291,16 @@ func (k *Key) Attest(opts AttestOpts) (*pb.Attestation, error) {
 		return nil, fmt.Errorf("failed to encode public area: %w", err)
 	}
 	attestation.AkCert = k.CertDERBytes()
+	extraData := opts.Nonce
+	if opts.HashNonce {
+		var err error
+		extraData, err = internal.HashNonce(k.PublicArea(), extraData)
+		if err != nil {
+			return nil, fmt.Errorf("failed to hash the input nonce: %w", err)
+		}
+	}
 	for _, sel := range sels {
-		quote, err := k.Quote(sel, opts.Nonce)
+		quote, err := k.Quote(sel, extraData)
 		if err != nil {
 			return nil, err
 		}
--- a/server/verify.go
+++ b/server/verify.go
@@ -66,6 +66,10 @@ type VerifyOpts struct {
 	// "Calling EFI Application from Boot Option". This option is useful when
 	// the host platform loads EFI Applications unrelated to OS boot.
 	AllowEFIAppBeforeCallingEvent bool
+	// HashNonce will apply the attestation key's signing scheme hash algorithm
+	// to the input Nonce field and use the resulting digest in place of the
+	// original Nonce.
+	HashNonce bool
 }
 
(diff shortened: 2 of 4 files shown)


server/verify.go
```diff
@@ -175,44 +187,44 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
 
 // validateAK validates AK cert in the attestation, and returns AK cert (if exists) and public key.
 // It also pulls out the GCE Instance Info if it exists.
-func validateAK(attestation *pb.Attestation, opts VerifyOpts) (*pb.MachineState, crypto.PublicKey, error) {
+func validateAK(attestation *pb.Attestation, opts VerifyOpts) (*pb.MachineState, tpm2.Public, crypto.PublicKey, error) {
+	// If the AK Cert is not in the attestation, use the AK Public Area.
```
jkl73, line 174: akpub was optional if akcert presents (even though we attach it anyway), so I don't think we should return error if it cannot get akpub

instead, we can return error when it specifically require HashNonce
```
if err != nil and opts.HashNonce {
    // error
}
```

server/verify.go
```diff
@@ -175,44 +187,44 @@ func VerifyAttestation(attestation *pb.Attestation, opts VerifyOpts) (*pb.Machin
...
-		akPubArea, err := tpm2.DecodePublic(attestation.GetAkPub())
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to decode AK public area: %w", err)
-		}
 		akPubKey, err := akPubArea.Key()
 		if err != nil {
-			return nil, nil, fmt.Errorf("failed to get AK public key: %w", err)
+			return nil, tpm2.Public{}, nil, fmt.Errorf("failed to get AK public key: %w", err)
 		}
 		if err := validateAKPub(akPubKey, opts); err != nil {
```
jkl73, line 184: Also , since we are using the "content" in the ak, we better `validateAKPub` before using its "content", this check is currently skipped if any ak cert presents

client/attest.go
```diff
@@ -287,7 +292,15 @@ func (k *Key) Attest(opts AttestOpts) (*pb.Attestation, error) {
 	}
 	attestation.AkCert = k.CertDERBytes()
 	for _, sel := range sels {
-		quote, err := k.Quote(sel, opts.Nonce)
+		extraData := opts.Nonce
```
jkl73, line 295: this can be outside the loop, I think the nonce is only related to the key, not the pcr bank?

client/attest.go
```diff
@@ -287,7 +292,15 @@ func (k *Key) Attest(opts AttestOpts) (*pb.Attestation, error) {
...
+		if opts.HashNonce {
+			var err error
+			extraData, err = internal.HashNonce(k.PublicArea(), extraData)
+			if err != nil {
+				return nil, fmt.Errorf("failed to hash the input nonce: %w", err)
+			}
+		}
```
jkl73: this can be outside the loop, I think the nonce is only related to the key, not the pcr bank?

server/verify.go
alexmwu: It's always attached in key.Attest, since 2021. And we return an error if we fail to get it.

alexmwu: I've opted to not include this check (for a breaking change) since it would propogate the `if opts.HashNonce` check down to every new new check in this func (decoding, getting crypto key, comparing to cert key) for questionable value.

jkl73: ack

jkl73: we may want to consider remove the optional: https://github.com/google/go-tpm-tools/blob/d2379fa8b41901accaa7017e56731dac97d02987/proto/attest.proto#L35

alexmwu: That field is ak_cert not ak_pub.

PR #586, opened 2025-08-21
Add experiment-gated migration to VerifyConfidentialSpace

jessieqliu:


launcher/agent/agent.go
```diff
@@ -251,16 +264,34 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
...
 	if len(resp.PartialErrs) > 0 {
 		a.logger.Error(fmt.Sprintf("Partial errors from VerifyAttestation: %v", resp.PartialErrs))
 	}
 	return resp.ClaimsToken, nil
 }
 
+func (a *agent) verify(ctx context.Context, req verifier.VerifyAttestationRequest, client verifier.Client, opts AttestAgentOpts) (*verifier.VerifyAttestationResponse, error) {
+	// If not specified in opts, use experiment to determine verify method.
+	method := opts.Method
+	if method == VerifyUnset {
```
JoshuaKrstic: Would this make more sense in the calling code? Looks like this method expects the caller to specify what sort of attestation they want.

launcher/teeserver/tee_server.go
```diff
@@ -133,15 +135,50 @@ func (a *attestHandler) getITAToken(w http.ResponseWriter, r *http.Request) {
 	a.attest(w, r, a.clients.ITA)
 }
 
+func (a *attestHandler) parseVerifyMethod(headers http.Header) agent.VerifyMethod {
+	if headers == nil {
+		a.logger.Info("No headers specified in request.")
+		return agent.VerifyUnset
+	}
+
+	methods, ok := headers[verifyMethodHeader]
```
JoshuaKrstic: Why do we have the workload specify the attestation type? This feels to me like an implementation detail that the caller should be unaware of

launcher/agent/agent.go
JoshuaKrstic: I think its because we have two callers of this method.
Why did we decide to go with a caller-specified switch with a default instead of having the experiment be the source of truth for which endpoint we hit?

JoshuaKrstic (dismissed):

launcher/agent/agent.go
jessieqliu: The method is optional - if not specified it will fall back to the experiment value. The reason for this approach is we need a more granular (and deterministic) way to decide whether VerifyAttestation for VerifyConfidentialSpace is being used for a given request. Otherwise we'll have/risk significant gaps in test coverage.

- We need regression tests that compare VerifyAttestation tokens to VerifyConfidentialSpace, meaning those workloads need to be certain of which API is being hit (the experiment config can change and/or have sync issues) and we aren't comparing two tokens from the same API.
- We need to support both VerifyConfidentialSpace and VerifyAttestation tests for a time, as we move to using VCS for future images but still support old ones that use VA. Once again, these tests need to be certain/concrete in which API is being used regardless of what the experiment value is because it's not necessarily reflective of all customer uses.
- With relying on the experiment sync, there's also the issue that the sync can fail (even in our tests it's been flaky lately) and we don't have a solid way to detect that from the API output (there's not a way to differentiate VCS-issued tokens from VA-issued ones), so we run the risk of an experiment issue causing our tests to be inaccurate without knowing it.

Putting it as a field in the request allows the workload to specify which method it uses when we make the request. This way, one workload can request API's from both (not intended for customer usage, but so we can flip this switch for our testing).

launcher/agent/agent.go
```diff
@@ -42,6 +42,18 @@ const (
 
 type principalIDTokenFetcher func(audience string) ([][]byte, error)
 
+// VerifyMethod represents the possible atrestation verification methods.
```
jkl73: typo: attestation

jkl73 (commented):
While using http header is okay. I wonder if it's easier to just have a second tee server endpoint for verifyCS.

That way the workload can decide which API to use. And we can still use the experiment in getToken call to redirect call to verifyCS.

jkl73 (approved):
lgtm

PR #587, opened 2025-08-22
TEE Server Error code Translation

Sibcgh:
This change addresses an issue where the `/v1/token` endpoint would incorrectly return a `400 Bad Request` for any failure from the underlying attestation agent, even if the error was due to service unavailability (like a `context deadline exceeded` during zone down testing).

The `getToken` handler now inspects the error returned from `attestAgent.AttestWithClient` and maps it to a more appropriate HTTP status code.

Additional test cases have been added to `TestCustomHandleAttestError` to ensure comprehensive coverage of different error types returned by the agent and validate that the correct HTTP status codes are propagated.


launcher/teeserver/tee_server.go
```diff
@@ -204,3 +205,21 @@ func (s *TeeServer) Shutdown(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (a *attestHandler) handleAttestError(w http.ResponseWriter, err error, message string) {
+	st, ok := status.FromError(err)
+	if ok {
+		switch st.Code() {
```
jessieqliu: Should this be more flexible than only Bad request or Internal? See go/http-canonical-mapping

launcher/teeserver/tee_server_test.go
```diff
@@ -330,3 +364,72 @@ func TestCustomTokenDataParsedSuccessfully(t *testing.T) {
...
+
+			ah.getToken(w, req)
+
+			if w.Code != tc.wantStatusCode {
+				t.Errorf("got status code %d, want %d", w.Code, tc.wantStatusCode)
+			}
+
+			_, err := io.ReadAll(w.Result().Body)
+			if err != nil {
+				t.Fatalf("failed to read response body: %v", err)
```
jessieqliu: Errorf. There no point in using Fatal at the end of the test since the test is about to exit anyway.

launcher/teeserver/tee_server.go
Sibcgh: Added in more definitive checks

launcher/teeserver/tee_server.go
```diff
@@ -15,8 +15,21 @@ import (
 	"github.com/google/go-tpm-tools/verifier"
 	"github.com/google/go-tpm-tools/verifier/models"
 	"github.com/google/go-tpm-tools/verifier/util"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
+var clientErrorCodes = map[codes.Code]bool{
```
JoshuaKrstic: It appears that this map is used for existence checking. If so, its best practice to use struct{} as the value and omit the key when you do the lookup.

launcher/teeserver/tee_server.go
```diff
@@ -204,3 +216,19 @@ func (s *TeeServer) Shutdown(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (a *attestHandler) handleAttestError(w http.ResponseWriter, err error, message string) {
+	st, ok := status.FromError(err)
+	if ok {
+		if clientErrorCodes[st.Code()] {
+			// User errors, like invalid arguments. These are mapped to 400 Bad Request.
```
JoshuaKrstic: Nit: Consider "Map user errors to 400 Bad Request" or "User errors are mapped to 400 Bad Request"

launcher/teeserver/tee_server_test.go
```diff
@@ -88,6 +90,38 @@ func TestGetDefaultToken(t *testing.T) {
 	}
 }
 
+func TestGetDefaultTokenServerError(t *testing.T) {
+	// An empty attestHandler is fine for now as it is not being used
```
JoshuaKrstic, line 92: What are you trying to say with this comment?

launcher/teeserver/tee_server_test.go
```diff
@@ -88,6 +90,38 @@ func TestGetDefaultToken(t *testing.T) {
...
+	data, err := io.ReadAll(w.Result().Body)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if w.Code != http.StatusInternalServerError {
+		t.Errorf("got return code: %d, want: %d", w.Code, http.StatusInternalServerError)
+	}
+	expectedError := "failed to retrieve attestation service token: internal server error from agent"
+	if string(data) != expectedError {
```
JoshuaKrstic: can you use cmp.Diff here?

launcher/teeserver/tee_server_test.go
```diff
@@ -330,3 +364,72 @@ func TestCustomTokenDataParsedSuccessfully(t *testing.T) {
...
+			req := httptest.NewRequest(http.MethodPost, "/v1/token", strings.NewReader(body))
+			w := httptest.NewRecorder()
+
+			ah.getToken(w, req)
+
+			if w.Code != tc.wantStatusCode {
+				t.Errorf("got status code %d, want %d", w.Code, tc.wantStatusCode)
+			}
+
+			_, err := io.ReadAll(w.Result().Body)
```
JoshuaKrstic, line 425: Have we set an expectation that the body should be readable for failures?

launcher/teeserver/tee_server_test.go
Sibcgh: autocomplete added that when i was trying to remove todo debug comments

launcher/teeserver/tee_server_test.go
Sibcgh: The other 4 tests use that same line where they read out the body of the error

launcher/teeserver/tee_server.go
Sibcgh: resolved and also changed 500 error comment as well

launcher/teeserver/tee_server_test.go
JoshuaKrstic: Lets remove this comment, I don't see an empty attestHandler unless I'm seriously misudnerstanding something.

JoshuaKrstic (approved):

PR #589, opened 2025-09-02
Add testutils package to make select test helpers publicly available

jessieqliu:
Adds a wrapper package around internal test functions as our need for testing outside of this repo expands.

Since `GetTPM` makes use of a package-level lock in `internal/test`, it is better to wrap the function than move it to a new package (same for any future functions in that package we also want to make available).

### DIFF
--- a/testutil/utils.go
+++ b/testutil/utils.go
@@ -8,6 +8,11 @@ import (
 	"github.com/google/go-tpm-tools/internal/test"
 )
 
+// GetTPM is a cross-platform testing helper function that retrives the
+// appropriate TPM device from the flags passed into "go test".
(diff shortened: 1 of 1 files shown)


testutil/utils.go
```diff
@@ -0,0 +1,18 @@
...
+package testutil
+
+import (
+	"io"
+	"testing"
+
+	"github.com/google/go-tpm-tools/internal/test"
+)
+
+// GetTPM is a cross-platform testing helper function that retrives the
```
alexmwu, line 11: nit: retrieves

alexmwu (approved):

PR #591, opened 2025-09-05
Replace eventlog logic to go-eventlog

jkl73:
Remove tcg eventlog logic in server/eventlog, use go-eventlog instead.
Update go-tdx-guest package
Add VerifyOpts.AllowEmptySBVar, which is an opt in go-eventlog Add FirmwareLogState -> MachineState conversion func and tests


server/eventlog.go
```diff
@@ -247,431 +309,32 @@ func getVerifiedCosState(coscel cel.CEL, registerType uint8) (*pb.AttestedCosSta
...
-		digest := hasher.Sum(nil)
-
-		pbEvents[i] = &pb.Event{
-			PcrIndex:       uint32(event.Index),
-			UntrustedType:  uint32(event.Type),
-			Data:           event.Data,
-			Digest:         event.Digest,
-			DigestVerified: bytes.Equal(digest, event.Digest),
-		}
+func convertToPbDatabase(cert *gepb.Database) *pb.Database {
```
alexmwu: cert should be named db or gedb

server/eventlog.go
```diff
@@ -247,431 +309,32 @@ func getVerifiedCosState(coscel cel.CEL, registerType uint8) (*pb.AttestedCosSta
...
-
-	return &pb.LinuxKernelState{CommandLine: cmdline}, nil
 }
 
-func getGrubKernelCmdlineSuffix(grubCmd []byte) int {
-	for _, prefix := range [][]byte{oldGrubKernelCmdlinePrefix, newGrubKernelCmdlinePrefix} {
-		if bytes.HasPrefix(grubCmd, prefix) {
-			return len(prefix)
+func matchWellKnownCert(cert *gepb.Certificate) (pb.WellKnownCertificate, error) {
+	if cert.GetWellKnown() != gepb.WellKnownCertificate_UNKNOWN {
```
alexmwu: nit: invert this if statement and return early to avoid nesting two conditionals

server/eventlog.go
```diff
@@ -44,50 +28,137 @@ var (
...
-	var grub *pb.GrubState
-	var kernel *pb.LinuxKernelState
-	if opts.Loader == GRUB {
-		grub, err = getGrubState(cryptoHash, rawEvents)
-		if err != nil {
-			errors = append(errors, err)
+	return convertToMachineState(tpmfirmwarestate)
+}
+
+func convertToMachineState(tpmfirmwarestate *gepb.FirmwareLogState) (*pb.MachineState, error) {
```
alexmwu: consider exporting this, as it is already implemented in google3

@yawangwang FYI

server/eventlog.go
```diff
@@ -44,50 +28,137 @@ var (
...
+				case gepb.GCEConfidentialTechnology_AMD_SEV:
+					return pb.GCEConfidentialTechnology_AMD_SEV
+				case gepb.GCEConfidentialTechnology_AMD_SEV_ES:
+					return pb.GCEConfidentialTechnology_AMD_SEV_ES
+				case gepb.GCEConfidentialTechnology_INTEL_TDX:
+					return pb.GCEConfidentialTechnology_INTEL_TDX
+				case gepb.GCEConfidentialTechnology_AMD_SEV_SNP:
+					return pb.GCEConfidentialTechnology_AMD_SEV_SNP
+				default:
+					return pb.GCEConfidentialTechnology_NONE
```
alexmwu: return an error by default

server/eventlog.go
```diff
@@ -44,50 +28,137 @@ var (
...
+		secureBootState = &pb.SecureBootState{
+			Enabled: tpmfirmwarestate.GetSecureBoot().GetEnabled(),
+			Db:      convertToPbDatabase(tpmfirmwarestate.GetSecureBoot().GetDb()),
+			Dbx:     convertToPbDatabase(tpmfirmwarestate.GetSecureBoot().GetDbx()),
+			Pk:      convertToPbDatabase(tpmfirmwarestate.GetSecureBoot().GetPk()),
+			Kek:     convertToPbDatabase(tpmfirmwarestate.GetSecureBoot().GetKek()),
 		}
 	}
 
+	var efiState *pb.EfiState
```
alexmwu, line 117: Please add support for https://github.com/google/go-eventlog/blob/main/proto/state.proto#L176-L181 here or create an issue

server/eventlog_test.go
```diff
@@ -561,202 +554,8 @@ var GdcHost = eventLog{
 	},
 }
 
-func TestParseEventLogs(t *testing.T) {
```
alexmwu, line 604: I know these are copied over to go-eventlog, but it would be good to keep these tests to ensure the output is still the same despite the conversion

server/eventlog.go
jkl73: done

server/eventlog.go
jkl73: done

server/eventlog.go
jkl73: done

server/eventlog.go
jkl73: done

server/eventlog.go
jkl73: added

server/eventlog_test.go
jkl73: added back, replace the grouperr with joined error logic

server/eventlog.go
```diff
@@ -43,51 +27,153 @@ var (
...
-		SecureBoot:  sbState,
+		Platform:    platformstate,
+		SecureBoot:  secureBootState,
 		Efi:         efiState,
 		RawEvents:   rawEvents,
-		Hash:        pcrs.GetHash(),
-		Grub:        grub,
-		LinuxKernel: kernel,
-	}, createGroupedError("failed to fully parse MachineState:", errors)
+		Hash:        tpmpb.HashAlgo(tpmfirmwarestate.Hash),
```
alexmwu: .GetHash()

server/eventlog_test.go
alexmwu, line 1: could you write a test that converts the serialized gtt MachineState to FirmwareLogState to ensure we get the same FirmwareLogState (or you can write a convert func to go the other way and test with that)

server/eventlog.go
jkl73: done

server/eventlog_test.go
jkl73: done

jkl73: only thing is log_type is lost during the conversion, so I ignore the field in the test.

PR #595, opened 2025-09-16
Implement structured Cloud logging

Sibcgh:
Replaced `os.Stdout`-based container log redirection with structured Cloud Logging stream adapters and introduced a standalone `CloudLogger`.

*   **Dedicated CloudLogger**: Added `NewCloudLogger` returning a struct that implements `Logger` by writing exclusively to Cloud Logging.
*   **Workload Log Redirection**: Added `SeverityWriter` (`NewInfoWriter` / `NewErrorWriter`) to stream container `stdout` (`INFO`) and `stderr` (`ERROR`) line-by-line to Cloud Logging.
*   **Serial Console Bypass**: Directs container logs strictly to Cloud Logging, preventing sensitive log leakage on `/dev/console`.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -827,8 +827,10 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 		streamOpt = cio.WithStreams(nil, w, w)
 		r.logger.Info("Container stdout/stderr will be redirected to serial and Cloud Logging. This may result in performance issues due to slow serial console writes.")
 	case spec.CloudLogging:
-		streamOpt = cio.WithStreams(nil, os.Stdout, os.Stdout)
-		r.logger.Info("Container stdout/stderr will be redirected to Cloud Logging.")
+		stdoutWriter := logging.NewInfoWriter(r.logger)
+		stderrWriter := logging.NewErrorWriter(r.logger)
+		streamOpt = cio.WithStreams(nil, stdoutWriter, stderrWriter)
+		r.logger.Info("Container stdout/stderr will be redirected to Cloud Logging with INFO and ERROR severities respectively.")
 	case spec.Serial:
 		streamOpt = cio.WithStreams(nil, r.serialConsole, r.serialConsole)
 		r.logger.Info("Container stdout/stderr will be redirected to serial logging. This may result in performance issues due to slow serial console writes.")
--- a/launcher/internal/logging/logging.go
+++ b/launcher/internal/logging/logging.go
@@ -7,6 +7,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 
@@ -27,6 +28,7 @@ const (
 )
 
 // Logger defines the interface for the CS image logger.
+// Callers should run `defer logger.Close()` after initialization to ensure logs are flushed and handles are closed.
 type Logger interface {
 	Log(severity clogging.Severity, msg string, args ...any)
 
@@ -35,6 +37,7 @@ type Logger interface {
 	Error(msg string, args ...any)
 
 	SerialConsoleFile() *os.File
+	// Close flushes buffered logs and closes underlying resources. Callers should defer Close().
 	Close()
 }
 
@@ -51,6 +54,13 @@ type logger struct {
 	instanceName      string
 	cloudClient       *clogging.Client
 	serialConsoleFile *os.File
+}
+
+type cloudLogger struct {
+	cloudLogger  cLogger
+	resource     *mrpb.MonitoredResource
+	instanceName string
+	cloudClient  *clogging.Client
 }
 
 type payload map[string]any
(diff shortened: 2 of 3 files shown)


launcher/internal/logging/logging.go
```diff
@@ -31,6 +32,8 @@ type Logger interface {
 
 	SerialConsoleFile() *os.File
 	Close()
+
+	CloudOnlyWriter() io.Writer
```
brongan: Consider [the minimal viable interface our code will accept](https://dave.cheney.net/2016/08/20/solid-go-design) as well as the go proverb ["The bigger the interface, the weaker the abstraction."](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=5m17s), as well as the Liskov Substitution Principle. 

A `Logger` interface reasonably has methods for `Log`, `Info`, `Warn`, `Error`, and `Close`. Structures that implement the `Logger` interface might log to sinks like `stdout`, `Cloud logging`, or maybe even multiplex or broadcast logs. 

Adding a `CloudOnlyWriter` method that returns an `io.Writer` to a `Logger` interface breaks this abstraction. It forces all implementations of the Logger interface to support Cloud Logging.

brongan (changes requested):

launcher/internal/logging/logging.go
```diff
@@ -255,3 +263,42 @@ func (l *slogger) SerialConsoleFile() *os.File {
...
+	severity clogging.Severity
+}
+
+// NewSeverityWriter returns an io.Writer that writes to the provided Logger with a specific severity.
+func NewSeverityWriter(l Logger, severity clogging.Severity) io.Writer {
+	return &SeverityWriter{l: l, severity: severity}
+}
+
+// NewStdoutWriter returns an io.Writer that writes stdout logs to Cloud Logging.
+func NewStdoutWriter(l Logger) io.Writer {
```
jessieqliu: nit: `NewInfoWriter` and `NewErrorWriter` below - writing to STDOUT and STDERR is context important to the caller but not this particular library. As far as this library is concerned, these are just using a certain severity by default and the specific logging channel depends on the provided Logger.

launcher/container_runner.go
```diff
@@ -800,8 +800,10 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 		streamOpt = cio.WithStreams(nil, w, w)
 		r.logger.Info("Container stdout/stderr will be redirected to serial and Cloud Logging. This may result in performance issues due to slow serial console writes.")
 	case spec.CloudLogging:
-		streamOpt = cio.WithStreams(nil, os.Stdout, os.Stdout)
-		r.logger.Info("Container stdout/stderr will be redirected to Cloud Logging.")
+		stdoutWriter := logging.NewStdoutWriter(r.logger)
+		stderrWriter := logging.NewStderrWriter(r.logger)
+		streamOpt = cio.WithStreams(nil, stdoutWriter, stderrWriter)
+		r.logger.Info("Container stdout/stderr will be redirected to Cloud Logging with separate severities.")
```
jessieqliu: nit: we can specify the severities will be INFO and ERROR respectively.

launcher/internal/logging/logging.go
```diff
@@ -4,254 +4,284 @@ package logging
...
+		next.handlers[i] = sub.WithGroup(name)
 	}
-	l.slg.Log(context.Background(), level, msg, args...)
+	return next
 }
 
-// Info logs msg and args at 'Info' severity.
-func (l *slogger) Info(msg string, args ...any) {
-	l.slg.Info(msg, args...)
+// LevelWriter converts a *slog.Logger into an io.Writer.
```
jessieqliu: Out of curiosity, could `io.Writer` compatibility be achieved by adding `Write(...)` to the existing `logger` type so it implements both `slog.Logger` and `io.Writer` rather than wrapping a `logger` in a new struct?

Sibcgh: A struct can only implement `Write(...)` once, which locks it to a single default behavior. Wrapper structs let us create multiple `io.Writer` adapters from the same logger with different severities (`Info` vs. `Error`) or routing destinations.

launcher/internal/logging/logging.go
Sibcgh: Agreed. I narrowed `Logger` to just core logging methods (`Log`, `Info`, `Warn`, `Error`, `Close`) and segregated sink-specific methods (`CloudOnlyWriter`, `SerialConsoleFile`) into optional interfaces to avoid interface pollution and dummy `nil` returns.

brongan: Optional interfaces violate the Liskov Substitution Principle. I don't believe that we need them. I suggest leaving the logging interface completely opaque to the type of sink.

launcher/internal/logging/logging.go
```diff
@@ -33,9 +34,26 @@ type Logger interface {
 	Info(msg string, args ...any)
 	Warn(msg string, args ...any)
 	Error(msg string, args ...any)
+	Close()
```
brongan: Does this interface have an expectation that users of `Logger` run `defer Logger.Close()`?
If that is the case, please add it as a comment.

launcher/internal/logging/logging.go
```diff
@@ -33,9 +34,26 @@ type Logger interface {
...
+}
+
+// SerialConsoleProvider defines an optional interface for loggers that write to a serial console file.
+type SerialConsoleProvider interface {
 	SerialConsoleFile() *os.File
-	Close()
+}
+
+// GetSerialConsoleFile returns the underlying serial console os.File if l supports it, or nil otherwise.
+func GetSerialConsoleFile(l Logger) *os.File {
```
brongan: Can we implement this in a way where we avoid downcasting interfaces into concrete types?

launcher/internal/logging/logging.go
```diff
@@ -165,7 +188,10 @@ func addArgs(pl payload, args []any) {
 	addArgs(pl, args[2:])
 }
 
-func (l *logger) writeLog(severity clogging.Severity, msg string, args ...any) {
+func (l *logger) writeCloudLog(severity clogging.Severity, msg string, args ...any) {
+	if l.cloudLogger == nil {
+		return
```
brongan, line 269: Rather than having a `writeCloudLog` function that only works if `l.cloudLogger` is not nil and otherwise does nothing without returning an error, I suggest we instead of a `CloudLogger` struct that implements the `Logger` interface by logging to Cloud Logging.

brongan (changes requested):
Good improvements!

launcher/internal/logging/logging.go
Sibcgh: Since this PR is now simplified to focus strictly on the `SeverityWriter` adapter and container stream redirection, let's merge this first. When [PR #856](https://github.com/google/go-tpm-tools/pull/856/) lands on top, `SeverityWriter` can cleanly wrap the new `cloudLogger` instance directly.

brongan (approved):

jessieqliu:
Code changes LGTM but approval pending passing of presubmit tests.

launcher/internal/logging/logging.go
```diff
@@ -27,6 +28,7 @@ const (
 )
 
 // Logger defines the interface for the CS image logger.
+// Callers should run `defer logger.Close()` after initialization to ensure logs are flushed and handles are closed.
```
lindsayzad, line 31: If this should always happen, is it possible to make it part of the initialization itself, such as in NewCloudLogger?

launcher/container_runner.go
```diff
@@ -827,8 +827,10 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 		streamOpt = cio.WithStreams(nil, w, w)
 		r.logger.Info("Container stdout/stderr will be redirected to serial and Cloud Logging. This may result in performance issues due to slow serial console writes.")
 	case spec.CloudLogging:
-		streamOpt = cio.WithStreams(nil, os.Stdout, os.Stdout)
-		r.logger.Info("Container stdout/stderr will be redirected to Cloud Logging.")
+		stdoutWriter := logging.NewInfoWriter(r.logger)
+		stderrWriter := logging.NewErrorWriter(r.logger)
+		streamOpt = cio.WithStreams(nil, stdoutWriter, stderrWriter)
+		r.logger.Info("Container stdout/stderr will be redirected to Cloud Logging with INFO and ERROR severities respectively.")
```
lindsayzad, line 833: Were you able to do any testing that the logs are redirected as intended?

Sibcgh: Yes, works as intended.

launcher/internal/logging/logging.go
Sibcgh: We can't do this inside `NewCloudLogger` because `defer` runs as soon as `NewCloudLogger` returns, which would close the underlying logging client before the caller can even use it. 

Instead, callers (like `main`) need to defer `logger.Close()` so buffered logs are flushed cleanly right before process exit.

lindsayzad (approved):

yawangwang:
Functionality wise, this PR will allow workload send logs to cloud logging with severities, but this will bypass the fluent-bit routing if redirect_log=cloud_logging. Is this the intended behavior we want?

cc: @jkl73

PR #596, opened 2025-09-23
cel.AppendEventPCR extends to all PCR banks

jkl73:
BREAKING CHANGE: cel.AppendEventPCR and cel.AppendEvent stop taking in []crypto.Hash, instead they will get all available PCR banks from the TPM capability.

Make AllocatedPCRs public

### DIFF
--- a/cel/canonical_eventlog_test.go
+++ b/cel/canonical_eventlog_test.go
@@ -28,10 +28,10 @@ func TestCELEncodingDecoding(t *testing.T) {
 	cel := &CEL{}
 
 	cosEvent := CosTlv{ImageDigestType, []byte("sha256:781d8dfdd92118436bd914442c8339e653b83f6bf3c1a7a98efcfb7c4fed7483")}
-	appendPcrEventOrFatal(t, cel, tpm, test.DebugPCR, measuredHashes, cosEvent)
+	appendPcrEventOrFatal(t, cel, tpm, test.DebugPCR, cosEvent)
 
 	cosEvent2 := CosTlv{ImageRefType, []byte("docker.io/bazel/experimental/test:latest")}
-	appendPcrEventOrFatal(t, cel, tpm, test.ApplicationPCR, measuredHashes, cosEvent2)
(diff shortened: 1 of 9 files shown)


server/verify_test.go
```diff
@@ -385,11 +383,11 @@ func TestVerifyFailWithTamperedCELContent(t *testing.T) {
 	cosEvent := cel.CosTlv{EventType: cel.ImageRefType, EventContent: []byte("docker.io/bazel/experimental/test:latest")}
 	cosEvent2 := cel.CosTlv{EventType: cel.ImageDigestType, EventContent: []byte("sha256:781d8dfdd92118436bd914442c8339e653b83f6bf3c1a7a98efcfb7c4fed7483")}
 
-	if err := c.AppendEventPCR(rwc, cel.CosEventPCR, measuredHashes, cosEvent); err != nil {
+	if err := c.AppendEventPCR(rwc, cel.CosEventPCR, cosEvent); err != nil {
 		t.Fatalf("failed to append event: %v", err)
 	}
 
-	if err := c.AppendEventPCR(rwc, cel.CosEventPCR, measuredHashes, cosEvent2); err != nil {
```
alexmwu, line 392: can we check all the allocated PCRs change in a test somewhere?

cel/canonical_eventlog_test.go
```diff
@@ -162,15 +162,55 @@ func TestCELReplayFailMissingPCRsInBank(t *testing.T) {
...
+	}
+	for _, bank := range pcrs {
+		// make sure debug pcr is NOT empty after the append
+		if isZeroBytes(bank.Pcrs[uint32(test.DebugPCR)]) {
+			t.Fatalf("PCR %d in bank %s is empty after appending event", test.DebugPCR, bank.Hash.String())
+		}
+	}
+}
+
+func isZeroBytes(bs []byte) bool {
```
alexmwu, line 205: nit: allZeros := make([]byte, len(bs))
return allZeros == bs

jkl73: done

server/verify_test.go
jkl73: added

PR #597, opened 2025-09-29
Remove Attested COS specific CEL parsing logic

meetrajvala:
This PR removes the Attested COS state specific CEL parsing logic which would help us use the `VerifyAttestation` method for the clients who wants to parse the CEL differently. This PR keeps the `ParseCOSCELPCR` method (under `eventlog.go`) as is if existing client want to use it. 

It updates the implementation of fake verifier with additional logic of parsing COS state. This would make agent tests (in launcher) continue to work without any issue. 

**Testing:**
- Unit tests of server package passed successfully.
- Successfully built the gotpm library.

### DIFF
--- a/server/verify_test.go
+++ b/server/verify_test.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"bytes"
 	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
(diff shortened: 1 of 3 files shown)


server/verify_test.go
```diff
@@ -278,149 +273,6 @@ func TestVerifySHA1Attestation(t *testing.T) {
 	}
 }
 
-func TestVerifyAttestationWithCEL(t *testing.T) {
```
alexmwu, line 281: Can we keep these tests somewhere? Perhaps in the fake implementation or in g3?

meetrajvala: Sure, recently we have added test similar to`TestVerifyAttestationWithCEL` in g3 as part of refactoring changes. We will add `TestVerifyFailWithTamperedCELContent` test in g3.

PR #600, opened 2025-10-16
Add SVSM e-vTPM support to go-tpm-tools

jxu023:
SVSM e-vTPM currently can only attest to an EK, so we need EK-based key attestation for the TPM.

This PR implements https://trustedcomputinggroup.org/wp-content/uploads/EK-Based-Key-Attestation-with-TPM-Firmware-Version-V1-RC1_9July2025.pdf#page=8 and is based on the reference code https://github.com/TrustedComputingGroup/tpm-fw-attestation-reference-code

### DIFF
--- a/proto/attest/attest.pb.go
+++ b/proto/attest/attest.pb.go
@@ -253,7 +253,8 @@ func (GPUDeviceCCMode) EnumDescriptor() ([]byte, []int) {
 
 // Information uniquely identifying a GCE instance. Can be used to create an
 // instance URL, which can then be used with GCE APIs. Formatted like:
-//   https://www.googleapis.com/compute/v1/projects/{project_id}/zones/{zone}/instances/{instance_name}
+//
+//	https://www.googleapis.com/compute/v1/projects/{project_id}/zones/{zone}/instances/{instance_name}
 type GCEInstanceInfo struct {
 	state         protoimpl.MessageState
 	sizeCache     protoimpl.SizeCache
@@ -360,6 +361,7 @@ type Attestation struct {
 	// environment.
 	//
 	// Types that are assignable to TeeAttestation:
+	//
 	//	*Attestation_SevSnpAttestation
 	//	*Attestation_TdxAttestation
 	TeeAttestation isAttestation_TeeAttestation `protobuf_oneof:"tee_attestation"`
@@ -575,6 +577,7 @@ type PlatformState struct {
 	unknownFields protoimpl.UnknownFields
 
 	// Types that are assignable to Firmware:
+	//
 	//	*PlatformState_ScrtmVersionId
 	//	*PlatformState_GceVersion
 	Firmware isPlatformState_Firmware `protobuf_oneof:"firmware"`
@@ -933,6 +936,7 @@ type Certificate struct {
 	// the enum. Otherwise, it will contain the raw DER.
 	//
 	// Types that are assignable to Representation:
+	//
 	//	*Certificate_Der
 	//	*Certificate_WellKnown
 	Representation isCertificate_Representation `protobuf_oneof:"representation"`
@@ -1631,6 +1635,7 @@ type MachineState struct {
 	// verified.
 	//
 	// Types that are assignable to TeeAttestation:
+	//
 	//	*MachineState_SevSnpAttestation
 	//	*MachineState_TdxAttestation
 	TeeAttestation isMachineState_TeeAttestation `protobuf_oneof:"tee_attestation"`


client/import_certify.go
```diff
@@ -0,0 +1,140 @@
+package client
+
+import (
+	"fmt"
+
+	legacytpm2 "github.com/google/go-tpm/legacy/tpm2"
```
chrisfenner: Are we bringing in `legacytpm2` just for `AlgECC` and `AlgRSA`? Why not use `tpm2.TPMAlgID` (`TPMAlgRSA` and `TPMAlgECC`)?

client/import_certify.go
```diff
@@ -0,0 +1,140 @@
...
+	}.Execute(tpm)
+	if err != nil {
+		return nil, err
+	}
+	return cp, nil
+}
+
+// SolveChallengeImportCertify solves a challenge from the TPM EK.
+func SolveChallengeImportCertify(tpm transport.TPM, req *tpb.ImportBlob, keyAlgo legacytpm2.Algorithm) (*rpb.CertifiedBlob, error) {
+	// SVSM currently only supports attesting an RSA EK.
```
chrisfenner: I would recommend supporting using the ECC EK (e.g., P256) sooner than later. It's much faster (esp. key generation) and a little bit more secure (trade-off is win/win).

cmd/attest.go
```diff
@@ -121,16 +121,21 @@ hardware and guarantees a fresh quote.
...
-		var out []byte
-		if format == "binarypb" {
-			out, err = proto.Marshal(attestation)
-			if err != nil {
-				return fmt.Errorf("failed to marshal attestation proto: %v", attestation)
-			}
-		} else {
-			out = []byte(marshalOptions.Format(attestation))
+		if svsm != "" {
+                       var cred *rpb.SevSnpSvsmAttestation = &rpb.SevSnpSvsmAttestation{}
```
chrisfenner: nit: why not simplify as `cred := &rpb.SevSnpSvsmAttestation{}`

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+	h.Write(digest[:])
+	if subtle.ConstantTimeCompare(ha.Digest, h.Sum(nil)) != 1 {
+		return errInvalidHMAC
+	}
+	return nil
+}
+
+// verifyAttest checks that the attestation structure has valid data
+func verifyAttest(attest *tpm2.TPMSAttest) error {
+	if attest.Magic != tpm2.TPMGeneratedValue {
```
chrisfenner, line 166: also check that `attest.Type == TPMSTAttestCertify`

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+		return fmt.Errorf("%w: unexpected prefix %0x", errInvalidAttestation, attest.Magic)
+	}
+
+	// TODO: check qualified signer?
```
chrisfenner: I don't think we need to check the qualified signer, since the signer must be the same HMAC key we put into the challenge blob in order for the HMAC verify to pass

server/svsm.go
```diff
@@ -0,0 +1,43 @@
...
+	"bytes"
+	"crypto/sha512"
+	"fmt"
+
+	sevabi "github.com/google/go-sev-guest/abi"
+	"github.com/google/go-sev-guest/verify"
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+)
+
+// VerifySvsmAttestation checks the SNP attestation report to ensure it was requested by
```
chrisfenner: Nit: SVSM (consistent capitalization)

chrisfenner (approved):
No serious complaints from my end, although I focused my review on the crypto & attestation flow rather than the proto design.

client/import_certify.go
jxu023: Updated to use tpm2.TPMAlgID around these files, cmd/register.go now converts the uint16 to TPMAlgID (since we're using the existing keyAlgo flag shared with other commands).

client/import_certify.go
jxu023: Created a bug to do this in backlog

cmd/attest.go
jxu023: Done

server/import_certify.go
jxu023: Done

server/import_certify.go
jxu023: Removed TODO

server/svsm.go
jxu023: Done

client/svsm.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/google/go-tpm/tpm2"
+	sabi "github.com/google/go-sev-guest/abi"
+	sg "github.com/google/go-sev-guest/client"
+	sevpb "github.com/google/go-sev-guest/proto/sevsnp"
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+	tpb "github.com/google/go-tpm-tools/proto/tpm"
+)
+
+// MakeSVSMAttestation creates a SevSnpSvsmAttestation containing all the information needed to verify the SVSM e-vTPM.
+func MakeSVSMAttestation(tpm transport.TPM, req *tpb.ImportBlob, keyAlgo tpm2.TPMAlgID, teeNonce [sabi.ReportDataSize]byte) (*rpb.SevSnpSvsmAttestation, error) {
```
alexmwu: Would it be better to make the ReportDataSize a runtime check in case that ever changes?

cmd/attest.go
```diff
@@ -51,6 +54,7 @@ key or AK for a custom attestation key. By default it uses AK.
 by default rsa is used.
 --tee-nonce attaches a 64 bytes extra data to the attestation report of TDX and SEV-SNP 
 hardware and guarantees a fresh quote.
+--svsm will specify svsm as the service provider (Linux only) for the attestation report.
```
alexmwu: --svsm will specify SVSM as the...

proto/register_credential.proto
```diff
@@ -0,0 +1,27 @@
...
+option go_package = "github.com/google/go-tpm-tools/proto/register_credential";
+
+// Contains proof that the ak_pub is co-resident with an EK
+message CertifiedBlob {
+  // TPM2BPublic not including length field
+  bytes ak_pub = 1;
+  // TPMS_ATTEST, contains firmware version
+  bytes certify_info = 2;
+  // TPMTSignature
+  bytes signature = 3;
```
alexmwu: TPM2 signature, encoded as a TPMT_SIGNATURE
bytes raw_sig

for consistency with https://github.com/google/go-tpm-tools/blob/main/proto/tpm.proto#L45C6-L46

proto/register_credential.proto
```diff
@@ -0,0 +1,27 @@
...
+
+package register_credential;
+
+import "proto/sevsnp.proto";
+
+option go_package = "github.com/google/go-tpm-tools/proto/register_credential";
+
+// Contains proof that the ak_pub is co-resident with an EK
+message CertifiedBlob {
+  // TPM2BPublic not including length field
```
alexmwu: // Attestation Key (AK) Public Area, encoded as a TPMT_PUBLIC

for consistency with https://github.com/google/go-tpm-tools/blob/main/proto/attest.proto#L23


also isn't "TPM2BPublic not including length field" just TPMT_PUBLIC?

proto/register_credential.proto
```diff
@@ -0,0 +1,27 @@
...
+  bytes ak_pub = 1;
+  // TPMS_ATTEST, contains firmware version
```
alexmwu: // TPM2 certification, encoded as a TPMS_ATTEST

for consistency with https://github.com/google/go-tpm-tools/blob/main/proto/tpm.proto#L43C17-L44C1

proto/register_credential.proto
```diff
@@ -0,0 +1,27 @@
...
+  bytes signature = 3;
+}
+
+// Contains all information needed to verify the binding from TPM EK to SNP
+// Attestation report.
+message SevSnpSvsmAttestation {
+  sevsnp.Attestation sev_snp_attestation = 1;
+  CertifiedBlob certified_blob = 2;
+  // vTPM service manifest retrieved from configfs-tsm via manifest blob
+  // its format is defined by the SVSM spec
```
alexmwu: Capitalize: Its format is...

client/import_certify.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpm2/transport"
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+	tpb "github.com/google/go-tpm-tools/proto/tpm"
+)
+
+// this file aims to implement the attester side of https://trustedcomputinggroup.org/wp-content/uploads/EK-Based-Key-Attestation-with-TPM-Firmware-Version-V1-RC1_9July2025.pdf#page=8
+// for reference: https://github.com/TrustedComputingGroup/tpm-fw-attestation-reference-code
+
+func ekPub(tpm transport.TPM) ([]byte, error) {
```
alexmwu: Note ekPub and ekPolicy are already implemented in go-tpm legacy with the key primitive. Please create an issue to remove these if/when we move completely over to TPM-direct.

client/import_certify.go
```diff
@@ -0,0 +1,139 @@
...
+		CertifyInfo: certified.CertifyInfo.Bytes(),
+		Signature:   tpm2.Marshal(certified.Signature),
+	}, nil
+}
+
+func ekPolicy(t transport.TPM, handle tpm2.TPMISHPolicy, nonceTPM tpm2.TPM2BNonce) error {
+	cmd := tpm2.PolicySecret{
+		AuthHandle:    tpm2.TPMRHEndorsement,
+		PolicySession: handle,
+		NonceTPM:      nonceTPM,
```
alexmwu, line 128: @chrisfenner why do we need the nonce? I didn't think the EK Credential Profile set it

cmd/attest.go
```diff
@@ -121,16 +121,21 @@ hardware and guarantees a fresh quote.
...
-		var out []byte
-		if format == "binarypb" {
-			out, err = proto.Marshal(attestation)
-			if err != nil {
-				return fmt.Errorf("failed to marshal attestation proto: %v", attestation)
-			}
-		} else {
-			out = []byte(marshalOptions.Format(attestation))
+		if svsm != "" {
+                       cred := &rpb.SevSnpSvsmAttestation{}
```
alexmwu: Is it just me or is the whitespace inconsistent here?

client/svsm.go
```diff
@@ -0,0 +1,93 @@
...
+	certified, err := SolveChallengeImportCertify(tpm, req, keyAlgo)
+	if err != nil {
+		return nil, fmt.Errorf("SolveChallengeImportCertify() = %w", err)
```
alexmwu: here and elsewhere, please match the error string style. for example: "failed to create import certify solution".

client/import_certify.go
```diff
@@ -0,0 +1,139 @@
...
+		InPublic:      tpm2.BytesAs2B[tpm2.TPMTPublic](public),
+	}.Execute(tpm)
+	if err != nil {
+		return nil, err
+	}
+	return cp, nil
+}
+
+// SolveChallengeImportCertify solves a challenge from the TPM EK.
+func SolveChallengeImportCertify(tpm transport.TPM, req *tpb.ImportBlob, keyAlgo tpm2.TPMAlgID) (*rpb.CertifiedBlob, error) {
```
alexmwu: Do we need to export this? Also does a name like createCertifiedAKBlob work better?

client/import_certify.go
```diff
@@ -0,0 +1,139 @@
...
+	ak, err := makeAK(tpm, keyAlgo)
+	if err != nil {
+		return nil, fmt.Errorf("AK() = %w", err)
+	}
+	defer tpm2.FlushContext{
+		FlushHandle: ak.ObjectHandle,
+	}.Execute(tpm)
+
+	// Certify a newly created AK
+	certified, err := tpm2.Certify{
```
alexmwu, line 103: Do we want to support a nonce in Certify so we know the certify was for a specific registration flow? @chrisfenner

client/svsm.go
alexmwu: For consistency with the rest of the library, most of these parameters should be fields in an Opts struct (see AttestOpts as an example).

something like func MakeSVSMAttestation(tpm transport.TPM, svsmOpts SVSMOpts)

cmd/pubkey.go
```diff
@@ -12,8 +12,15 @@ import (
 
 	"github.com/google/go-tpm/legacy/tpm2"
 	"github.com/spf13/cobra"
+	directtpm2 "github.com/google/go-tpm/tpm2"
 )
 
+var keyFormat string
+
+func addKeyFormatFlag(cmd *cobra.Command) {
+       cmd.PersistentFlags().StringVar(&keyFormat, "key-format", "pem", "type of format for the outputted key, defaults to pem, but can also specify tpm2-wire-format")
```
alexmwu: Be specific about the format (TPMT_PUBLIC/tpmt-public)

cmd/pubkey.go
```diff
@@ -98,3 +121,16 @@ func writeKey(pubKey crypto.PublicKey) error {
 		Bytes: asn1Bytes,
 	})
 }
+
+// ReadTPM2WireFormatKey reads a TPM2 wire format public key from the given reader.
+func ReadTPM2WireFormatKey(rw io.Reader) (*directtpm2.TPMTPublic, error) {
```
alexmwu: Does this need to be exported? readTPMTPublic

cmd/register.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/google/go-tpm/tpm2/transport"
+	"github.com/google/go-tpm/tpm2"
+	sevabi "github.com/google/go-sev-guest/abi"
+	"github.com/google/go-tpm-tools/client"
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+	tpb "github.com/google/go-tpm-tools/proto/tpm"
+	"github.com/google/go-tpm-tools/server"
+)
+
+// this file implements TPM registration which for now is only used for SVSM e-vTPMs.
```
alexmwu: Capitalize comments and end with periods

server/svsm.go
```diff
@@ -0,0 +1,43 @@
...
+	sevabi "github.com/google/go-sev-guest/abi"
+	"github.com/google/go-sev-guest/verify"
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+)
+
+// VerifySVSMAttestation checks the SNP attestation report to ensure it was requested by
+// SVSM from VMPL0 and that report data contains a hash of the tee nonce and vTPM service manifest.
+// other SNP attestation report values (including the launch measurement)
+// should be checked via gce tcb verifier's sev validate command
+func VerifySVSMAttestation(teeNonce [sevabi.ReportDataSize]byte, attestation *rpb.SevSnpSvsmAttestation, secret []byte) error {
```
alexmwu: VerifySVSMAttestation(attestation *rpb.SevSnpSvsmAttestation, opts VerifySVSMOpts)

server/import_certify.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return &tpb.ImportBlob{
+		PublicArea:    tpm2.Marshal(pub),
+		Duplicate:     duplicate,
+		EncryptedSeed: inSymSeed,
+	}, hmacKey, nil
+}
+
+// VerifyChallenge verifies the results of the TPM EK registration flow.
+func VerifyChallenge(req *rpb.CertifiedBlob, secret []byte) error {
```
alexmwu: Let's avoid exporting unless we need to. I'd prefer to export when we need it rather than intro breaking changes or unexport.

server/import_certify.go
```diff
@@ -0,0 +1,189 @@
...
+
+var (
+	errCertifiedWrongName = errors.New("incorrect name")
+	errWrongHashAlg       = errors.New("wrong hash algorithm")
+	errInvalidHMAC        = errors.New("invalid HMAC")
+	errInvalidAttestation = errors.New("attestation statement was invalid")
+)
+
+// CreateChallenge generates a new HMAC key and wraps it to the given EK.
+func CreateChallenge(tPublic *tpm2.TPMTPublic) (*tpb.ImportBlob, []byte, error) {
```
alexmwu: I don't have a great name but we should try to be clear since the Rest HMAC flow is a type of challenge. Maybe CreateRestrictedHMACBlob?

cmd/attest.go
```diff
@@ -121,16 +121,21 @@ hardware and guarantees a fresh quote.
...
 
-		var out []byte
-		if format == "binarypb" {
-			out, err = proto.Marshal(attestation)
-			if err != nil {
-				return fmt.Errorf("failed to marshal attestation proto: %v", attestation)
-			}
-		} else {
-			out = []byte(marshalOptions.Format(attestation))
+		if svsm != "" {
```
alexmwu: To be clear, we require the person to run the go-sev-guest CLI first and attach it to this proto via the --svsm flag? Wouldn't it be more usable to just create an appropriate SNP attestation in line here?

alexmwu (commented):
Can we add tests for these new additions? I remember go-sev-guest supports a mock https://github.com/google/go-sev-guest/blob/main/testing/test_cases.go#L304 but I'm not too familiar with it tbh

I'll need to do another round on the library crypto; thankfully Chris has already taken a look.

client/svsm.go
jxu023: Done.

cmd/attest.go
jxu023: Updated the comment here to note that this input is provided from the e-vTPM registration flow and used to populate the Tee Attestation field.

proto/register_credential.proto
jxu023: Done

proto/register_credential.proto
jxu023: Done

proto/register_credential.proto
jxu023: Done

proto/register_credential.proto
jxu023: Done

client/import_certify.go
jxu023: opened https://github.com/google/go-tpm-tools/issues/604

cmd/attest.go
jxu023: Fixed

client/svsm.go
jxu023: Fixed here and elsewhere

client/import_certify.go
jxu023: removed export via de-capitalization - changed name to createCertifiedAKBlob

client/import_certify.go
jxu023: per the EK based remote attestation w/ firmware spec, we can use a nonce with CertifyCreation(); I'll wait on chris to respond on the necessity. 

I suppose adding a nonce means we should also called the TPM clear() operation? Otherwise re-trying this section of code on the same TPM will fail.

client/svsm.go
jxu023: Done.

cmd/pubkey.go
jxu023: changed to tpmt-public

cmd/pubkey.go
jxu023: removed export, renamed

cmd/register.go
jxu023: Done.

server/svsm.go
jxu023: Done.

server/import_certify.go
jxu023: unexported

server/import_certify.go
jxu023: Done.

cmd/attest.go
jxu023: The SNP attestation is created in svsm.go:MakeSVSMAttestation() which bundles creating the certified AK blob with getting the attestation report. I figured putting it there makes sense since these should be verified together, and it's one step towards removing the TeeAttestation from AttestOpts which seems to be getting deprecated.

jxu023:
> Can we add tests for these new additions? I remember go-sev-guest supports a mock https://github.com/google/go-sev-guest/blob/main/testing/test_cases.go#L304 but I'm not too familiar with it tbh
> 
> I'll need to do another round on the library crypto; thankfully Chris has already taken a look.

I've uploaded another revision, still working on unit tests. I'll be updating the SVSMOpts to include configfsi and LeveledQuoteProvider interfaces so that I can inject a test double there

jxu023:
Updated with unit tests

client/import_certify.go
```diff
@@ -0,0 +1,141 @@
...
+
+func createCertifiedAKBlob(tpm transport.TPM, req *tpb.ImportBlob, keyAlgo tpm2.TPMAlgID) (*rpb.CertifiedBlob, error) {
+	// SVSM currently only supports attesting an RSA EK.
+	ek, err := tpm2.CreatePrimary{
+		PrimaryHandle: tpm2.TPMRHEndorsement,
+		InPublic:      tpm2.New2B(tpm2.RSAEKTemplate),
+	}.Execute(tpm)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create RSA EK: %w", err)
+	}
```
alexmwu: Could this and ekPub call an ek() func that returns the CreatePrimary resp object

proto/register_credential.proto
```diff
@@ -0,0 +1,27 @@
...
+package register_credential;
+
+import "proto/sevsnp.proto";
+
+option go_package = "github.com/google/go-tpm-tools/proto/register_credential";
+
+// Contains proof that the ak_pub is co-resident with an EK
+message CertifiedBlob {
+  // Attestation Key (AK) Public Area, encoded as a TPMT_PUBLIC
+  bytes ak_pub = 1;
```
alexmwu: // CertifiedBlob contains proof that the certified object's public area (Name) is loaded in the TPM.

CertifiedBlob can apply to any object, not just an AK. The the first field should just be the public area of the certified object.

client/import_certify_export_test.go
```diff
@@ -0,0 +1,17 @@
...
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+	tpb "github.com/google/go-tpm-tools/proto/tpm"
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpm2/transport"
+)
+
+func EKPub(tpm transport.TPM) ([]byte, error) {
+	return ekPub(tpm)
+}
+func CreateCertifiedAKBlob(tpm transport.TPM, req *tpb.ImportBlob, keyAlgo tpm2.TPMAlgID) (*rpb.CertifiedBlob, error) {
```
alexmwu: You can put these in an internal package in client if we need it for use in test (see the internal directory as an example)

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+	"crypto/subtle"
+	"errors"
+	"fmt"
+
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+	tpb "github.com/google/go-tpm-tools/proto/tpm"
+	"github.com/google/go-tpm/tpm2"
+)
+
+// this file aims to implement the verifier side of https://trustedcomputinggroup.org/wp-content/uploads/EK-Based-Key-Attestation-with-TPM-Firmware-Version-V1-RC1_9July2025.pdf#page=8
```
alexmwu: Capitalize comments.

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+// for reference: https://github.com/TrustedComputingGroup/tpm-fw-attestation-reference-code
+
+var (
+	errCertifiedWrongName = errors.New("incorrect name")
```
alexmwu, line 20: nit: capitalize name or call it a TPM name

proto/register_credential.proto
alexmwu: On that note, this message probably belongs in tpm.proto

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	return &tpb.ImportBlob{
+		PublicArea:    tpm2.Marshal(pub),
+		Duplicate:     duplicate,
+		EncryptedSeed: inSymSeed,
+	}, hmacKey, nil
+}
+
+func VerifyCertifiedAKBlob(req *rpb.CertifiedBlob, secret []byte) error {
```
alexmwu: This should probably be unexported if we're not exporting the client func. Since it's needed in the client tests, you can put it in an internal directory

server/import_certify.go
alexmwu: nit: Name or TPM name

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+	rand.Read(obfuscate)
+	rand.Read(hmacKey[:])
+
+	// Unique for a KEYEDHASH object is H_nameAlg(obfuscate | key)
+	// See Part 1, "Public Area Creation"
+	h := sha256.New()
+	h.Write(obfuscate)
+	h.Write(hmacKey[:])
+
+	pub := &tpm2.TPMTPublic{
```
alexmwu, line 112: This is heavily duplicated with https://github.com/google/go-tpm-tools/blob/main/server/key_conversion.go#L52-L64. Can we extend that to pass in the correct Object attributes (sealedDataObjects leave it unset; imported HMAC keys should set)

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+					HashAlg: tpm2.TPMAlgSHA256,
+				}),
+			},
+		}),
+		Unique: tpm2.NewTPMUPublicID(tpm2.TPMAlgKeyedHash, &tpm2.TPM2BDigest{
+			Buffer: h.Sum(nil),
+		}),
+	}
+
+	priv := &tpm2.TPMTSensitive{
```
alexmwu, line 134: duplicated with https://github.com/google/go-tpm-tools/blob/main/server/key_conversion.go#L66-L77

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+	h := hmac.New(sha256.New, hmacKey[:])
+	h.Write(digest[:])
+	if subtle.ConstantTimeCompare(ha.Digest, h.Sum(nil)) != 1 {
+		return errInvalidHMAC
+	}
+	return nil
+}
+
+// verifyAttest checks that the attestation structure has valid data
+func verifyAttest(attest *tpm2.TPMSAttest) error {
```
alexmwu: verifyAttestCertify

server/import_certify.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	if attest.Magic != tpm2.TPMGeneratedValue {
+		return fmt.Errorf("%w: unexpected prefix %0x", errInvalidAttestation, attest.Magic)
+	}
+
+	return nil
+}
+
+// verifyCertify checks the certifyInfo against the given name.
+func verifyCertify(name *tpm2.TPM2BName, certifyInfo *tpm2.TPMSCertifyInfo) error {
```
alexmwu: verifyCertifyInfo

client/svsm.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get leveled quote provider: %w", err)
+	}
+	rawWithCerts, err := qp.GetRawQuoteAtLevel(reportData, leastPrivilegedVMPL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get raw quote: %v", err)
+	}
+	reportWithCerts, err := sabi.ReportCertsToProto(rawWithCerts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse SEV-SNP report with certs: %v", err)
```
rossjudson: sabi.ReportsCertsToProto uses the %v err formatter, dropping context. Elsewhere, sabi.ReportToProto uses %v -- not dropping context.

Is this inconsistency intentional?

server/svsm.go
```diff
@@ -0,0 +1,51 @@
...
+	"crypto/sha512"
+	"fmt"
+
+	"github.com/google/go-sev-guest/verify"
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+)
+
+type VerifySVSMOpts struct {
+	TEENonce    []byte
+	Attestation *rpb.SevSnpSvsmAttestation
```
alexmwu: nit: VerifySVSMAttestation operates on a rpb.SevSnpSvsmAttestation, so this should not be a field in the opts struct but a top-level param.

server/svsm.go
```diff
@@ -0,0 +1,51 @@
...
+// SVSM from VMPL0 and that report data contains a hash of the tee nonce and vTPM service manifest.
+// Other SNP attestation report values (including the launch measurement)
+// should be checked via gce tcb verifier's sev validate command.
+func VerifySVSMAttestation(opts *VerifySVSMOpts) error {
+	err := VerifyCertifiedAKBlob(opts.Attestation.GetCertifiedBlob(), opts.Secret)
+	if err != nil {
+		return fmt.Errorf("challenge verification failed: %w", err)
+	}
+
+	h := sha512.New()
```
alexmwu: It would be useful to link the spec or the code for this protocol.

server/svsm.go
```diff
@@ -0,0 +1,51 @@
...
+	// Service manifest is the Ekpub.
+	h.Write(opts.Attestation.GetVtpmServiceManifest())
+	expectedReportData := h.Sum(nil)
+	report := opts.Attestation.GetSevSnpAttestation().GetReport()
+
+	if !bytes.Equal(report.GetReportData(), expectedReportData) {
+		return fmt.Errorf("report data does not match expected value")
+	}
+	if report.Vmpl != 0 {
+		return fmt.Errorf("attestation report was not requested from VMPL0, SVSM should be in VMPL0")
```
alexmwu: nit: this can be a sentinel error.

server/svsm.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/google/go-sev-guest/verify"
+	rpb "github.com/google/go-tpm-tools/proto/register_credential"
+)
+
+type VerifySVSMOpts struct {
+	TEENonce    []byte
+	Attestation *rpb.SevSnpSvsmAttestation
+	Secret      []byte
+	VerifyOpts  *verify.Options
+}
```
alexmwu: should we allow them to pass in go-sev-guest validate.Options? They will eventually need to compare the SVSM's measurement anyway right?

cmd/verify.go
```diff
@@ -150,6 +123,14 @@ func verifyGceTechnology(attestation *pb.Attestation) error {
 	}
 }
 
+var skipGceTechnology bool
+
+func addSkipGceTechnologyFlag(cmd *cobra.Command) {
```
alexmwu: Why do we need this flag?

cmd/register.go
```diff
@@ -0,0 +1,158 @@
...
+		if err != nil {
+			return fmt.Errorf("could not verify attestation: %s", err)
+		}
+		return nil
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(registerCmd)
+	challengeCommands := []*cobra.Command{
```
alexmwu, line 123: Is it possible to further nest it under gotpm register *svsm*? Either that or we pass in `--svsm` in such a way that we're able to know which registration protocol to use. We need to be able to separate out other registration methods (activate credential / import certify the "normal" way)

cmd/attest.go
alexmwu: The point was if we want to make it all verifiable in one go, why not just add it as part of attest rather than force someone to run solve-challenge and attach the svsm separately? IOW, `--svsm` should attach the SNP attestation related to an SVSM rather than have the user create the SNP attestation and pass it in.

alexmwu (approved):

client/import_certify.go
```diff
@@ -0,0 +1,141 @@
...
+	defer tpm2.FlushContext{
```
jkl73: looks like this is already flushed below line 95

client/import_certify.go
```diff
@@ -0,0 +1,141 @@
...
+
+func makeAK(tpm transport.TPM, keyAlgo tpm2.TPMAlgID) (*tpm2.CreatePrimaryResponse, error) {
+	var public []byte
+	var err error
+	if keyAlgo == tpm2.TPMAlgECC {
+		public, err = AKTemplateECC().Encode()
+	}
+	if keyAlgo == tpm2.TPMAlgRSA {
+		public, err = AKTemplateRSA().Encode()
+	}
```
jkl73, line 33: just to be safe, make sure keyAlgo is either ecc or rsa (maybe use a switch statement)

client/import_certify.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/google/go-tpm/tpm2/transport"
+)
+
+// This file aims to implement the attester side of https://trustedcomputinggroup.org/wp-content/uploads/EK-Based-Key-Attestation-with-TPM-Firmware-Version-V1-RC1_9July2025.pdf#page=8
+// For reference: https://github.com/TrustedComputingGroup/tpm-fw-attestation-reference-code
+
+func ekPub(tpm transport.TPM) ([]byte, error) {
+	cp, err := tpm2.CreatePrimary{
+		PrimaryHandle: tpm2.TPMRHEndorsement,
+		InPublic:      tpm2.New2B(tpm2.RSAEKTemplate),
```
jkl73, line 19: make tpm2.RSAEKTemplate a constant in this file, since this referred more than one time.

client/import_certify.go
```diff
@@ -0,0 +1,141 @@
...
+	defer tpm2.FlushContext{
+		FlushHandle: ek.ObjectHandle,
+	}.Execute(tpm)
+
+	// Import the restricted HMAC key.
+	imported, err := tpm2.Import{
+		ParentHandle: tpm2.AuthHandle{
+			Handle: ek.ObjectHandle,
+			Name:   ek.Name,
+			Auth:   tpm2.Policy(tpm2.TPMAlgSHA256, 16, ekPolicy),
```
jkl73: I thought the auth size is suppose to be 32 instead of 16? (I might be wrong)

cmd/register.go
```diff
@@ -0,0 +1,158 @@
...
+
+var solveChallengeCmd = &cobra.Command{
+	Use:   "solve-challenge",
+	Short: "Solve a challenge from the TPM EK",
+	Args:  cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		challenge := &tpb.ImportBlob{}
+		err := readProtoFromPath(input, challenge)
+		if err != nil {
+			return fmt.Errorf("could not read challnge: %s", err)
```
jkl73: typo: challenge

jkl73 (approved):
lgtm two overall thoughts: 
* The cmd looks a bit complicated, we might think how to streamline it.
* svsm specific code in client/server might be more suitable to put into go-sev-guest repo

server/import_certify.go
jxu023: From looking around, it looks like golang prevents you from casting between pointers to structs unless you use the `unsafe` package.

I've added a note to https://github.com/google/go-tpm-tools/issues/604 to note that we should de-duplicate it here as well.

server/import_certify.go
jxu023: From looking around, it looks like golang prevents you from casting between pointers to structs unless you use the `unsafe` package.

I've added a note to https://github.com/google/go-tpm-tools/issues/604 to note that we should de-duplicate it here as well.

server/import_certify.go
alexmwu: Can't you serialize to TPM format then deserialize to tpm2.TPMTPublic?

client/import_certify.go
jxu023: I used alex's suggestion and created an ekResponse() function wrapper.

client/import_certify.go
jxu023: modified to switch and added error on unsupported keyAlgo

client/import_certify.go
jxu023: Done.

client/import_certify.go
jxu023: this was added to ensure flushing from the err conditions despite the duplicate flush.

added explicit flush from the err cond early returns instead.

client/import_certify.go
jxu023: I tried with 32 and 16, and they both seem to work: from checking the TPM library ref: https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-2.0-Library-Part-1-Version-184_pub.pdf

It notes in _17.6.3.2 Session Nonce Size_
> The minimum size for nonceCaller in TPM2_StartAuthSession() is 16 octets
> ... The maximum size that may be requested for nonceTPM is the size of the digest produced by the authorization session hash
> For SHA-1, the maximum size for nonceTPM is 20 octets and for SHA256 it is 32 octets.

Octet means byte apparently.

i think both work. I've updated the value to 32 cuz why not, more secure probably.

client/import_certify_export_test.go
jxu023: the gotpm cmd for register has been made SVSM agnostic, so this is now exported and used from the gotpm register cmd

Deleted this export file since no longer needed.

cmd/attest.go
jxu023: created svsm subcommand to attach the SVSM attestation info

cmd/register.go
jxu023: fixed

cmd/register.go
jxu023: Moved svsm logic to svsm subcommand under attest and verify, register.go is now SVSM-free (except for one comment noting that we're only creating an EK RSA because that's all that's currently supported on SVSM).

cmd/verify.go
jxu023: added an svsm subcommand instead of verify, so we don't need this now.

My intention was to disable the snp validate check here since it will fail on an SVSM VM since its report data calculation doesn't factor in the vtpm service manifest.

proto/register_credential.proto
jxu023: moved to tpm proto

server/import_certify.go
jxu023: Capitalized

server/import_certify.go
jxu023: I wasn't clear on what to do here, could you clarify?

server/import_certify.go
jxu023: exported since now gotpm register cmd uses this instead of the SVSM wrapper

server/import_certify.go
jxu023: updated name

server/import_certify.go
jxu023: updated name

server/svsm.go
jxu023: moved to top-level param

server/svsm.go
jxu023: passed in validate options, and also attached endorsement into the SVSM message. Also added verification on endorsement by adding gcetcbendorsement as a dependency.

Now we no longer need to make separate CLI call to gce-tcb-verifier/gce-tcb-endorsement!

server/svsm.go
jxu023: added spec and code links

server/svsm.go
jxu023: replaced with sentinel errs here and elsewhere

client/svsm.go
jxu023: TIL about %w vs %v. I've replaced usages with %w in new code, except in unit tests where t.Fatalf() doesn't support %w.

jxu023:
> The cmd looks a bit complicated, we might think how to streamline it.

I've removed SVSM logic from tpm registration, so hopefully it's a bit simpler now.

> svsm specific code in client/server might be more suitable to put into go-sev-guest repo

I've moved svsm specific code to the cmd package instead of client/server. This way we can do a drop-in replacement when i move some of that to either go-sev-guest or tcb-verifier

I've uploaded a new revision which includes these changes:
* added go-sev-guest v0.14.0 to deps so that i can abandon https://github.com/google/go-tpm-tools/pull/608 if we merge this
*  added 2 second sleep prior to SVSM attestation report request (since Coconut SVSM will crash on hitting the guest request ratelimit)
* made svsm subcommands for gotpm attest/verify
* added gce-tcb-endorsement attestation/verification into svsm subcommands

@alexmwu @jkl73 lemme know how the changes look

Notably we no longer need to separately call gce-tcb-verifier CLI
full example usage:
```
TEE_NONCE="123456789012345678901234567890123456789012345678901234567890123456789012345
"
ALGO="ecc"

"${gotpm}" pubkey endorsement --key-format=tpmt-public \
  --algo=rsa --output=ek.bin
"${gotpm}" register create-challenge --input=ek.bin \
  --output=challenge.bin --secret-output=secret.bin
"${gotpm}" register solve-challenge --input=challenge.bin --algo=${ALGO} \
  --output=certified_blob.bin
"${gotpm}" register verify-challenge --input=certified_blob.bin \
  --secret-input=secret.bin

"${gotpm}" attest svsm --nonce=1234 --output=svsm_attestation.bin \
  --algo=${ALGO} --key=AK --tee-nonce="${TEE_NONCE}" \
  --tee-technology=sev-snp
"${gotpm}" verify debug svsm --nonce=1234 --input=svsm_attestation.bin \
  --tee-technology=sev-snp --certified-ak-blob=certified_blob.bin \
  --ek-pub=ek.bin --tee-nonce="${TEE_NONCE}"
```

server/import_certify.go
jxu023: ah, that'd slipped my mind; i'll try that

client/import_certify_test.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	testcases := []struct {
+		name    string
+		keyAlgo tpm2.TPMAlgID
+	}{
+		{"RSA", tpm2.TPMAlgRSA},
+		{"ECC", tpm2.TPMAlgECC},
+	}
+	for _, tc := range testcases {
```
alexmwu, line 45: nit: could the CreateCertifiedAKBlob be a receiver method on the ek Key. You can then call ek.PublicArea() below and ek.CreateCertifiedAKBlob(tpm, challenge, tc.keyAlgo)

server/import_certify.go
alexmwu: "incorrect TPM name"

proto/tpm.proto
```diff
@@ -52,3 +52,13 @@ message PCRs {
...
+
+// Contains information corresponding an object certified with TPM2_Certify.
+message CertifiedBlob {
+  // Public area of certified object, encoded as a TPMT_PUBLIC
+  bytes pub_area = 1;
+  // TPM2 certification, encoded as a TPMS_ATTEST
+  bytes certify_info = 2;
+  // TPM2 signature, encoded as a TPMT_Signature
+  bytes raw_sig = 3;
+}
```
alexmwu, line 64: nit: newline at end of file

proto/attest.proto
```diff
@@ -315,4 +329,4 @@ message Policy {
   // When the attestation is on SEV-SNP, this is the policy. Unset means no
   // constraints.
   SevSnpPolicy sev_snp = 3;
-}
+}
```
alexmwu, line 332: nit: newline at end of file.

alexmwu (approved):

PR #607, opened 2025-10-24
Update end-to-end testing to use fake verifier

mayafleischer:
Add fake verifier metadata variable and use of existing fake verifier, when variable present.

### DIFF
--- a/launcher/image/test/test_oda_with_signed_container.yaml
+++ b/launcher/image/test/test_oda_with_signed_container.yaml
@@ -5,9 +5,9 @@ substitutions:
   '_VM_NAME_PREFIX': 'oda-signedcontainer'
   '_ZONE': 'us-west1-a'
   # If the workload image changes, the commit author should change the cosign signature as well to not break tests.
-  '_WORKLOAD_IMAGE': 'us-west1-docker.pkg.dev/confidential-space-images-dev/cs-integ-test-images/ipc/happypath@sha256:d0662189b04f5872460d34bc85af008acda4330ffc65ef072fa84e43223cf05b'
+  '_WORKLOAD_IMAGE': 'us-west1-docker.pkg.dev/confidential-space-images-dev/cs-integ-test-images/ipc/happypath@sha256:836d4ad39fa0fec5b386c4051a175036bd13d74a75d57933d30f6b804e509c8c'
   '_SIGNATURE_REPO': 'us-docker.pkg.dev/confidential-space-images-dev/cs-cosign-tests/oda'
-  '_EXPECTED_SIG': 'MEUCIQC7mQSHRcTBOrkriS19oFzuopKC7ZW2g94kZ1kJ8521vgIgPQKeKUGzJPkU27IcqfWOrydy6fS0PPihSxmQXD0UHrQ='
+  '_EXPECTED_SIG': 'MEUCIFoV6g8jU+EDQZJsySY5KcHYh/yRepxDFSUjhcJZ5du8AiEAooEHwjKsPIxO24NYE565E56DLU9H9vNX+jTP1nYwfNE='
 
 steps:
 - name: 'gcr.io/cloud-builders/gcloud'


verifier/fake/fakeverifier.go
```diff
@@ -182,3 +196,41 @@ func extractPCRBank(attestation *attest.Attestation, hashAlgo tpm.HashAlgo) (*re
 	}
 	return nil, fmt.Errorf("no PCRs found matching hash %s", hashAlgo.String())
 }
+
+func fakePrivateKey() crypto.Signer {
+	// Hardcoded fake key
+	const privateKeyPEM = `-----BEGIN PRIVATE KEY-----
```
alexmwu: this should go in a file, perhaps under something like `verifier/fake/signer_rsa`

launcher/image/testworkloads/customtoken/happypath/main.go
```diff
@@ -217,6 +91,27 @@ func decodeAndValidateToken(tokenBytes []byte, keyFunc func(t *jwt.Token) (any,
...
 
+func getTestRSAPublicKey(token *jwt.Token) (any, error) {
+	// Always return the same hardcoded public key.
+
+	// Verify the signing method
+	if _, ok := token.Method.(*jwt.SigningMethodRSA); !ok {
+		return nil, fmt.Errorf("unexpected signing method: %v", token.Header["alg"])
+	}
+
+	const publicKeyPEM = `-----BEGIN PUBLIC KEY-----
```
alexmwu: this should go in a file such as `verifier/fake/signer_rsa.pub`

verifier/fake/fakeverifier.go
```diff
@@ -34,6 +34,10 @@ func NewClient(signer crypto.Signer) verifier.Client {
 	nonce := make([]byte, 2)
 	binary.LittleEndian.PutUint16(nonce, 15)
 
+	if signer == nil {
```
alexmwu, line 38: you should also document in NewClient if the signer param is nil, we will use the keys in the `verifier/fake directory`

launcher/spec/launch_spec.go
```diff
@@ -141,6 +143,9 @@ func (s *LaunchSpec) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
+	//
+	s.FakeVerifierEnabled, _ = strconv.ParseBool(unmarshaledMap[fakeVerifierKey])
```
jessieqliu: Why not check the error here? In case the string value doesn't parse into bool.

jessieqliu (approved):
- The title is a little inaccurate since the fake verifier already exists. Please be more specific (ex. "Use fake verifier for end-to-end testing")
- Comments are marked resolved without the accompanying changes - do you need to push the latest version of code?

launcher/spec/launch_spec.go
mayafleischer: if fakeVerifierKey not present, unmarshaledMap[fakeVerifierKey] returns "".   If an empty string is passed into ParseBool, it just returns false so I thought was okay.

I guess could add

`	if val, ok := unmarshaledMap[fakeVerifierKey]; ok && val != "" {
		var err error
		if s.FakeVerifierEnabled, err = strconv.ParseBool(val); err != nil {
			return fmt.Errorf("invalid value for %v (not a boolean): %w", fakeVerifierKey, err)
		}
	}
	`
if that is better?

jessieqliu: The issue is not with the map lookup (although checking if the key exists would be better for consistency with the rest of this function) but that `strconv.ParseBool` only [parses correctly](https://pkg.go.dev/strconv#ParseBool) for a specific set of values and otherwise returns an error. The current code doesn't check the `err` returned by `ParseBool` at all, so the behavior can easily be undefined/unexpected if the operator sets any unsupported value for that launchspec field.

The code snippet above looks good.

verifier/fake/fakeverifier.go
mayafleischer: I know this should be simple to do, but I am getting errors opening the file when I run the tests.  Gemini is suggesting I use `embed`, but I don't think this would work in happypath/main.go anyway.  Am I missing something simple here?

alexmwu: One thing you could try is embedding and exporting in something like `verifier/fake/data.go`. Then add the verifier module to your go.mod in the happypath test directory

PR #620, opened 2025-12-09
Add TDX CCEL support to token command and refactor flags

yanchengchen7:
* Move `teeTechnology` flag to flags.go to share it between `attest` and `token` commands.
* Update `token` command to support TDX CCEL attestation by reading CCEL data and ACPI tables.
* Populate the `GceInstance` field in `VerifyAttestationRequest` using MDS data for the `token` command.

Example usage: `gotpm token --tee-technology tdx`


cmd/token.go
```diff
@@ -134,13 +138,46 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 		}
 		ak.Close()
 
+		// If teeTechnology is not set, try to detect it from the attestation.
+		if teeTechnology == "" {
+			if attestation.GetTdxAttestation() != nil {
+				teeTechnology = Tdx
+			}
+		}
+
```
alexmwu: I think we should remove the detection. The caller should specify the TEE attestation to verify, just like with the gotpm attest command

alexmwu (approved):

cmd/token.go
```diff
@@ -128,7 +132,33 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
...
+		case "":
+			if len(teeNonce) != 0 {
+				return fmt.Errorf("use of --tee-nonce requires specifying TEE hardware type with --tee-technology")
+			}
+		default:
+			// Change the return statement when more devices are added
+			return fmt.Errorf("tee-technology should be either empty or should have values %s or %s", SevSnp, Tdx)
+		}
+
+		attestation, err := ak.Attest(attestOpts)
```
yawangwang, line 154: Currently there's no binding b/w TPM and TDX attestation, so it's premature to enable tee-nonce and tee-technology flags for the token command.

alexmwu: This is duplicated with https://github.com/google/go-tpm-tools/blob/main/cmd/attest.go#L84-L105. Please refactor into a helper like `func addTEEOpts(opts client.AttestOpts) error`

cmd/token.go
```diff
@@ -141,6 +171,32 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
...
+				ccelTable, _ := os.ReadFile(internal.AcpiTableFile)
+				ccelData, _ := os.ReadFile(internal.CcelEventLogFile)
+
+				req.TDCCELAttestation = &verifier.TDCCELAttestation{
+					TdQuote:           rawQuote,
+					CcelAcpiTable:     ccelTable,
+					CcelData:          ccelData,
+					AkCert:            attestation.AkCert,
+					IntermediateCerts: attestation.IntermediateCerts,
+				}
```
yawangwang, line 218: This enables sending TDX CVM attestation to GCA for verification. However, since TDX CVM support is not yet GA, users will likely encounter errors when using the CLI. 

This PR should be held until TDX CVM and hardware binding reach GA.

agent/agent.go
```diff
@@ -539,7 +538,7 @@ func (t *tdxAttestRoot) Attest(nonce []byte) (any, error) {
 	// This naive trimming logic may cause the replay to fail.
 	ccelData = bytes.TrimRight(ccelData, "\xff")
 
-	ccelTable, err := os.ReadFile("/sys/firmware/acpi/tables/CCEL")
+	ccelTable, err := os.ReadFile(internal.AcpiTableFile)
```
alexmwu, line 541: This is not useful. Can we disable the check on GCA? https://github.com/google/go-eventlog/blob/5cd85087f9f9ddf6e02330fb549879cf5ff54b06/extract/extract.go#L69

alexmwu (approved):

internal/ccel.go
```diff
@@ -0,0 +1,8 @@
+package internal
+
+const (
+	// AcpiTableFile is the path to the CCEL ACPI table.
+	AcpiTableFile = "/sys/firmware/acpi/tables/CCEL"
+	// CcelEventLogFile is the path to the CCEL event log data.
+	CcelEventLogFile = "/sys/firmware/acpi/tables/data/CCEL"
```
susienme: I think the convention is to capitalize acronyms like ACPI and CCEL (go/go-style/decisions#initialisms).

cmd/attest.go
```diff
@@ -16,16 +16,7 @@ import (
...
-	teeTechnology string
-)
-
-// Add constants for other devices when required
-const (
-	// SevSnp is a constant denotes device name for teeTechnology
-	SevSnp = "sev-snp"
-	// Tdx is a constant denotes device name for teeTechnology
-	Tdx = "tdx"
+	key string
```
susienme: I don't think the variable block with `()` is unnecessary because we have a single variable left.

cmd/token.go
```diff
@@ -134,7 +153,16 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 		if err != nil {
 			return fmt.Errorf("failed to get an AK: %w", err)
 		}
-		attestation, err := ak.Attest(client.AttestOpts{Nonce: challenge.Nonce, CertChainFetcher: http.DefaultClient})
+
+		attestOpts := client.AttestOpts{Nonce: challenge.Nonce, CertChainFetcher: http.DefaultClient}
+
+		// Add logic to open other hardware devices when required.
```
susienme: Is this correct comment?

cmd/attest.go
```diff
@@ -82,26 +73,12 @@ hardware and guarantees a fresh quote.
 		attestOpts.Nonce = nonce
 
 		// Add logic to open other hardware devices when required.
```
susienme: Is this comment correct?

cmd/token.go
```diff
@@ -145,6 +173,33 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
...
+				// Try to read CCEL Table and Data
+				ccelTable, _ := os.ReadFile(internal.AcpiTableFile)
+				ccelData, _ := os.ReadFile(internal.CcelEventLogFile)
+
+				req.TDCCELAttestation = &verifier.TDCCELAttestation{
+					TdQuote:           rawQuote,
+					CcelAcpiTable:     ccelTable,
+					CcelData:          ccelData,
+					AkCert:            attestation.AkCert,
+					IntermediateCerts: attestation.IntermediateCerts,
```
yawangwang: You can simply remove these two fields for TDX CVM.

cmd/token.go
```diff
@@ -145,6 +173,33 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 			GcpCredentials: principalTokens,
 			Attestation:    attestation,
 			TokenOptions:   &models.TokenOptions{Audience: audience, Nonces: customNonce, TokenType: "OIDC"},
+			GceInstance:    fmt.Sprintf("projects/%s/zones/%s/instances/%s", projectNumber, zone, instanceID),
```
yawangwang: This field should only be set for TDX CVM, not ideal to make it as a default option.

cmd/flags.go
```diff
@@ -319,3 +332,29 @@ func getEK(rwc io.ReadWriter) (*client.Key, error) {
...
 	}
 }
+
+// getTEEDevice based on teeTechnology set in the global flag vars.
+func getTEEDevice() (client.TEEDevice, error) {
+	switch teeTechnology {
+	case SevSnp:
+		device, err := client.CreateSevSnpQuoteProvider()
+		if err != nil {
+			return nil, fmt.Errorf("failed to open %s device: %v", SevSnp, err)
```
susienme: Why is the error message different from the one below? I think `create` is correct?

cmd/token.go
```diff
@@ -134,7 +153,16 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 		if err != nil {
 			return fmt.Errorf("failed to get an AK: %w", err)
 		}
-		attestation, err := ak.Attest(client.AttestOpts{Nonce: challenge.Nonce, CertChainFetcher: http.DefaultClient})
+
+		attestOpts := client.AttestOpts{Nonce: challenge.Nonce, CertChainFetcher: http.DefaultClient}
+
+		// Add logic to open other hardware devices when required.
+		attestOpts.TEEDevice, err = getTEEDevice()
```
susienme, line 149: You don't need to assign teeNonce like in attest.go?

cmd/token.go
```diff
@@ -145,6 +173,33 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
...
+			if attestation.GetTdxAttestation() != nil {
+				fmt.Fprintln(debugOutput(), "Using Explicit TDCCELAttestation Path (ACPI tables)")
+
+				rawQuote, err := tabi.QuoteToAbiBytes(attestation.GetTdxAttestation())
+				if err != nil {
+					return fmt.Errorf("failed to convert TDX quote to bytes: %v", err)
+				}
+
+				// Try to read CCEL Table and Data
+				ccelTable, _ := os.ReadFile(internal.AcpiTableFile)
```
susienme: The inline comment implies it's acceptable for these reads to fail. But, consider logging the errors to `debugOutput()`.

cmd/token.go
yanchengchen7: Done. I removed the automatic TEE detection. The caller must now explicitly select the TEE using `--tee-technology`, consistent with the `attest` command.

cmd/token.go
yanchengchen7: Done. I moved the shared TEE device selection and validation logic into `getTEEDevice()`, which is now used by both `attest` and `token`. The nonce assignment remains command-specific because `token` intentionally uses the verifier challenge.

cmd/token.go
yanchengchen7: Done. I removed `AkCert` and `IntermediateCerts` from the TDX CVM request construction.

cmd/token.go
yanchengchen7: Done. `GceInstance` is now populated only inside the TDX CVM/CCEL path.

internal/ccel.go
yanchengchen7: Done. Renamed.

cmd/attest.go
yanchengchen7: Done.

cmd/token.go
yanchengchen7: Good catch. Changed it to: “Open the explicitly requested TEE device.”

cmd/attest.go
yanchengchen7: Updated this comment as well.

cmd/flags.go
yanchengchen7: Done.

cmd/token.go
yanchengchen7: Done.

cmd/token.go
yanchengchen7: This is intentionally left unset. The `token` command does not expose `--tee-nonce`. When `TEENonce` is empty, the TDX quote provider uses `attestOpts.Nonce`, which contains the verifier-issued challenge nonce.

cmd/token.go
yanchengchen7: tee-nonce is not exposed in token command now.

agent/agent.go
yanchengchen7: GCA disabled the check now.

cmd/flags.go
```diff
@@ -15,19 +15,28 @@ import (
...
+	cloudLog      bool
+	customNonce   []string
+)
+
+// Add constants for other devices when required
+const (
+	// SevSnp is a constant denotes device name for teeTechnology
+	SevSnp = "sev-snp"
+	// Tdx is a constant denotes device name for teeTechnology
+	Tdx = "tdx"
```
susienme: I think these two constants are not exposed? Then, `sevSNP` and `tdx`?
If exposed, `SEVSNP` and `TDX`
go/go-style/decisions#initialisms

verifier/rest/rest_test.go
```diff
@@ -46,6 +46,47 @@ func TestConvertEmpty(t *testing.T) {
 	}
 }
 
+func TestConvertRequestToREST(t *testing.T) {
```
susienme, line 49: Consider following the table-driven test model.

verifier/client.go
```diff
@@ -50,6 +50,7 @@ type VerifyAttestationRequest struct {
 	// TDCCELAttestation is for TDX CCEL RTMR attestation
 	TDCCELAttestation *TDCCELAttestation
 	NvidiaAttestation *attestationpb.NvidiaAttestationReport
+	GceInstance       string
```
susienme: s/GceInstance/GCEInstance
go/go-style/decisions#initialisms

susienme (approved):

agent/agent.go
```diff
@@ -639,11 +639,10 @@ func (t *tdxAttestRoot) Attest(nonce []byte) (any, error) {
 		return nil, err
 	}
 
-	ccelData, err := os.ReadFile("/sys/firmware/acpi/tables/data/CCEL")
```
yawangwang, line 642: Let's keep this path as-is w/o moving to internal package since `agent` now working as a standalone module is also referenced in a separate repo `KPS`. The goal of `agent` is to make it as independent as much as possible w/o adding unnecessary dependencies.

yawangwang (approved):

agent/agent.go
yanchengchen7: I see, reverted back to the strings.

PR #627, opened 2026-01-14
Improve error message for unsupported CVM types

yanchengchen7:
This PR improves the error handling for Intel TDX and AMD SEV-SNP attestation failures in the gotpm token command, returns a user-friendly message explaining that the Google Cloud Attestation service does not support token issuance for SEV-SNP or TDX VMs.

### DIFF
--- a/cmd/token.go
+++ b/cmd/token.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"log"
 	"net/http"
-	"strings"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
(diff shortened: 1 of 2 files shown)


cmd/token.go
```diff
@@ -143,7 +144,10 @@ The OIDC token includes claims regarding the GCE VM, which is verified by Attest
 
 		resp, err := verifierClient.VerifyAttestation(ctx, req)
 		if err != nil {
-			return err
+			if strings.Contains(err.Error(), "unexpected_tdx_ccel_attestation") || strings.Contains(err.Error(), "unexpected_snp_attestation") {
+				return fmt.Errorf("verification failed: Google Cloud Attestation service does not support token issuance for Confidential VMs using Intel TDX or AMD SEV-SNP. This service currently only supports VMs using AMD SEV for token generation. To verify the attestation of this VM, please use an alternative method: 1. Verify Locally with go-tpm-tools: Instead of requesting a GCA token, fetch and validate the raw attestation quote. 2. Use Google Cloud Confidential Space: This product has built-in attestation capabilities suitable for these VM types: https://docs.cloud.google.com/confidential-computing/confidential-space/ 3. Use another verifier service")
```
jessieqliu: nit: This is pretty long for an error - I think something like Rene's suggestion in the original bug would be better for readability.

yanchengchen7: Sounds good, I've updated the error message to be more concise.

PR #629, opened 2026-01-17
[launcher] update base cos image to 125

jkl73:
Change rtmr interface to sysfs.

### DIFF
--- a/cel/canonical_eventlog.go
+++ b/cel/canonical_eventlog.go
@@ -139,9 +139,7 @@ func generateDigestMap(hashAlgos []crypto.Hash, event Content) (map[crypto.Hash]
 	return digestsMap, nil
 }
 
-// AppendEventRTMR appends a new RTMR record to the CEL. rtmrIndex indicates the RTMR to extend.
-// The index showing up in the record will be rtmrIndex + 1.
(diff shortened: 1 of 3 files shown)


cel/canonical_eventlog.go
```diff
@@ -169,9 +166,28 @@ func (c *CEL) AppendEventRTMR(client configfsi.Client, rtmrIndex int, event Cont
 	return nil
 }
 
+// AppendEventRTMRSysfs appends a new RTMR record to the CEL. rtmrIndex indicates the RTMR to extend.
+// The index showing up in the record will be rtmrIndex + 1.
+// This function uses thesysfs interface to extend the RTMR.
+func (c *CEL) AppendEventRTMRSysfs(rtmrIndex int, event Content) error {
```
alexmwu, line 172: This is a bit smelly. consider moving to go-eventlog's cel package and using that AppendEvent which makes the implementation of extension an interface (which gives a loose coupling where the caller defines how the extension happens).

jkl73: migrate canonical eventlog to go-eventlog requires some changes, will do it in next PR

PR #631, opened 2026-01-28
Add VerifyOpts flag to force a specific hash algo

alexmwu:
VerifyAttestation today looks through all of its supported hash algorithms to verify the incoming quote signatures. This has a few drawbacks:
- the verifier does not specify its preferred algorithm and so a malicious attester can downgrade the hash algorithm supported to a weaker algorithm by stripping the quotes.
- when we add support for now algorithms in any underlying libraries, this can break existing verifiers that rely on the implicit algorithm selection https://github.com/google/go-tpm-tools/pull/591 specifically introduced SHA384 support for TPM event log parsing. this caused breakages to verifiers that relied on extracted digests in the MachineState being SHA256 hashes

### DIFF
--- a/server/verify.go
+++ b/server/verify.go
@@ -7,6 +7,7 @@ import (
 	"encoding/asn1"
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/google/go-tpm-tools/internal"
 	pb "github.com/google/go-tpm-tools/proto/attest"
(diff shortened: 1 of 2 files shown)


yawangwang (approved):

server/verify.go
```diff
@@ -313,10 +333,17 @@ func VerifyAKCert(akCert *x509.Certificate, trustedRootCerts []*x509.Certificate
 	return nil
 }
 
-// Retrieve the supported quotes in order of hash preference.
-func supportedQuotes(quotes []*tpmpb.Quote) []*tpmpb.Quote {
+func supportedAlgo(algo tpm2.Algorithm) error {
+	if slices.Contains(pcrHashAlgs, algo) {
+		return nil
+	}
+	return fmt.Errorf("hash algorithm %s is not supported", algo)
```
jkl73: also print the pcrHashAlgs for better debug

server/verify.go
```diff
@@ -70,6 +76,10 @@ type VerifyOpts struct {
 	// to length 1 (0 or 1) during eventlog parsing.
 	// This can be used when the SecureBoot variable is not initialized.
 	AllowEmptySBVar bool
+	// The TPM hash algorithm to use to verify attestations.
+	// SHA512, SHA384, SHA256, and SHA1 are supported. To use SHA1, also set
+	// AllowSHA1.
```
jkl73, line 81: // By default (tpm2.AlgUnknown), will try all hashes in the order (SHA512 > SHA384 > SHA256 > [SHA1]) until one quote + eventlog can be verified.

PR #633, opened 2026-01-31
[KCC] Add generate_kem_keypair and generate_binding_keypair FFIs

meetrajvala:


kcc/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+                algo: HpkeAlgorithm {
+                    kem: KemAlgorithm::DhKemX25519HkdfSha256,
+                    kdf: KdfAlgorithm::HkdfSha256,
+                    aead: AeadAlgorithm::Aes256Gcm,
+                },
+                binding_public_key: vec![0u8; 32],
+            },
+        };
+
+        let mut test_key = vec![0u8; 32];
```
atulpatildbz, line 238: unused

kcc/src/types.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub created_at: Instant,
+    pub delete_after: Option<Instant>,
+    pub spec: PublicSpec,
+}
+
+#[derive(Debug)]
+pub enum PublicSpec {
+    KemWithBindingPub { algo: HpkeAlgorithm, kem_public_key: Vec<u8>, binding_public_key: Vec<u8> },
+    Binding { algo: HpkeAlgorithm, binding_public_key: Vec<u8> },
+    Sek { algo: SigningAlgorithm, verifying_key: Vec<u8> },
```
atulpatildbz, line 17: not needed currently.

kcc/src/vault.rs
```diff
@@ -0,0 +1,72 @@
...
+
+// From <linux/memfd.h>
+const MFD_CLOEXEC: u32 = 0x0001;
+const MFD_SECRET: u32 = 0x0008;
+
+/// The Owner of the Backing Store (The "Box")
+#[derive(ZeroizeOnDrop)]
+pub struct Vault {
+    mmap: MmapMut,
+}
```
atulpatildbz, line 16: `MmapMut` doesn't implement `Zeroize` it seems. https://docs.rs/memmap2/latest/memmap2/struct.MmapMut.html#trait-implementations

Implement `Drop` manually to zero the mmap contents via `self.mmap[..].zeroize()` before unmapping.

kcc/src/vault.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl Vault {
+    pub fn new(name: &str, data: &[u8]) -> Result<Self, IoError> {
+        let name_cstr = CString::new(name)
+            .map_err(|_| IoError::new(std::io::ErrorKind::InvalidInput, "Invalid name"))?;
+        let fd = unsafe {
+            libc::syscall(libc::SYS_memfd_create, name_cstr.as_ptr(), MFD_CLOEXEC | MFD_SECRET)
```
atulpatildbz, line 23: `MFD_SECRET` isn't a valid flag for memfd_create. https://man7.org/linux/man-pages/man2/memfd_create.2.html

1. Call the actual `memfd_secret` syscall . ([man page reference](https://man7.org/linux/man-pages/man2/memfd_secret.2.html))
2. Use `ftruncate` to set the size.
3. Use `mmap` to write the data (since `memfd_secret` fds don't support `write()`).

kcc/src/vault.rs
```diff
@@ -0,0 +1,72 @@
+use memmap2::MmapMut;
+use std::ffi::CString;
+use std::fs::File;
+use std::io::{Error as IoError, Write};
+use std::os::unix::io::{FromRawFd, RawFd};
+use zeroize::ZeroizeOnDrop;
+
+// From <linux/memfd.h>
+const MFD_CLOEXEC: u32 = 0x0001;
+const MFD_SECRET: u32 = 0x0008;
```
atulpatildbz, line 10: in [linux/memfd.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/memfd.h#L12) `0x0008` points to `MFD_NOEXEC_SEAL`

LMK if i'm missing something

kcc/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    fn drop(&mut self) {
+        self.stop_reaper.store(true, Ordering::Relaxed);
+        if let Some(handle) = self.reaper_handle.take() {
+            handle.join().expect("Reaper thread panicked");
+        }
+    }
+}
+
+pub struct KeyManager {
+    pub binding_keys: KeyRegistry,
```
NilanjanDaw, line 85: As discussed today, we need to separate these out into individual Key Management Agents, one for the Workload daemon, and the other for the Key Protection Service.

kcc/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub fn generate_binding_keypair(&self) -> (KeyHandle, Vec<u8>) {
+        clear_stack_on_return(2, || {
+            let key_id = Uuid::new_v4();
+            let (pub_key, mut priv_key) = hpke::Kem::X25519HkdfSha256.generate_keypair();
+            let secret = Vault::new(&key_id.to_string(), &priv_key).expect("Vault creation failed");
+            priv_key.zeroize();
+            let meta = KeyMetadata {
+                id: key_id,
+                created_at: Instant::now(),
+                delete_after: Some(Instant::now() + Duration::from_secs(3600)),
```
NilanjanDaw, line 114: The binding key lifespan is a parameter passed down from the orchestration layer.

kcc/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+            binding_keys: KeyRegistry::new(),
+            kem_keys: KeyRegistry::new(),
+        }
+    }
+
+
+    pub fn generate_binding_keypair(&self) -> (KeyHandle, Vec<u8>) {
+        clear_stack_on_return(2, || {
+            let key_id = Uuid::new_v4();
+            let (pub_key, mut priv_key) = hpke::Kem::X25519HkdfSha256.generate_keypair();
```
NilanjanDaw, line 108: Even though for initial phases, we will exclusively support X25519, we need to make these configurable. The `KemAlgorithm` will be passed as parameter indicating the DHKEM algorithm to use for the request.

kcc/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+            let meta = KeyMetadata {
+                id: key_id,
+                created_at: Instant::now(),
+                delete_after: Some(Instant::now() + Duration::from_secs(3600)),
+                spec: PublicSpec::Binding {
+                    algo: HpkeAlgorithm {
+                        kem: KemAlgorithm::DhKemX25519HkdfSha256,
+                        kdf: KdfAlgorithm::HkdfSha256,
+                        aead: AeadAlgorithm::Aes256Gcm,
+                    },
```
NilanjanDaw, line 120: Again the exact algorithm chain should be passed from the orchestration layer and not hardcoded here.

kcc/src/types.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
```
NilanjanDaw, line 19: I would prefer defining the Algorithms below as a common protobuf definition and not in rust. And then use Prost to have a Rust binding for those protos. That would allow us to have a common definition between both the orchestration and the custody core components.

kcc/src/types.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u32)]
+pub enum KemAlgorithm {
+    DhKemX25519HkdfSha256 = 1,
```
NilanjanDaw, line 23: For each of these, we should have a default unknown definition, like `KEM_ALGORITHM_UNSPECIFIED` to prevent accidental values from being set.

kcc/src/vault.rs
NilanjanDaw: Yeah I think I wrote `MFD_SECRET` in my doc. It should be `MFD_NOEXEC_SEAL` instead. We should add it to prevent this memory from ever being executable.

kcc/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+            };
+            let record = KeyRecord { meta, secret };
+            let handle = self.binding_keys.add_key(record);
+            (handle, pub_key)
+        })
+    }
+
+    pub fn generate_kem_keypair(&self, binding_pub_key: &[u8]) -> (KeyHandle, Vec<u8>) {
+        clear_stack_on_return(2, || {
+            let key_id = Uuid::new_v4();
```
NilanjanDaw, line 132: The KEM keys and the Binding keys will share the same UUID. When creating the binding keys, a new UUID will be generated and returned back to the orchestration layer. This same UUID will be passed back to the `generate_kem_keypair` FFI. So when the KEM keys are generated, these keys are mapped to the same UUID and not a new one. This way we can look up the correct KEM key <-> Binding Key pair during the "decap and seal" and subsequent "open" operation.

kcc/src/error.rs
```diff
@@ -0,0 +1,14 @@
+
+use thiserror::Error;
+
```
NilanjanDaw, line 3: I think we should move this to a protobuf file, to ensure a single definition across the orchestration and custody core layers.

kcc/src/vault.rs
NilanjanDaw: Regarding using `ftruncate`, while it is a valid method. A memory `write()` should have similar effects. As the [man-page](https://man7.org/linux/man-pages/man2/memfd_create.2.html) mentions:
> The initial size of the file is set to 0.  Following the call, the
       file size should be set using [ftruncate(2)](https://man7.org/linux/man-pages/man2/ftruncate.2.html).  (Alternatively, the
       file may be populated by calls to [write(2)](https://man7.org/linux/man-pages/man2/write.2.html) or similar.)

So I think the `file.write_all(data)?;` below should be good enough to set the file size.

Regarding `write()` not being supported, I am not sure why that will be. Unless we add `F_SEAL_WRITE` we should be able to write to the `fd` location similar to `memfd` backed files.

kcc/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub fn new() -> Self {
+        let keys = Arc::new(RwLock::new(HashMap::<KeyHandle, KeyRecord>::new()));
+        let stop_reaper = Arc::new(AtomicBool::new(false));
+
+        let keys_clone = keys.clone();
+        let stop_clone = stop_reaper.clone();
+
+        let reaper_handle = Some(thread::spawn(move || {
+            while !stop_clone.load(Ordering::Relaxed) {
+                thread::sleep(Duration::from_secs(10));
```
NilanjanDaw, line 45: Let's make this configurable?

kcc/src/lib.rs
meetrajvala: It is being used on line 241.

kcc/src/types.rs
meetrajvala: Removed.

PR #634, opened 2026-01-31
[KCC] implement destroy_key_pair FFI

meetrajvala:


kcc/src/ffi.rs
```diff
@@ -82,6 +82,30 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+/// * `key_handle_ptr` - Pointer to a 16-byte buffer containing the key handle.
+///
+/// # Returns
+/// 0 on success, -1 on failure (e.g., key not found).
+///
+/// # Safety
+/// Assumes `key_handle_ptr` is a valid pointer to 16 bytes.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_destroy_key(key_handle_ptr: *const u8) -> i32 {
+    let handle_bytes = slice::from_raw_parts(key_handle_ptr, 16);
```
NilanjanDaw, line 98: This makes me a bit nervous, probably a good idea to add some checks to see if the `key_handle_ptr` is valid?

kcc/src/lib.rs
```diff
@@ -152,6 +152,16 @@ impl KeyManager {
             (handle, kem_pub)
         })
     }
+
+    pub fn destroy_key(&self, handle: KeyHandle) -> Result<(), Error> {
+        if self.binding_keys.destroy_key(handle).is_ok() {
```
NilanjanDaw, line 157: I don't think this delete function will work. This assumes the the Binding and KEM keys will have unique UUIDs. But that is not the case, each Binding Key, KEM Key pair will share the same UUID. We can probably pass a reference to the key-registry (`binding_keys` or `kem_keys`) to lookup as a parameter.

kcc/src/ffi.rs
```diff
@@ -82,6 +82,30 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+/// # Arguments
+/// * `key_handle_ptr` - Pointer to a 16-byte buffer containing the key handle.
+///
+/// # Returns
+/// 0 on success, -1 on failure (e.g., key not found).
+///
+/// # Safety
+/// Assumes `key_handle_ptr` is a valid pointer to 16 bytes.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_destroy_key(key_handle_ptr: *const u8) -> i32 {
```
NilanjanDaw, line 97: Lets have two separate FFIs, `key_manager_destroy_kem_key` and `key_manager_destroy_binding_key` to disambiguate the `KeyRecord` to delete the keys from.

PR #635, opened 2026-02-03
[launcher] Fix CEL RTMR extending

jkl73:
Add image tests for TDX machines
Add TDX support in the fakeverifier

### DIFF
--- a/launcher/cloudbuild.yaml
+++ b/launcher/cloudbuild.yaml
@@ -134,6 +134,22 @@ steps:
     exit
 
 - name: 'gcr.io/cloud-builders/gcloud'
+  id: DebugImageTestsTDX
+  waitFor: ['DebugImageBuild']
+  env:
+  - 'OUTPUT_IMAGE_PREFIX=${_OUTPUT_IMAGE_PREFIX}'
+  - 'OUTPUT_IMAGE_SUFFIX=${_OUTPUT_IMAGE_SUFFIX}'
+  - 'PROJECT_ID=$PROJECT_ID'
+  script: |
+    #!/usr/bin/env bash
+
+    cd launcher/image/test
+    echo "running debug image tests on ${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX}"
+    gcloud builds submit --config=test_debug_cloudbuild.yaml --region us-west1 \
+      --substitutions _CC=TDX,_IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
+    exit
+
+- name: 'gcr.io/cloud-builders/gcloud'
   id: HardenedImageTests
   waitFor: ['HardenedImageBuild']
   env:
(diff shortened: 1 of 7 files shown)


launcher/image/test/create_vm.sh
```diff
@@ -48,10 +49,20 @@ create_vm() {
   ADDTL_DISK_RANGE=$(($MAX_DISK_SIZE_GB - $MIN_DISK_SIZE + 1))
   DISK_SIZE_GB=$(($MIN_DISK_SIZE + ($RANDOM % $ADDTL_DISK_RANGE)))
 
-  gcloud compute instances create $VM_NAME --confidential-compute --maintenance-policy=TERMINATE \
-    --machine-type=n2d-standard-2 --boot-disk-size=$DISK_SIZE_GB --scopes=cloud-platform --zone $ZONE \
-    --image=$IMAGE_NAME --image-project=$PROJECT_NAME --shielded-secure-boot $APPEND_METADATA \
-    $APPEND_METADATA_FILE
+  if [[ "${CC}" == "sev" ]]; then
```
alexmwu: nit: the checks should happen above and you should still only have one paramaterized command

launcher/cloudbuild.yaml
```diff
@@ -114,7 +114,23 @@ steps:
     cd launcher/image/test
     echo "running http server tests on ${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX}"
     gcloud builds submit --config=test_http_server.yaml --region us-west1 \
-      --substitutions _IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
+      --substitutions _CC=sev,_ZONE=asia-east1-a,_IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
+    exit
+
+- name: 'gcr.io/cloud-builders/gcloud'
```
alexmwu: is this a valuable test for TDX?

launcher/cloudbuild.yaml
```diff
@@ -114,7 +114,23 @@ steps:
     cd launcher/image/test
     echo "running http server tests on ${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX}"
     gcloud builds submit --config=test_http_server.yaml --region us-west1 \
-      --substitutions _IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
+      --substitutions _CC=sev,_ZONE=asia-east1-a,_IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
```
alexmwu: Why are you overriding the CC and ZONE substitution here?

launcher/cloudbuild.yaml
```diff
@@ -130,7 +146,23 @@ steps:
     cd launcher/image/test
     echo "running debug image tests on ${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX}"
     gcloud builds submit --config=test_debug_cloudbuild.yaml --region us-west1 \
-      --substitutions _IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
+      --substitutions _CC=sev,_IMAGE_NAME=${OUTPUT_IMAGE_PREFIX}-debug-${OUTPUT_IMAGE_SUFFIX},_IMAGE_PROJECT=${PROJECT_ID}
```
alexmwu: Here and elsewhere, why are you overriding the CC substitution? It's sev by default, no?

verifier/fake/fakeverifier.go
```diff
@@ -59,33 +63,73 @@ func (fc *fakeClient) VerifyAttestation(_ context.Context, req verifier.VerifyAt
...
+	var ms *attest.MachineState
+	var err error
+	var msJSON []byte
 
-	pcrBank, err := extractPCRBank(req.Attestation, ms.GetHash())
-	if err != nil {
-		return nil, fmt.Errorf("failed to extract PCR bank: %w", err)
-	}
+	if req.TDCCELAttestation != nil {
+		tdQuote, err := tabi.QuoteToProto(req.TDCCELAttestation.TdQuote)
```
alexmwu: these should be functions.

launcher/cloudbuild.yaml
jkl73: not really, should be covered by other tdx test, removed

verifier/fake/fakeverifier.go
jkl73: done

launcher/image/test/create_vm.sh
jkl73: done

launcher/cloudbuild.yaml
jkl73: was added earlier to make it explicit on CC and Zone selection.

revert the overwrite here since don't need to do HttpServerTestsTDX

launcher/cloudbuild.yaml
jkl73: the new substitution (_CC) added in the test cloudbuild yaml is an empty string ''.

So the substitutions needs to be explicit on those tests, otherwise the create_vm.sh will get a empty CC arg (instead of using it's default value)

alexmwu: why can't the create_vm.sh script handle an empty CC arg and use SEV by default? That's the current behavior, where every test runs on SEV

jkl73: I see, added empty arg case for cc

PR #636, opened 2026-02-03
Bootstrap KCC directory

meetrajvala:
KCC directory is having 2 sub directories, (1) `wsd_kcc`: this will have FFIs required for WSD (2) `kps_kcc`: this will have FFIs required for KPS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ files/go-tpm-tools
 main
 go.work
 go.work.sum
+kcc/wsd_kcc/target/*
+kcc/kps_kcc/target/*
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -13,3 +13,7 @@ files/go-tpm-tools
 main
 go.work
 go.work.sum
+kcc/wsd_kcc/target/*
```
alexmwu, line 16: You're checking in flycheck0/stderr, stdout. Is that intentional?

meetrajvala: It was not intentional.removed them.

alexmwu (approved):

PR #640, opened 2026-02-04
[KeyManager] Implement common key structures and proto bindings

NilanjanDaw:
Introduces a new `common` crate within the `keymanager` workspace to house the shared data structures and protobuf definitions.
Key changes:
- Adds `keymanager/km_common` crate with `uuid`, `memmap2`, `zeroize`, and `prost` dependencies.
- Defines core data structures in `key_types.rs`: `KeySpec`, `KeyMetadata`, `Vault`, `KeyRecord`, and `KeyRegistry`.
- Implements protobuf bindings for cryptographic algorithms in `proto/algorithms.proto`.
- Configures `build.rs` to generate Rust bindings from protos using `protoc-bin-vendored`.
- Updates `keymanager/Cargo.toml` to include `km_common` in the workspace.

### DIFF
--- a/keymanager/km_common/build.rs
+++ b/keymanager/km_common/build.rs
@@ -1,6 +1,7 @@
 use std::io::Result;
 
 fn main() -> Result<()> {
+    println!("cargo::rerun-if-changed=proto");
     unsafe {
         std::env::set_var("PROTOC", protoc_bin_vendored::protoc_bin_path().unwrap());
     }
--- a/keymanager/km_common/src/key_types.rs
+++ b/keymanager/km_common/src/key_types.rs
@@ -7,39 +7,48 @@ use uuid::Uuid;
 /// Represents the purpose of the Key and its associated algorithms.
 #[derive(Clone)]
 pub enum KeySpec {
-    // Represents the composite key used by the Key Protection Service for the decaps-and-encrypt flow.
+    /// Represents the composite key used by the Key Protection Service for the decaps-and-encrypt flow.
     KemWithBindingPub {
-        // The KEM and binding public keys share the same algorithm suite.
+        /// The KEM and binding public keys share the same algorithm suite.
         algo: HpkeAlgorithm,
-        kem_public_key: Vec<u8>,     // The KEM public key
-        binding_public_key: Vec<u8>, // Binding public key for HPKE encrypt after decaps
+        /// The KEM public key
+        kem_public_key: Vec<u8>,
+        /// Binding public key for HPKE encrypt after decaps
+        binding_public_key: Vec<u8>,
     },
     Binding {
         algo: HpkeAlgorithm,
-        binding_public_key: Vec<u8>, // The Binding key-pair
+        /// The Binding key-pair
+        binding_public_key: Vec<u8>,
     },
 }
 
-// Internal Rust struct to hold the Key Metadata
+/// Internal Rust struct to hold the Key Metadata
 #[derive(Clone)]
 pub struct KeyMetadata {
-    pub id: Uuid,              // UUID key handle for internal tracking
+    /// UUID key handle for internal tracking
+    pub id: Uuid,
     pub created_at: Instant,
-    pub delete_after: Instant, // TTL-bound deletion time
-    pub spec: KeySpec,         // (non-secret) Cryptographic material
+    /// TTL-bound deletion time
+    pub delete_after: Instant,
+    /// (non-secret) Cryptographic material
+    pub spec: KeySpec,
 }
 
 pub struct Vault {
     // placeholder
 }
 
+/// Internal struct to hold the Key Metadata and the secret key material.
 pub struct KeyRecord {
     pub meta: KeyMetadata,
-    pub private_key: Vault, // memfd_secrets backed secret key-material
+    /// memfd_secrets backed secret key-material
+    pub private_key: Vault,
 }
 
 pub type KeyHandle = Uuid;
 
+/// Thread-safe registry for storing encryption keys.
 #[derive(Default, Clone)]
 pub struct KeyRegistry {
     keys: Arc<RwLock<HashMap<KeyHandle, KeyRecord>>>,


keymanager/common/src/key_types.rs
```diff
@@ -0,0 +1,50 @@
...
+pub struct KeyMetadata {
+    pub id: Uuid,              // UUID key handle for internal tracking
+    pub created_at: Instant,
+    pub delete_after: Instant, // TTL-bound deletion time
+    pub spec: KeySpec,         // (non-secret) Cryptographic material
+}
+
+#[derive(ZeroizeOnDrop, Debug)] // zerioize components before dropping
+pub struct Vault {
+    pub mmap: MmapMut,
```
atulpatildbz: Do consumers need raw mutable access? i guess not. if so we can keep this private.

if we need to expose it as bytes 
```rs
pub struct Vault {
    mmap: MmapMut, // Private!
}

impl Vault {
    pub fn as_bytes(&self) -> &[u8] {
        &self.mmap
    }
}
```

keymanager/common/src/key_types.rs
```diff
@@ -0,0 +1,50 @@
...
+// Internal Rust struct to hold the Key Metadata
+#[derive(Debug, Clone)]
+pub struct KeyMetadata {
+    pub id: Uuid,              // UUID key handle for internal tracking
+    pub created_at: Instant,
+    pub delete_after: Instant, // TTL-bound deletion time
+    pub spec: KeySpec,         // (non-secret) Cryptographic material
+}
+
+#[derive(ZeroizeOnDrop, Debug)] // zerioize components before dropping
```
atulpatildbz: `MmapMut` doesn't actually implement the `Zeroize` trait. https://docs.rs/memmap2/latest/memmap2/struct.MmapMut.html#trait-implementations

The reason this compiles and works is because Rust automatically "dereferences" `MmapMut` into a slice `[u8]` (via the `DerefMut` trait), and slices do support `zeroize()`

If the `memmap2` library ever changes how it handles dereferencing in a future version, this could silently stop scrubbing memory without breaking the build.

I'd suggest implementing `Drop` manually to ensure the behavior we want.

```rs
impl Drop for Vault {
    fn drop(&mut self) {
        self.mmap[..].zeroize();
    }
}
```

keymanager/common/proto/algorithms.proto
```diff
@@ -0,0 +1,28 @@
...
+// Key Derivation Function (KDF)
+enum KdfAlgorithm {
+  KDF_ALGORITHM_UNSPECIFIED = 0;
+  KDF_ALGORITHM_HKDF_SHA256 = 1;
+}
+
+// Authenticated Encryption with Associated Data (AEAD)
+enum AeadAlgorithm {
+  AEAD_ALGORITHM_UNSPECIFIED = 0;
+  AEAD_ALGORITHM_AES_256_GCM = 2;
```
atulpatildbz: should be `1`?

keymanager/km_common/Cargo.toml
atulpatildbz, line 1: In a Rust workspace, every crate must have a unique name. You cannot have two crates named common in the same dependency graph, even if they are in different directories.

Should we rename this dir to something like km_common?

keymanager/common/src/key_types.rs
atulpatildbz: typo zerioize -> zeroize

atulpatildbz (commented):
Thanks for making this change

keymanager/common/proto/algorithms.proto
NilanjanDaw: Good catch. Fixed.

keymanager/common/src/key_types.rs
NilanjanDaw: Oh I was not aware of this. The mmap will be moved to a separate location, see my comment below. I will handle the explicit zeroize on drop there.

NilanjanDaw: Removing the entire section for now.

keymanager/common/src/key_types.rs
NilanjanDaw: You're correct. They don't, this was a placeholder I was testing with. The memfd_secret implementation will have more details on how the mmap will be used.

I will remove these for now.

keymanager/km_common/Cargo.toml
NilanjanDaw: Good point, we should leave the generic name for top level directories. I will update this to `km_common`

keymanager/km_common/src/key_types.rs
```diff
@@ -0,0 +1,48 @@
...
+use uuid::Uuid;
+
+/// Represents the purpose of the Key and its associated algorithms.
+#[derive(Debug, Clone)]
+pub enum KeySpec {
+    // Represents the composite key used by the Key Protection Service for the decaps-and-encrypt flow.
+    KemWithBindingPub {
+        // The KEM and binding public keys share the same algorithm suite.
+        algo: HpkeAlgorithm,
+        kem_public_key: Vec<u8>,     // The KEM key-pair
```
atulpatildbz: the comment should likely say "The KEM public key"

keymanager/km_common/src/key_types.rs
```diff
@@ -0,0 +1,48 @@
...
+// Internal Rust struct to hold the Key Metadata
+#[derive(Debug, Clone)]
+pub struct KeyMetadata {
+    pub id: Uuid,              // UUID key handle for internal tracking
+    pub created_at: Instant,
+    pub delete_after: Instant, // TTL-bound deletion time
+    pub spec: KeySpec,         // (non-secret) Cryptographic material
+}
+
+#[derive(Debug)] // zerioize components before dropping
```
atulpatildbz: Will this print raw secret bytes to logs if anyone debug-prints a `KeyRecord`?

keymanager/km_common/src/key_types.rs
```diff
@@ -0,0 +1,48 @@
...
+    // placeholder
+}
+
+#[derive(Debug)]
+pub struct KeyRecord {
+    pub meta: KeyMetadata,
+    pub private_key: Vault, // memfd_secrets backed secret key-material
+}
+
+pub type KeyHandle = Uuid;
```
atulpatildbz, line 49: instead of alias can we use a tuple struct?

```rs
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub struct KeyHandle(pub Uuid);
```

With alias `Arc<RwLock<HashMap<KeyHandle, KeyRecord>>>,` is same as `Arc<RwLock<HashMap<Uuid, KeyRecord>>>,`

NilanjanDaw: The Key Management System has a notion of KeyHandles through which the upper layers will track the Key IDs. This is just to use the same naming convention across all components.

keymanager/km_common/src/key_types.rs
NilanjanDaw: Hmm that is possible. I have removed all `Debug` across all key material structs as a precaution.

keymanager/km_common/src/key_types.rs
NilanjanDaw: Ah my bad. Updated thanks.

keymanager/km_common/build.rs
```diff
@@ -0,0 +1,9 @@
+use std::io::Result;
+
+fn main() -> Result<()> {
+    unsafe {
```
zhalvorsen, line 5: Its probably worthwhile to include a line like this so it only re-generates the proto *.rs files when they change. It can make for more efficient build times.

```Rust
println!("cargo::rerun-if-changed=proto");
```

keymanager/km_common/src/key_types.rs
```diff
@@ -0,0 +1,46 @@
+use crate::algorithms::HpkeAlgorithm;
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
+use std::time::Instant;
+use uuid::Uuid;
+
+/// Represents the purpose of the Key and its associated algorithms.
+#[derive(Clone)]
+pub enum KeySpec {
+    // Represents the composite key used by the Key Protection Service for the decaps-and-encrypt flow.
```
zhalvorsen: As a general comment, I like to use three slashes (`///`) for all struct and parameter comments. This tells the compiler that it is a doc comment so it shows up while hovering or during auto-completion.

keymanager/km_common/build.rs
NilanjanDaw: Thanks for the suggest. Added.

keymanager/km_common/src/key_types.rs
NilanjanDaw: That makes sense. Updated the `//` comments to `///` doc-comments.

PR #642, opened 2026-02-05
[KeyManager] Implement memfd_secret based Vault for secure key storage

NilanjanDaw:
Implements a secure Vault struct in `keymanager/common` that stores sensitive key material in a memory-backed file created via `memfd_secrets`.

Key Changes:
    - **Sealed Storage**: Uses `memfd_create` with `MFD_CLOEXEC` and `MFD_NOEXEC_SEAL` to create an anonymous file in RAM.
    - **Secure Cleanup**: Implements ZeroizeOnDrop (via the zeroize crate) to ensure the memory map is cleared when the Vault is dropped.
    - **Restricted Access**: The underlying file descriptor is dropped immediately after mapping (except in test builds), ensuring no direct file access remains.

### DIFF
--- a/keymanager/km_common/src/protected_mem.rs
+++ b/keymanager/km_common/src/protected_mem.rs
@@ -12,6 +12,7 @@ use std::os::unix::io::{FromRawFd, RawFd};
 use zeroize::Zeroize;
 
 /// The system call number for `memfd_secret` on x86_64.
+#[cfg(target_arch = "x86_64")]
 const SYS_MEMFD_SECRET: i64 = 447;
 
 /// A secure container for sensitive data backed by `memfd_secret`.
@@ -42,11 +43,14 @@ impl Drop for Vault {
 impl Vault {
     /// Creates a new `Vault` containing the provided data.
     ///
+    /// The provided `data` is copied into the vault and then zeroized in-place
+    /// to ensure the original sensitive data is cleared from memory.
+    ///
     /// # Errors
     ///
     /// Returns an error if the `memfd_secret` syscall fails or if the memory
     /// cannot be mapped.
-    pub fn new(data: &[u8]) -> IoResult<Self> {
+    pub fn new(data: &mut [u8]) -> IoResult<Self> {
         // Create the secret memory file descriptor.
         // O_CLOEXEC ensures the FD is closed on exec.
         let fd = unsafe { libc::syscall(SYS_MEMFD_SECRET, libc::O_CLOEXEC as libc::c_long) };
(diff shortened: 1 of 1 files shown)


keymanager/km_common/src/protected_mem.rs
```diff
@@ -0,0 +1,191 @@
...
+        let file = unsafe { File::from_raw_fd(fd) };
+
+        // Set the size of the secret memory region.
+        file.set_len(data.len() as u64)?;
+
+        // Map the secret memory region into the process's address space.
+        let mut mmap = unsafe { MmapMut::map_mut(&file)? };
+
+        // Copy the sensitive data into the secure region.
+        if !data.is_empty() {
```
bluegate010: Nit: is this check necessary?

NilanjanDaw: yeah I don't think its required, since its a newly created memory location. Removed.

keymanager/km_common/src/protected_mem.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+impl Vault {
+    /// Creates a new `Vault` containing the provided data.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the `memfd_secret` syscall fails or if the memory
+    /// cannot be mapped.
+    pub fn new(data: &[u8]) -> IoResult<Self> {
```
zhalvorsen: It would be good to either take a mutable reference. Then you can zeroize the memory after copying it to the `Vault`. Alternatively you could take a `Box` (like a `unique_ptr` in C++) by value so you can be sure it hasn't been copied around externally by accident.

zhalvorsen (approved):

keymanager/km_common/src/protected_mem.rs
```diff
@@ -0,0 +1,197 @@
...
+//! location for sensitive cryptographic material.
+
+use memmap2::MmapMut;
+use std::fs::File;
+use std::io::{Error as IoError, Result as IoResult};
+use std::os::unix::io::{FromRawFd, RawFd};
+use zeroize::Zeroize;
+
+/// The system call number for `memfd_secret` on x86_64.
+const SYS_MEMFD_SECRET: i64 = 447;
```
atulpatildbz, line 16: can we add `#[cfg(target_arch = "x86_64")]` to enforce x86_64 target during compilation?

NilanjanDaw: Ack. Added.

keymanager/km_common/src/protected_mem.rs
NilanjanDaw: Makes sense, added a `mut` reference and an explicit zeroize. Thanks for pointing this out.

PR #643, opened 2026-02-05
[KeyManager] Add HPKE and DHKEM crypto primitives

NilanjanDaw:
Introduces a `crypto` module in `km_common` to handle cryptographic operations backed by BoringSSL (`bssl-crypto`).

Key Contributions:
1. **HPKE Decryption (`decrypt`)**:
     - Implements `decrypt` using `bssl_crypto::hpke` for "single-shot" Hybrid Public Key Encryption (HPKE) opening.
    - Supports the standard HPKE suite: `DHKEM(X25519, HKDF-SHA256)`, `HKDF-SHA256`, and `AES-256-GCM`.
    - Maps internal `HpkeAlgorithm` types to BoringSSL parameters.
  
3. **DHKEM Decapsulation (`decaps`)**:
    - Implements a standalone `decaps` function for `DHKEM(X25519, HKDF-SHA256)`.
    - Manually constructs the KEM context (suite ID, labeled IKM/Info) compliant with RFC 9180 to derive the shared secret directly from an encapsulated key and private key.
    - Validates implementation against test vectors accounting for BoringSSL's internal private key clamping.

4. **Dependencies & Error Handling**:
    - Adds `bssl-crypto` for underlying crypto operations.
    - Uses `thiserror` for structured error definitions (`KeyLenMismatch`, `DecapsError`, `HpkeDecryptionError`).

### DIFF
--- a/keymanager/km_common/src/crypto.rs
+++ b/keymanager/km_common/src/crypto.rs
@@ -271,6 +271,38 @@ mod tests {
     }
 
     #[test]
+    fn test_hpke_bad_aad() {
+        let hpke_algo = HpkeAlgorithm {
+            kem: KemAlgorithm::DhkemX25519HkdfSha256 as i32,
+            kdf: KdfAlgorithm::HkdfSha256 as i32,
+            aead: AeadAlgorithm::Aes256Gcm as i32,
+        };
+
+        let kem_algo = KemAlgorithm::DhkemX25519HkdfSha256;
+
+        let (pk_r, sk_r) = generate_keypair(kem_algo).expect("HPKE generation failed");
+
+        let pt = b"hello world";
+        let aad = b"foo";
+        let info = b"";
+
+        let hpke_kem = hpke::Kem::X25519HkdfSha256;
+        let hpke_kdf = hpke::Kdf::HkdfSha256;
+        let hpke_aead = hpke::Aead::Aes256Gcm;
+        let params = hpke::Params::new(hpke_kem, hpke_kdf, hpke_aead);
+
+        let (mut sender_ctx, enc) = hpke::SenderContext::new(&params, pk_r.as_bytes(), info)
+            .expect("HPKE setup sender failed");
+        let ciphertext = sender_ctx.seal(pt, aad);
+
+        // Tamper with aad
+        let tampered_aad = b"bar";
+
+        let result = hpke_open(&sk_r, &enc, &ciphertext, tampered_aad, &hpke_algo);
+        assert!(matches!(result, Err(Error::HpkeDecryptionError)));
+    }
+
+    #[test]
     fn test_hpke_seal_success() {
         let hpke_algo = HpkeAlgorithm {
             kem: KemAlgorithm::DhkemX25519HkdfSha256 as i32,


keymanager/km_common/src/protected_mem.rs
```diff
@@ -0,0 +1,197 @@
...
+//! location for sensitive cryptographic material.
+
+use memmap2::MmapMut;
+use std::fs::File;
+use std::io::{Error as IoError, Result as IoResult};
+use std::os::unix::io::{FromRawFd, RawFd};
+use zeroize::Zeroize;
+
+/// The system call number for `memfd_secret` on x86_64.
+const SYS_MEMFD_SECRET: i64 = 447;
```
atulpatildbz, line 17: can we add `#[cfg(target_arch = "x86_64")]`  since the syscall number might differ on other platforms?

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,353 @@
...
+    #[error("HPKE encryption error")]
+    HpkeEncryptionError,
+    #[error("Unsupported algorithm")]
+    UnsupportedAlgorithm,
+    #[error("Crypto library error")]
+    CryptoError,
+}
+
+/// Generates an X25519 keypair for the given KEM algorithm.
+pub fn generate_x25519_keypair(algo: KemAlgorithm) -> Result<(Vec<u8>, Vec<u8>), Error> {
```
bluegate010: If a new KemAlgorithm is defined in the future that isn't based on X25519 (concrete example: ML-KEM), this function can be used, which already takes in an algorithm identifier. So you could rename to just `generate_keypair`.

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,353 @@
...
+            .ok_or(Error::DecapsError)?;
+
+        // DHKEM(X25519, HKDF-SHA256)
+        // LabeledExtract(salt, label, ikm) = HKDF-Extract(salt, "HPKE-v1" || suite_id || label || ikm)
+        // suite_id = "KEM" || I2OSP(kem_id, 2)
+        let suite_id = [b'K', b'E', b'M', 0, 0x20];
+
+        // Extract eae_prk
+        // labeled_ikm = "HPKE-v1" || suite_id || "eae_prk" || shared_key
+        let mut labeled_ikm = Vec::with_capacity(7 + 5 + 7 + 32);
```
bluegate010: Optional: you could use Vec's [concat()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.concat) function to avoid the need for these magic numbers.

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,353 @@
...
+    })
+}
+
+/// BoringSSL lacks a DHKEM decap API which can perform and DH + KDF operation to generate a shared secret key.
+/// Manual implementation to decapsulate a shared secret from an encapsulated key using an X25519 private key.
+pub fn decaps_x25519(priv_key_bytes: &[u8], enc: &[u8]) -> Result<Vec<u8>, Error> {
+    clear_stack_on_return(2, || {
+        if priv_key_bytes.len() != 32 || enc.len() != 32 {
+            return Err(Error::KeyLenMismatch);
+        }
```
bluegate010: The check here is essentially duplicating the try_into().map_err(...) calls on lines 40 and 46, right?

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,353 @@
...
+    HpkeEncryptionError,
+    #[error("Unsupported algorithm")]
+    UnsupportedAlgorithm,
+    #[error("Crypto library error")]
+    CryptoError,
+}
+
+/// Generates an X25519 keypair for the given KEM algorithm.
+pub fn generate_x25519_keypair(algo: KemAlgorithm) -> Result<(Vec<u8>, Vec<u8>), Error> {
+    clear_stack_on_return(2, || match algo {
```
clundin25: nit: A constant for the `2` (Pages) would make the code easier to read.

keymanager/km_common/src/crypto.rs
clundin25: nit: Consider wrapping `(Vec<u8>, Vec<u8>)` with new types to avoid mixing keys

NilanjanDaw: Makes sense updated `generate_x25519_keypair` to `generate_keypair`

keymanager/km_common/src/crypto.rs
NilanjanDaw: Updated, thanks.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Hmm yeah that makes sense. Removed the initial check.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Done, added a constant to replace the magic number there.

keymanager/km_common/src/crypto.rs
NilanjanDaw: That makes sense, created two new NewTypes PublicKey and PrivateKey. Also replaced them at other places

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,385 @@
...
+        // LabeledExtract(salt, label, ikm) = HKDF-Extract(salt, "HPKE-v1" || suite_id || label || ikm)
+        // suite_id = "KEM" || I2OSP(kem_id, 2)
+        let suite_id = [b'K', b'E', b'M', 0, 0x20];
+
+        // Extract eae_prk
+        // labeled_ikm = "HPKE-v1" || suite_id || "eae_prk" || shared_key
+        let labeled_ikm = [b"HPKE-v1".as_slice(), &suite_id, b"eae_prk", &shared_key].concat();
+
+        let prk = hkdf::HkdfSha256::extract(&labeled_ikm, hkdf::Salt::None);
+
```
atulpatildbz: https://datatracker.ietf.org/doc/rfc9180/ has this signature LabeledExtract(salt, label, ikm)

but here it looks like we've swapped salt and label?

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl From<Vec<u8>> for PublicKey {
+    fn from(v: Vec<u8>) -> Self {
+        Self(v)
+    }
+}
+
+/// A wrapper around a private key to avoid mixing with public keys.
+#[derive(PartialEq, Eq)]
```
clundin25: You can also include `ZeroizeOnDrop` if you want to avoid leaving this in memory when it goes out of scope (but perhaps that is redundant with `clear_stack_on_return`?)

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,385 @@
...
+            .ok_or(Error::DecapsError)?;
+
+        // DHKEM(X25519, HKDF-SHA256)
+        // LabeledExtract(salt, label, ikm) = HKDF-Extract(salt, "HPKE-v1" || suite_id || label || ikm)
+        // suite_id = "KEM" || I2OSP(kem_id, 2)
+        let suite_id = [b'K', b'E', b'M', 0, 0x20];
+
+        // Extract eae_prk
+        // labeled_ikm = "HPKE-v1" || suite_id || "eae_prk" || shared_key
+        let labeled_ikm = [b"HPKE-v1".as_slice(), &suite_id, b"eae_prk", &shared_key].concat();
```
atulpatildbz: `.concat` creates a heap. and `clear_stack_on_return` won't clear that heap i think.

can you wrap this in `Zeroizing::new(`?

keymanager/km_common/src/crypto.rs
NilanjanDaw: Ah good idea, added `ZeroizeOnDrop` in any case.

keymanager/km_common/src/protected_mem.rs
NilanjanDaw: Ah yes, its coming :p I had made the changes in the previous PR, its being propagated :)

keymanager/km_common/src/crypto.rs
clundin25: It may be a good idea to wrap all the private types with structures that `ZeroizeOnDrop`.

1. It simplifies error handling (Early returns without zeroizing sensitive data).
2. As the code shifts around, critical data will still be zeroized.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Yikes, got a bit of tremor. But I think we are good here, the [BoringSSL `hkdf::extract` implementation](https://boringssl.googlesource.com/boringssl/+/refs/heads/main/rust/bssl-crypto/src/hkdf.rs#153) takes `ikm` then `salt`
```
pub fn extract(secret: &[u8], salt: Salt) -> Prk
```

keymanager/km_common/src/crypto.rs
NilanjanDaw: Thanks for pointing this out. I have added a `ZeroizeOnDrop` to the `PrivateKey` wrapper. Additionally wrapped the `labeled_ikm` and `labeled_info` with `Zeroizing::new()`

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,385 @@
...
+
+impl From<Vec<u8>> for PublicKey {
+    fn from(v: Vec<u8>) -> Self {
+        Self(v)
+    }
+}
+
+/// A wrapper around a private key to avoid mixing with public keys.
+#[derive(PartialEq, Eq)]
+pub struct PrivateKey(pub Vec<u8>);
```
zhalvorsen: Possible future improvement: I'm not sure what the rest of your API looks like, but I would recommend giving as few accesses to the raw private key bytes as possible and doing as much functionality within this struct as you can. It leaves you with less foot guns and easier traceability and zeroization claims. I'd start with making the wrapped vector private:


```suggestion
pub struct PrivateKey(Vec<u8>);
```

zhalvorsen: I'd also make this type `ZeroizeOnDrop`

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,385 @@
...
+        let mut result = vec![0u8; 32];
+        prk.expand_into(&labeled_info, &mut result)
+            .map_err(|_| Error::DecapsError)?;
+
+        Ok(result)
+    })
+}
+
+/// Decapsulates the shared secret from an encapsulated key using the specified KEM algorithm.
+pub fn decaps(priv_key: &PrivateKey, enc: &[u8], algo: KemAlgorithm) -> Result<Vec<u8>, Error> {
```
zhalvorsen: Alternatively, you could use the type system to your advantage here. You could create a `trait` for private keys that has these functions  and have different key types per algorithm so you don't have to keep track of all the key types in each function

keymanager/km_common/src/crypto.rs
atulpatildbz: Sorry to alarm you :)

Thanks for clarifying.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Added `ZeroizeOnDrop`

keymanager/km_common/src/crypto.rs
NilanjanDaw: Thanks for the comment, that makes sense. I did a major redesign of the crypto lib, to use `traits` for both the public and private keys and moved most of the functions into key specific implementations. Can you please review the [change-set](https://github.com/google/go-tpm-tools/pull/643/changes/3af7e9d6299a8a6d96fa96d98afadc90fa9e3edd) once.

keymanager/km_common/src/crypto/x25519.rs
```diff
@@ -0,0 +1,190 @@
...
+        let prk = hkdf::HkdfSha256::extract(&labeled_ikm, hkdf::Salt::None);
+
+        let pub_key = priv_key.to_public();
+
+        // 3. Expand shared_secret
+        // shared_secret = LabeledExpand(eae_prk, "shared_secret", enc || pkR, L)
+        // LabeledExpand(prk, label, info, L) = HKDF-Expand(prk, "HPKE-v1" || suite_id || label || info, L)
+        let labeled_info = Zeroizing::new(
+            [
+                &[0x00u8, 0x20] as &[u8], // L = 32
```
clundin25:
```suggestion
                &[0x00u8, 0x20][..], // L = 32
```

You can use a slice to convert a sized slice (in this case `&[u8; 2]` to `&[u8]`

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,306 @@
...
+
+mod x25519;
+pub use x25519::{X25519PrivateKey, X25519PublicKey};
+
+const CLEAR_STACK_PAGES: usize = 2;
+
+/// A trait for public keys with algorithm-specific implementations.
+pub(crate) trait PublicKeyOps: Send + Sync {
+    /// Encrypts a plaintext using HPKE.
+    fn hpke_seal_internal(
```
clundin25, line 17: nit: Worth mentioning what is returned

keymanager/km_common/src/crypto/x25519.rs
```diff
@@ -0,0 +1,190 @@
...
+                &pub_key,
+            ]
+            .concat(),
+        );
+
+        let mut result = vec![0u8; 32];
+        prk.expand_into(labeled_info.as_ref(), &mut result)
+            .map_err(|_| Error::DecapsError)?;
+
+        Ok(result)
```
clundin25: Does this shared secret need to be zeroized by the caller? Is it worth wrapping in a struct that zeroizes on drop?

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,306 @@
...
+    ) -> Result<(Vec<u8>, Vec<u8>), Error>;
+
+    /// Returns the raw bytes of the public key.
+    fn as_bytes(&self) -> &[u8];
+}
+
+/// A trait for private keys with algorithm-specific implementations.
+pub(crate) trait PrivateKeyOps: Send + Sync {
+    /// Decapsulates the shared secret from an encapsulated key.
+    fn decaps_internal(&self, enc: &[u8]) -> Result<Vec<u8>, Error>;
```
atulpatildbz: `Vec<u8>` vector is allocated on the heap. which won't zeroize itself when dropped. can we return `Zeroizing<Vec<u8>>`?

keymanager/km_common/src/crypto/x25519.rs
```diff
@@ -0,0 +1,190 @@
...
+
+                let (mut sender_ctx, encapsulated_key) =
+                    hpke::SenderContext::new(&params, self.as_ref(), b"")
+                        .ok_or(Error::HpkeEncryptionError)?;
+
+                let ciphertext = sender_ctx.seal(plaintext, aad);
+                Ok((encapsulated_key, ciphertext))
+            }
+            _ => Err(Error::UnsupportedAlgorithm),
+        }
```
zhalvorsen: fyi: optional: you can use `let else` syntax when you are looking for a specific value to clean things up a bit:

```Rust
let (
        Ok(KemAlgorithm::DhkemX25519HkdfSha256),
        Ok(KdfAlgorithm::HkdfSha256),
        Ok(AeadAlgorithm::Aes256Gcm),
    ) = (
        KemAlgorithm::try_from(algo.kem),
        KdfAlgorithm::try_from(algo.kdf),
        AeadAlgorithm::try_from(algo.aead),
    )
    else {
        return Err(Error::UnsupportedAlgorithm);
    };
    let params = hpke::Params::new(
            hpke::Kem::X25519HkdfSha256,
            hpke::Kdf::HkdfSha256,
            hpke::Aead::Aes256Gcm,
    );

    let (mut sender_ctx, encapsulated_key) =
        hpke::SenderContext::new(&params, self.as_ref(), b"")
        .ok_or(Error::HpkeEncryptionError)?;

    let ciphertext = sender_ctx.seal(plaintext, aad);
    Ok((encapsulated_key, ciphertext))
```

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+/// Decrypts a ciphertext using HPKE (Hybrid Public Key Encryption).
+pub fn hpke_open(
+    priv_key: &PrivateKey,
+    enc: &[u8],
+    ciphertext: &[u8],
+    aad: &[u8],
+    algo: &HpkeAlgorithm,
+) -> Result<Vec<u8>, Error> {
```
atulpatildbz: similarly here result is allocated on the heap (in `Vec`) .  the caller drops this Vec without zeroizing it, the secret stays in memory

would be safer to return `Zeroizing<Vec<u8>>`. lmk what you think

keymanager/km_common/src/crypto.rs
bluegate010: Might be useful to define a ZeroizingVec type alias or something and use it throughout, just to make it a bit harder to accidentally use a raw Vec. @clundin25 / @zhalvorsen might have better ideas here.

keymanager/km_common/src/crypto.rs
bluegate010: Ah I'm sorry, I didn't realize that when I suggested moving to .concat(), that would lead to additional heap issues.

I see you have an implementation of `labeled_extract()` in https://github.com/google/go-tpm-tools/pull/649 that's earmarked for tests, but makes it clear when a Vec is being created. Could consider adopting that pattern here.

keymanager/km_common/src/crypto.rs
clundin25: My recommendation is to wrap sensitive types with a new type that implements `ZeroizeOnDrop`.

Using raw Vec types like this could lead to implementation bugs, e.g. mixing up types or forgetting to zeroize sensitive data.

zhalvorsen: Yeah, I think that would be good. I would maybe also wrap it in a `Box` so it doesn't get copied around by accident. So maybe something like:

```Rust
#[derive(ZeroizeOnDrop)]
pub struct Secret(Box<[u8]>);
```

keymanager/km_common/src/crypto.rs
NilanjanDaw: Thanks, added return doc-strings for this and other interfaces.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Resolving this, now that we have moved to a trait format.

keymanager/km_common/src/crypto/x25519.rs
NilanjanDaw: Created a new `SecretBox` to handle secret key material and zeroize on drop.

keymanager/km_common/src/crypto/x25519.rs
NilanjanDaw: Done, updated both the open and seal functions.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Thank you everyone for the recommendations. I have created a new `SecretBox` type to handle sensitive material, and migrated the private keys and other internal secret material handling to this new type.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Thanks, based on this and other comments/recommendations I am migrating all sensitive key-material to a new `SecretBox` type.

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,318 @@
...
+use clear_on_drop::clear_stack_on_return;
+use thiserror::Error;
+
+mod x25519;
+pub use x25519::{X25519PrivateKey, X25519PublicKey};
+
+const CLEAR_STACK_PAGES: usize = 2;
+
+/// A trait for public keys with algorithm-specific implementations.
+pub(crate) trait PublicKeyOps: Send + Sync {
```
bluegate010, line 13: These traits assume that every public key can do HPKE sealing and every private key can do both KEM decaps and HPKE open. If there were a future use-case for an ECDSA or ML-DSA key, it would be hard to represent those in this type system.

Would it make sense to have individual traits for each of these capabilities? Leaves room in the future for Sign and Verify traits.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Moved the `labeled_extract` and `labeled_expand` to separate functions with `SecretBox` based storage. This should make it easier to handle the secret material clean-up.

keymanager/km_common/src/crypto.rs
NilanjanDaw: That makes sense. Let me do this in a follow up PR. Filed Issue #660 to track the change.

keymanager/km_common/src/crypto/x25519.rs
NilanjanDaw: Ack. Done.

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,318 @@
...
+        let ciphertext = sender_ctx.seal(pt, aad);
+
+        let decrypted =
+            hpke_open(&sk_r, &enc, &ciphertext, aad, &hpke_algo).expect("Decryption failed");
+
+        assert_eq!(decrypted.as_slice(), pt);
+    }
+
+    #[test]
+    fn test_hpke_open_failure() {
```
atulpatildbz, line 240: should we also have a test to verify aad checks?

e.g. sender sends with `aad = "foo"` and receiver open with `aad = "bar"`

NilanjanDaw: Ack. Added a new `test_hpke_bad_aad` test.

PR #644, opened 2026-02-05
[keymanager] Add BoringSSL as git submodule for build dependency

atulpatildbz:
Add boring SSL git submodule to provide bssl-crypto crate dependency.

More context:
Keymanager's rust code would depend on the `bssl-crypto` crate. `bssl-crypto` is a wrapper around the BoringSSL C library, which requires following process:

1. building the underlying C library (`libcrypto.a`) from BoringSSL source code. 
2. Rust bindings, pointing `Cargo.toml` to local Rust source and the build script to compiled C artifacts

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "keymanager/boringssl"]
+	path = keymanager/boringssl
+	url = https://boringssl.googlesource.com/boringssl


NilanjanDaw (approved):
Do you also want to add a method to pin a version via a config, pull the code and build the `bssl-crypto` lib to this PR?

Can be a future PR too.

atulpatildbz:
> Do you also want to add a method to pin a version via a config, pull the code and build the `bssl-crypto` lib to this PR?
> 
> Can be a future PR too.

the submodule already does commit SHA based pinning. 
BoringSSL doesn't do semver releases, it's commit based. the version config in this case would just duplicate what submodule shows.
build step for bssl-crypto will be handled by Cargo.toml pointing at the local submodule path. 

I think a config would be nice-to-have for ergonomics but let me know if you still find it necessary

PR #645, opened 2026-02-06
[KeyManager] Implement FFI for KEM and binding key generation

NilanjanDaw:
Introduces FFI-accessible functions for generating and managing cryptographic keys within the Key Protection Service (KPS) and Workload Service (WS).
Key Changes:
- **FFI Compatibility**: Configured `prost_build` in km_common to apply `#[repr(C)]` to the `HpkeAlgorithm` struct, enabling its use in C-style interfaces.
- **Key Generation Logic**: Added create_key_record in km_common to centralize secure keypair generation (X25519), ensuring private keys are immediately moved to secure Vault storage and zeroized from temporary memory.
- Key Management: Implemented `KeyRegistry` and a global registry instance in both KPS and WS to track generated keys by UUID.
- **FFI Exports**:
    - Added `key_manager_generate_kem_keypair` to KPS for generating KEM keys with associated binding public keys.
    - Added `key_manager_generate_binding_keypair` to WS for generating standard binding keys.
- **Testing**: Added comprehensive unit tests for FFI functions, including error handling for invalid algorithms and null pointers.

This PR is built on top of PR #643  [[KeyManager] Add HPKE and DHKEM crypto primitives](https://github.com/google/go-tpm-tools/pull/643).
Please review commit [#38731051 onwards](https://github.com/google/go-tpm-tools/pull/645/changes/1229d3b4bef9a963f6834b22c27f70efee62ba83..13fbcf6f2e881e5ff86e354bb6ade057ebbf3b57)

### DIFF
--- a/keymanager/key_protection_service/key_custody_core/src/lib.rs
+++ b/keymanager/key_protection_service/key_custody_core/src/lib.rs
@@ -1,8 +1,9 @@
 use km_common::algorithms::HpkeAlgorithm;
 use km_common::crypto::PublicKey;
-use km_common::key_types::{KeyRegistry, KeySpec};
+use km_common::key_types::{KeyRecord, KeyRegistry, KeySpec};
 use std::slice;
 use std::sync::LazyLock;
+use std::time::Duration;
 
 static KEY_REGISTRY: LazyLock<KeyRegistry> = LazyLock::new(KeyRegistry::default);
 
@@ -12,13 +13,8 @@ fn generate_kem_keypair_internal(
     binding_pubkey: PublicKey,
     expiry_secs: u64,
 ) -> Result<(uuid::Uuid, PublicKey), i32> {
-    let result = km_common::key_types::create_key_record(algo, expiry_secs, |algo, kem_pub_key| {
-        KeySpec::KemWithBindingPub {
-            algo,
-            kem_public_key: kem_pub_key,
-            binding_public_key: binding_pubkey.clone(),
-        }
-    });
+    let result =
+        KeyRecord::create_bound_kem_key(algo, binding_pubkey, Duration::from_secs(expiry_secs));
 
     match result {
         Ok(record) => {
@@ -30,7 +26,7 @@ fn generate_kem_keypair_internal(
             KEY_REGISTRY.add_key(record);
             Ok((id, pubkey))
         }
-        Err(e) => Err(e),
+        Err(_) => Err(-1),
     }
 }
 
@@ -65,13 +61,12 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
     expiry_secs: u64,
     out_uuid: *mut u8,
     out_pubkey: *mut u8,
-    out_pubkey_len: *mut usize,
+    out_pubkey_len: usize,
 ) -> i32 {
     // Safety Invariant Checks
     if binding_pubkey.is_null()
         || binding_pubkey_len == 0
         || out_pubkey.is_null()
-        || out_pubkey_len.is_null()
         || out_uuid.is_null()
     {
         return -1;
@@ -79,6 +74,9 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
 
     // Convert to Safe Types
     let binding_pubkey_slice = unsafe { slice::from_raw_parts(binding_pubkey, binding_pubkey_len) };
+    let out_uuid = unsafe { slice::from_raw_parts_mut(out_uuid, 16) };
+    let out_pubkey = unsafe { slice::from_raw_parts_mut(out_pubkey, out_pubkey_len) };
+
     let binding_pubkey = match PublicKey::try_from(binding_pubkey_slice.to_vec()) {
         Ok(pk) => pk,
         Err(_) => return -1,
@@ -87,18 +85,11 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
     // Call Safe Internal Function
     match generate_kem_keypair_internal(algo, binding_pubkey, expiry_secs) {
         Ok((id, pubkey)) => {
-            unsafe {
-                let buf_len = *out_pubkey_len;
-                if buf_len != pubkey.as_bytes().len() {
-                    return -2;
-                }
-                std::ptr::copy_nonoverlapping(id.as_bytes().as_ptr(), out_uuid, 16);
-                std::ptr::copy_nonoverlapping(
-                    pubkey.as_bytes().as_ptr(),
-                    out_pubkey,
-                    pubkey.as_bytes().len(),
-                );
+            if out_pubkey_len != pubkey.as_bytes().len() {
+                return -2;
             }
+            out_uuid.copy_from_slice(id.as_bytes());
+            out_pubkey.copy_from_slice(pubkey.as_bytes());
             0 // Success
         }
         Err(e) => e,
@@ -137,7 +128,7 @@ mod tests {
         let binding_pubkey = [1u8; 32];
         let mut uuid_bytes = [0u8; 16];
         let mut pubkey_bytes = [0u8; 32];
-        let mut pubkey_len: usize = pubkey_bytes.len();
+        let pubkey_len: usize = pubkey_bytes.len();
         let algo = HpkeAlgorithm {
             kem: KemAlgorithm::DhkemX25519HkdfSha256 as i32,
             kdf: KdfAlgorithm::HkdfSha256 as i32,
@@ -152,7 +143,7 @@ mod tests {
                 3600,
                 uuid_bytes.as_mut_ptr(),
                 pubkey_bytes.as_mut_ptr(),
-                &mut pubkey_len,
+                pubkey_len,
             )
         };
 
@@ -167,7 +158,7 @@ mod tests {
         let binding_pubkey = [1u8; 32];
         let mut uuid_bytes = [0u8; 16];
         let mut pubkey_bytes = [0u8; 32];
-        let mut pubkey_len: usize = pubkey_bytes.len();
+        let pubkey_len: usize = pubkey_bytes.len();
         let algo = HpkeAlgorithm {
             kem: 999, // Invalid KEM
             kdf: KdfAlgorithm::HkdfSha256 as i32,
@@ -182,7 +173,7 @@ mod tests {
                 3600,
                 uuid_bytes.as_mut_ptr(),
                 pubkey_bytes.as_mut_ptr(),
-                &mut pubkey_len,
+                pubkey_len,
             )
         };
 
@@ -195,7 +186,7 @@ mod tests {
         let binding_pubkey = [1u8; 32];
         let mut uuid_bytes = [0u8; 16];
         let mut pubkey_bytes = [0u8; 64];
-        let mut pubkey_len: usize = pubkey_bytes.len();
+        let pubkey_len: usize = pubkey_bytes.len();
         let algo = HpkeAlgorithm {
             kem: KemAlgorithm::DhkemX25519HkdfSha256 as i32,
             kdf: KdfAlgorithm::HkdfSha256 as i32,
@@ -210,7 +201,7 @@ mod tests {
                 3600,
                 uuid_bytes.as_mut_ptr(),
                 pubkey_bytes.as_mut_ptr(),
-                &mut pubkey_len,
+                pubkey_len,
             )
         };
 
@@ -236,7 +227,7 @@ mod tests {
                 3600,
                 uuid_bytes.as_mut_ptr(),
                 std::ptr::null_mut(),
-                std::ptr::null_mut(),
+                0,
             )
         };
 
@@ -261,7 +252,7 @@ mod tests {
                 3600,
                 uuid_bytes.as_mut_ptr(),
                 std::ptr::null_mut(),
-                std::ptr::null_mut(),
+                0,
             )
         };
 
--- a/keymanager/km_common/src/crypto.rs
+++ b/keymanager/km_common/src/crypto.rs
@@ -90,11 +90,10 @@ pub enum PrivateKey {
     X25519(X25519PrivateKey),
 }
 
-impl PrivateKey {
-    /// Consumes the `PrivateKey` and returns the inner `SecretBox`.
-    pub fn into_secret(self) -> SecretBox {
-        match self {
-            PrivateKey::X25519(sk) => sk.into_secret(),
+impl From<PrivateKey> for SecretBox {
+    fn from(key: PrivateKey) -> SecretBox {
+        match key {
+            PrivateKey::X25519(sk) => SecretBox::from(sk),
         }
     }
 }
@@ -131,6 +130,8 @@ pub enum Error {
     HpkeEncryptionError,
     #[error("Unsupported algorithm")]
     UnsupportedAlgorithm,
+    #[error("Invalid key")]
+    InvalidKey,
     #[error("Crypto library error")]
     CryptoError,
 }
--- a/keymanager/km_common/src/crypto/x25519.rs
+++ b/keymanager/km_common/src/crypto/x25519.rs
@@ -55,10 +55,9 @@ impl PublicKeyOps for X25519PublicKey {
 /// X25519-based private key implementation.
 pub struct X25519PrivateKey(SecretBox);
 
-impl X25519PrivateKey {
-    /// Consumes the `X25519PrivateKey` and returns the inner `SecretBox`.
-    pub fn into_secret(self) -> SecretBox {
-        self.0
+impl From<X25519PrivateKey> for SecretBox {
+    fn from(key: X25519PrivateKey) -> SecretBox {
+        key.0
     }
 }
 
--- a/keymanager/km_common/src/key_types.rs
+++ b/keymanager/km_common/src/key_types.rs
@@ -1,6 +1,6 @@
-use crate::algorithms::{HpkeAlgorithm, KemAlgorithm};
+use crate::algorithms::{AeadAlgorithm, HpkeAlgorithm, KdfAlgorithm, KemAlgorithm};
 use crate::crypto;
-use crate::crypto::PublicKey;
+use crate::crypto::{PublicKey, secret_box};
 use crate::protected_mem::Vault;
 use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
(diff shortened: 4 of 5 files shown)


keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,222 @@
-pub fn add(left: u64, right: u64) -> u64 {
-    left + right
+use km_common::algorithms::HpkeAlgorithm;
+use km_common::key_types::{KeyRecord, KeyRegistry, KeySpec};
+use lazy_static::lazy_static;
+use std::slice;
+
+lazy_static! {
+    static ref KEY_REGISTRY: KeyRegistry = KeyRegistry::default();
+}
```
clundin25: Slightly different ergonomics but the standard library could be used to eliminate this dependency:

https://doc.rust-lang.org/std/sync/struct.LazyLock.html

keymanager/km_common/src/key_types.rs
```diff
@@ -51,3 +53,142 @@ pub type KeyHandle = Uuid;
...
+        .ok()
+        .and_then(|k| crypto::generate_keypair(k).ok())
+    {
+        Some(pair) => pair,
+        None => return Err(-1),
+    };
+
+    let id = Uuid::new_v4();
+    let vault = Vault::new(&mut priv_key.0);
+    priv_key.0.zeroize();
```
zhalvorsen: note: you shouldn't need to zeroize here because it should be done in the `Vault::new` function

keymanager/km_common/src/protected_mem.rs
```diff
@@ -107,7 +107,10 @@ mod tests {
         assert_eq!(vault.as_bytes(), &original_data);
         assert_eq!(vault.as_ref(), &original_data);
         assert_ne!(data, original_data, "Source data should be zeroed");
-        assert!(data.iter().all(|&b| b == 0), "Source data should be all zeros");
+        assert!(
```
zhalvorsen: Was this change done by `rustfmt`. Is there are CI rule to make sure this is done before submitting?

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,229 @@
...
+                    if buf_len >= pubkey.as_ref().len() {
+                        std::ptr::copy_nonoverlapping(
+                            pubkey.as_ref().as_ptr(),
+                            out_pubkey,
+                            pubkey.as_ref().len(),
+                        );
+                        *out_pubkey_len = pubkey.as_ref().len();
+                    } else {
+                        return -2; // buffer too small
+                    }
```
zhalvorsen: nit: you can reduce the nested code a layer and improve readability a little by returning the error here:


```suggestion
                    if buf_len < pubkey.as_ref().len() {
                        return -2; // buffer too small
                    }
                    std::ptr::copy_nonoverlapping(
                        pubkey.as_ref().as_ptr(),
                        out_pubkey,
                        pubkey.as_ref().len(),
                    );
                    *out_pubkey_len = pubkey.as_ref().len();
```

keymanager/km_common/src/key_types.rs
bluegate010: You could have Vault::new take ownership of the private key so it's not possible to attempt to zeroize twice.

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,398 @@
+use crate::algorithms::{AeadAlgorithm, HpkeAlgorithm, KdfAlgorithm, KemAlgorithm};
+use bssl_crypto::{hkdf, hpke, x25519};
+use clear_on_drop::clear_stack_on_return;
+use thiserror::Error;
+use zeroize::{ZeroizeOnDrop, Zeroizing};
+
+const CLEAR_STACK_PAGES: usize = 2;
```
atulpatildbz: curious how did we arrive on the size `2`?

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Thank you, I was not aware of this. Updated to use `LazyLock`

keymanager/km_common/src/key_types.rs
NilanjanDaw: Redesigned the implementation around a new SecretBox to simplify secret handling. Please review this once more.

keymanager/km_common/src/protected_mem.rs
NilanjanDaw: Yes this was `rustfmt`. I have been manually running this before committing along with `cargo clippy` but let me wire up some automation to do this automatically.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Thank you, updated the code to reduce nesting.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,233 @@
...
+            let id = record.meta.id;
+            let pubkey = match &record.meta.spec {
+                KeySpec::KemWithBindingPub { kem_public_key, .. } => kem_public_key.clone(),
+                _ => return -1,
+            };
+            KEY_REGISTRY.add_key(record);
+            unsafe {
+                if out_pubkey.is_null() || out_pubkey_len.is_null() || out_uuid.is_null() {
+                    return -1;
+                }
```
clundin25: Nit: Seems like all the parameter validation could be moved to the very top, I think that would make it easier to read.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,233 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_generate_kem_keypair(
+    algo: HpkeAlgorithm,
+    binding_pubkey: *const u8,
+    binding_pubkey_len: usize,
+    expiry_secs: u64,
+    out_uuid: *mut u8,
+    out_pubkey: *mut u8,
+    out_pubkey_len: *mut usize,
+) -> i32 {
```
clundin25, line 65: Right now this function is mixing business logic & FFI safety invariants.

I recommend checking safety invariants (null ptr, buffer size, etc) in this function, converting it into safe types, and calling a safe function that operates on the sanitized data.

That will reduce the surface area of this FFI boundary.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,233 @@
...
+/// Generates a new KEM keypair associated with a binding public key.
+///
+/// ## Arguments
+/// * `algo` - The HPKE algorithm to use for the keypair.
+/// * `binding_pubkey` - A pointer to the binding public key bytes.
+/// * `binding_pubkey_len` - The length of the binding public key.
+/// * `expiry_secs` - The expiration time of the key in seconds from now.
+/// * `out_uuid` - A pointer to a 16-byte buffer where the key UUID will be written.
+/// * `out_pubkey` - A pointer to a buffer where the public key will be written.
+/// * `out_pubkey_len` - A pointer to a `usize` that contains the size of `out_pubkey` buffer.
```
clundin25, line 42: Do you need to be precise with the allocation, or is it okay to overcommit & shrink the buffer?

This implies it's okay to over commit

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,173 @@
...
+/// * `-1` if an error occurred during key generation.
+/// * `-2` if the `out_pubkey` buffer is too small.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_generate_binding_keypair(
+    algo: HpkeAlgorithm,
+    expiry_secs: u64,
+    out_uuid: *mut u8,
+    out_pubkey: *mut u8,
+    out_pubkey_len: *mut usize,
+) -> i32 {
```
clundin25, line 61: Same comment on mixing business & validation logic

keymanager/key_protection_service/key_custody_core/src/lib.rs
zhalvorsen: Similarly, it's good to isolate the unsafe code as much as possible so it can be easy to see where the unsafe review should be focused. This check is all safe code so it would be good to remove it from the unsafe block.

keymanager/km_common/src/crypto/secret_box.rs
```diff
@@ -0,0 +1,73 @@
...
+///
+/// This struct is intended to hold sensitive data that needs to be stored on the heap.
+/// The `Box<[u8]>` ensures that the data is not accidentally copied, and `ZeroizeOnDrop`
+/// ensures that the memory is cleared when the struct goes out of scope.
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct SecretBox(Box<[u8]>);
+
+impl SecretBox {
+    /// Creates a new `SecretBox` from a `Vec<u8>`.
+    pub fn new(data: Vec<u8>) -> Self {
```
zhalvorsen, line 13: Similar to the previous setup. It would be good to zeroize the input here so there isn't another copy of the secret floating around. Feel free to do this in a follow-on PR.

bluegate010: `into_boxed_slice` changes ownership so there shouldn't be an extra copy?

keymanager/km_common/src/crypto/secret_box.rs
bluegate010, line 1: This same file shows up as new in https://github.com/google/go-tpm-tools/pull/643. Similar to other core files. Makes it a bit hard to keep track of which feedback has been given.

keymanager/km_common/src/crypto/secret_box.rs
zhalvorsen: Good call. I was thinking that it didn't a second allocation, but https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_boxed_slice says:

```
This conversion does not allocate on the heap and happens in place.
```

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Hmm at the time, I was wondering about how to support different key-sizes. But now that I think more, it actually does not make sense and probably just opens us up to nasty bugs. Updated to strict key-size matching.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Thank you, make sense, updated the function to separate the safety checks and the business logic, and also tried to reduce the scope of the unsafe blocks.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Thank you, make sense, updated the function to separate the safety checks and the business logic, and also tried to reduce the scope of the unsafe blocks.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack, thank you, this is a good advice. Updated the function to separate the safety checks and the business logic, and also tried to reduce the scope of the unsafe blocks.

keymanager/km_common/src/crypto/secret_box.rs
NilanjanDaw: My apologies. The PRs are stacked which is making the diff difficult to review. The parent PR has been merged. So this one should now be easier to review.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,270 @@
...
+/// * `-2` if the `out_pubkey` buffer size does not match the key size.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_generate_kem_keypair(
+    algo: HpkeAlgorithm,
+    binding_pubkey: *const u8,
+    binding_pubkey_len: usize,
+    expiry_secs: u64,
+    out_uuid: *mut u8,
+    out_pubkey: *mut u8,
+    out_pubkey_len: *mut usize,
```
zhalvorsen: Because this value is no longer mutated in the function, it would be better to take the value directly to avoid more unsafe logic.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,270 @@
...
+                let buf_len = *out_pubkey_len;
+                if buf_len != pubkey.as_bytes().len() {
+                    return -2;
+                }
+                std::ptr::copy_nonoverlapping(id.as_bytes().as_ptr(), out_uuid, 16);
+                std::ptr::copy_nonoverlapping(
+                    pubkey.as_bytes().as_ptr(),
+                    out_pubkey,
+                    pubkey.as_bytes().len(),
+                );
```
zhalvorsen: Alternatively, you could convert the pointer types into slices at the beginning of the function, then use regular slice copying logic. This would reduce the amount of unsafe code to only interpreting pointers as slices.

```Rust
let out_uuid = unsafe { slice::from_raw_parts_mut(out_uuid, 16) };
let out_pubkey = unsafe { slice::from_raw_parts_mut(out_pubkey, out_pubkey_len) };

// Later, during the match statement
if out_pubkey.len() != pubkey.as_bytes().len() {
    return -2;
}
out_uuid.copy_from_slice(id.as_bytes());
out_pubkey.copy_from_slice(pubkey.as_bytes());
```

keymanager/km_common/src/crypto/x25519.rs
```diff
@@ -55,6 +55,13 @@ impl PublicKeyOps for X25519PublicKey {
 /// X25519-based private key implementation.
 pub struct X25519PrivateKey(SecretBox);
 
+impl X25519PrivateKey {
+    /// Consumes the `X25519PrivateKey` and returns the inner `SecretBox`.
+    pub fn into_secret(self) -> SecretBox {
+        self.0
+    }
+}
```
zhalvorsen: You can use the `std` `From` trait for this. `From` automatically creates an `Into` implementation as well. So adding the following code would let you call `priv.into()` to convert it to a `SecretBox`.


```suggestion
impl From<X25519PrivateKey> for SecretBox {
    fn from(key: X25519PrivateKey) -> SecretBox {
        key.0
    }
}
```

keymanager/km_common/src/crypto.rs
```diff
@@ -81,6 +90,15 @@ pub enum PrivateKey {
     X25519(X25519PrivateKey),
 }
 
+impl PrivateKey {
+    /// Consumes the `PrivateKey` and returns the inner `SecretBox`.
+    pub fn into_secret(self) -> SecretBox {
```
zhalvorsen: Similar use of the `From` trait here

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,270 @@
...
+        Err(_) => return -1,
+    };
+
+    // Call Safe Internal Function
+    match generate_kem_keypair_internal(algo, binding_pubkey, expiry_secs) {
+        Ok((id, pubkey)) => {
+            unsafe {
+                let buf_len = *out_pubkey_len;
+                if buf_len != pubkey.as_bytes().len() {
+                    return -2;
```
atulpatildbz: in case of this failure, we're not removing the key that was added to the registry by `generate_binding_keypair_internal`

what do you think about comparing `*out_pubkey_len` with known values based on algorithm before calling `generate_kem_keypair_internal`?

or do you think this is overkill?

keymanager/km_common/src/key_types.rs
```diff
@@ -51,3 +52,141 @@ pub type KeyHandle = Uuid;
...
+
+    let id = Uuid::new_v4();
+    let vault = Vault::new(priv_key.into_secret());
+    let vault = vault.map_err(|_| -1)?;
+
+    let record = KeyRecord {
+        meta: KeyMetadata {
+            id,
+            created_at: Instant::now(),
+            delete_after: Instant::now() + Duration::from_secs(expiry_secs),
```
atulpatildbz: since `expiry_secs` is user controlled, there's a risk of panic on overflow. can we use `checked_add`?

```
let duration = Duration::from_secs(expiry_secs);
let delete_after = Instant::now().checked_add(duration).ok_or(-1)?;
```

keymanager/km_common/src/key_types.rs
```diff
@@ -51,3 +52,141 @@ pub type KeyHandle = Uuid;
...
+/// Helper function to create a KeyRecord and generate the underlying keypair.
+pub fn create_key_record<F>(
+    algo: HpkeAlgorithm,
+    expiry_secs: u64,
+    spec_builder: F,
+) -> Result<KeyRecord, i32>
+where
+    F: FnOnce(HpkeAlgorithm, PublicKey) -> KeySpec,
+{
+    let (pub_key, priv_key) = match KemAlgorithm::try_from(algo.kem)
```
atulpatildbz: We are only validating `algo.kem` here.

If `algo.kdf` or `algo.aead` contain invalid values (e.g. `999`), they will be stored in the KeyRegistry without error. This will cause failures later when the key is used for sealing/opening, making debugging difficult.

can we add validation for `algo.kdf` and `algo.aead` as well, for example:

if KdfAlgorithm::try_from(algo.kdf).is_err() || AeadAlgorithm::try_from(algo.aead).is_err() {
    return Err(-1);
}

keymanager/km_common/src/key_types.rs
atulpatildbz: where are we doing cleanup of expired keys?

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: makes sense, removed `mut` from both this one and other KCC function.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack updated.

keymanager/km_common/src/crypto/x25519.rs
NilanjanDaw: Oh okay I was unaware of this, thank you. Update to `std From` trait.

keymanager/km_common/src/crypto.rs
NilanjanDaw: Done.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: I thought about it a bit. I think at this point it might be a bit overkill? Although I can think of a malicious adversary overloading the system by creating invalid keys. How about I take this in a future PR?

Added issue #661 to track this.

keymanager/km_common/src/key_types.rs
NilanjanDaw: This is good advice. Updated. Thanks!

keymanager/km_common/src/key_types.rs
NilanjanDaw: Good catch. Updated the code to check for all suite variables.

keymanager/km_common/src/key_types.rs
NilanjanDaw: Expired key clean up is handled in PR #647 along with the general Key Destroy functionality.

keymanager/km_common/src/crypto/x25519.rs
```diff
@@ -5,7 +5,7 @@ use bssl_crypto::{hkdf, hpke, x25519};
 
 /// X25519-based public key implementation.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct X25519PublicKey([u8; 32]);
+pub struct X25519PublicKey(pub(crate) [u8; 32]);
```
clundin25, line 8: BTW, the spec calls for sanitizing the public key in 7.1.2.

This is already being done by BoringSSL, correct?

atulpatildbz (approved):

PR #646, opened 2026-02-06
[KeyManager]: (build) boringssl auto-build using cmake crate

atulpatildbz:
Integrates BoringSSL into the keymanager build lifecycle by introducing `build.rs` for orchestration.

We've adopted `sys-crate` pattern for handling C/C++ dependencies. 
Instead of relying on external shell scripts, we used the `cmake` build dependency to compile the native library (`boringssl`) as a static artifact during the cargo build step.

The `RUST_BINDINGS` CMake definition is injected to ensure BoringSSL generates the exact rust_wrapper symbols required by the `bssl-crypto` Rust crate

The resulting artifacts (`libcrypto.a`, `libssl.a`) are statically linked into the final binary, simplifying deployment.

**Implementation Details**

* `bssl-sys/build.rs`: Implemented a custom build script that uses `cmake::Config` to drive the BoringSSL compilation.
  * Automatically orchestrates the `cmake` build process, handling platform detection and toolchain selection.
  * sets -DRUST_BINDINGS=ON to trigger the generation of Rust-compatible bindgen wrappers
  * detects and links the produced static archives (`crypto`, `ssl`, `rust_wrapper`) and propagates the search paths via `cargo:rustc-link-search`

**Verification**

* Tested with changes in #643 
* Tested on Linux/x86_64. Ran `cargo test -p km_common and cargo test -p bssl-crypto`. All 131 tests passed (including 93 bssl-crypto unit tests and 15 km_common tests), confirming that the underlying crypto symbols were correctly linked and callable
* Validated that `cargo build` succeeds from a fresh clone without needing manual `cmake` invocation in the terminal

### DIFF
--- a/keymanager/third_party/bssl-sys/build.rs
+++ b/keymanager/third_party/bssl-sys/build.rs
@@ -89,18 +89,15 @@ fn main() {
     let target = env::var("TARGET").unwrap();
     let out_dir = env::var("OUT_DIR").unwrap();
     let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
-    
+
     // Locate the BoringSSL source relative to this cargo manifest
     // keymanager/third_party/bssl-sys -> keymanager/boringssl
     let bssl_source_dir = Path::new(&manifest_dir).join("../../boringssl");
 
     // Auto-init git submodule if BoringSSL source is missing.
-    // Follows the curl-sys / libgit2-sys pattern: attempt silently, discard
-    // errors (on crates.io the source is already present as a regular directory,
-    // so git may not be available and that's fine).
     if !bssl_source_dir.join("CMakeLists.txt").exists() {
         let _ = Command::new("git")
-            .args(["submodule", "update", "--init", "boringssl"])
+            .args(["submodule", "update", "--init", "--recursive", "boringssl"])
             .current_dir(Path::new(&manifest_dir).join("../.."))
             .status();
     }
@@ -131,7 +128,7 @@ fn main() {
     // BoringSSL install target puts libs in `lib/` and includes in `include/`.
     // BUT `bssl_sys` target might not install the wrapper?
     // Let's verify where `cmake` crate puts it. It usually puts build artifacts in `build/`.
-    
+
     // cmake::Config::build() guarantees this path exists on success (it
     // panics on failure), but assert for clarity since the layout matters.
     let build_dir = dst.join("build");
@@ -145,11 +142,17 @@ fn main() {
     // Note: We might need to look in `dst/lib` if it was installed, or `build_dir` if not.
     // BoringSSL puts static libs in the top level of build dir usually, or `crypto/` `ssl/` subdirs.
     // Let's add multiple search paths to be safe, similar to original script logic but adapted.
-    
+
     println!("cargo:rustc-link-search=native={}", build_dir.display());
-    println!("cargo:rustc-link-search=native={}/crypto", build_dir.display());
+    println!(
+        "cargo:rustc-link-search=native={}/crypto",
+        build_dir.display()
+    );
     println!("cargo:rustc-link-search=native={}/ssl", build_dir.display());
-    println!("cargo:rustc-link-search=native={}/rust/bssl-sys", build_dir.display());
+    println!(
+        "cargo:rustc-link-search=native={}/rust/bssl-sys",
+        build_dir.display()
+    );
 
     // Also check `dst/lib` just in case `cmake` crate installed them there
     println!("cargo:rustc-link-search=native={}/lib", dst.display());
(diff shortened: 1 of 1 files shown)


keymanager/third_party/bssl-sys/build.rs
```diff
@@ -0,0 +1,184 @@
...
+    if !bssl_source_dir.join("CMakeLists.txt").exists() {
+        panic!(
+            "BoringSSL source not found at {}. Run 'git submodule update --init --recursive'",
+            bssl_source_dir.display()
+        );
+    }
+
+    // Use cmake crate to build BoringSSL
+    // It will automatically handle compilation flags, finding cmake, etc.
+    let dst = cmake::Config::new(&bssl_source_dir)
```
NilanjanDaw, line 120: Should we check if `cmake` is available on the system before running the config?

keymanager/third_party/bssl-sys/build.rs
```diff
@@ -0,0 +1,184 @@
...
+        .build();
+
+    // The cmake crate installs artifacts to `dst`.
+    // However, BoringSSL's internal structure when built might be different.
+    // Usually artifacts are in `dst/build` if we didn't install, but `cmake` crate defaults to install.
+    // BoringSSL install target puts libs in `lib/` and includes in `include/`.
+    // BUT `bssl_sys` target might not install the wrapper?
+    // Let's verify where `cmake` crate puts it. It usually puts build artifacts in `build/`.
+    
+    let build_dir = dst.join("build");
```
NilanjanDaw, line 134: nit: should we check if this path exists? Not sure if the path always exists if cmake is present.

keymanager/third_party/bssl-sys/build.rs
```diff
@@ -0,0 +1,184 @@
...
+    let prefix_inc_source_file = bssl_source_dir.join("rust/bssl-sys/boringssl_prefix_symbols_bindgen.rs.in");
+
+    let bindgen_out_file = Path::new(&out_dir).join("bindgen.rs");
+
+    let bindgen_source = std::fs::read_to_string(&bindgen_source_file).expect(&format!(
+        "Could not read bindings from '{}'. Did the build fail?",
+        bindgen_source_file.display(),
+    ));
+
+    println!("cargo:rerun-if-changed={}", bindgen_source_file.display());
```
NilanjanDaw, line 187: IIUC this reruns the build if the generated bindings change. We should probably also rerun the build if the BoringSSL submodule chnages.

```rs
println!("cargo:rerun-if-changed={}", bssl_source_dir.display());
```

keymanager/third_party/bssl-sys/build.rs
```diff
@@ -0,0 +1,184 @@
...
+        match env::var("CARGO_CFG_TARGET_OS").unwrap().as_ref() {
+            "macos" => Some("c++".into()),
+            _ => Some("stdc++".into()),
+        }
+    } else {
+        None
+    }
+}
+
+fn main() {
```
NilanjanDaw, line 88: I cloned this PR and ran `cargo build` but faced the following error
```
warning: `bssl-sys` (build script) generated 1 warning
error: failed to run custom build command for `bssl-sys v0.1.0 (/<path>/go-tpm-tools/keymanager/third_party/bssl-sys)`

Caused by:
  process didn't exit successfully: `/<path>/go-tpm-tools/keymanager/target/debug/build/bssl-sys-690353b8862fc20d/build-script-build` (exit status: 101)
  --- stderr

  thread 'main' panicked at third_party/bssl-sys/build.rs:97:9:
  BoringSSL source not found at /<path>/go-tpm-tools/keymanager/third_party/bssl-sys/../../boringssl. Run 'git submodule update --init --recursive'
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
```

The issue got fixed after a manual `git submodule update --init --recursive` run. Should we do something to automatically run the `submodule update` command so that CIs don't fail? Also I think it would help keep the code updated to the `tip-of-tree`.

But I will leave this up to you, if you want to handle this differently.

keymanager/third_party/bssl-sys/build.rs
atulpatildbz: The CMake created itself panics with "is cmake not installed?". Therefore i don't think pre-check is needed.

I've added a comment in the code explaining this so it's clear to future readers.

keymanager/third_party/bssl-sys/build.rs
atulpatildbz: Good catch. Added.

keymanager/third_party/bssl-sys/build.rs
atulpatildbz: added auto-init logic. 
if `CMakeLists.txt` is missing, we attempt  `git submodule update --init --recursive`

keymanager/third_party/bssl-sys/build.rs
atulpatildbz: added `assert`

NilanjanDaw (approved):

PR #647, opened 2026-02-06
[KeyManager] Add key destruction FFI and background reaper

NilanjanDaw:
Enhances key lifecycle management by introducing FFI functions for manual key destruction and a background reaper for automatic expiration.

Key Changes:
- **Key destroy FFI**: Added `key_manager_destroy_kem_key` (KPS) and `key_manager_destroy_binding_key` (WS) to allow external callers to explicitly remove keys from the registry.
- **Background Reaper**: Implemented a background thread in `KeyRegistry` that periodically scans for and purges expired KEM keys.
- **Registry initialization**: Updated KPS `KEY_REGISTRY` initialization to automatically start the reaper thread.
- **Testing**: Added unit tests for manual destruction (success, not found, null ptr) and verified reaper functionality.

This PR is built on top of #645 [[KeyManager] Implement FFI for KEM and binding key generation](https://github.com/google/go-tpm-tools/pull/645)
Please review [commit #025520d onwards](https://github.com/google/go-tpm-tools/pull/647/changes/991d285b9641efb5bd2912b69d5d945b21c80bde..9236663c3c421feafb00e96a2e41092d438d374a)

### DIFF
--- a/keymanager/km_common/src/key_types.rs
+++ b/keymanager/km_common/src/key_types.rs
@@ -87,16 +87,15 @@ impl KeyRegistry {
     ) -> std::thread::JoinHandle<()> {
         let keys_clone = Arc::clone(&self.keys);
         std::thread::spawn(move || {
-            while !stop_signal.load(std::sync::atomic::Ordering::Relaxed) {
+            loop {
                 std::thread::sleep(Duration::from_secs(REAPER_INTERVAL_SECS));
                 if stop_signal.load(std::sync::atomic::Ordering::Relaxed) {
                     break;
                 }
                 let now = Instant::now();
-                keys_clone
-                    .write()
-                    .unwrap()
-                    .retain(|_, key| key.meta.delete_after > now);
+                if let Ok(mut keys) = keys_clone.write() {
+                    keys.retain(|_, key| key.meta.delete_after > now);
+                }
             }
         })
     }


bluegate010:
This PR has lots of files that are already being reviewed in separate PRs. Will hold off on reviewing until this PR just has new files that are relevant to the reaper functionality.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -96,6 +103,35 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+/// * `0` on success.
+/// * `-1` if the UUID pointer is null or the key was not found.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i32 {
+    if uuid_bytes.is_null() {
+        return -1;
+    }
+    let uuid = unsafe {
+        let mut bytes = [0u8; 16];
+        std::ptr::copy_nonoverlapping(uuid_bytes, bytes.as_mut_ptr(), 16);
```
atulpatildbz, line 125: nit: only this line can be wrapped inside `unsafe`

keymanager/km_common/src/key_types.rs
```diff
@@ -225,4 +271,122 @@ mod tests {
...
+
+        let id = record.meta.id;
+        registry.add_key(record);
+
+        // Start reaper
+        let stop_signal = Arc::new(std::sync::atomic::AtomicBool::new(false));
+        let handle = registry.start_reaper(stop_signal.clone());
+
+        // Wait for < expiry (1s) - key should still be there
+        std::thread::sleep(Duration::from_secs(1));
```
atulpatildbz, line 371: is it possible to figure out a way use a channel to notify the test when the reaper completes a cycle?

Or expose a `trigger_reaper_scan()` private method just for tests? WDYT?

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack. Done.

keymanager/km_common/src/key_types.rs
NilanjanDaw: It might not be that straight forward since these are separate threads. But I have increased the buffer time between the events to tackle any timing issues. Let me know if you think we should introduce a channel.

keymanager/km_common/src/key_types.rs
```diff
@@ -50,13 +55,50 @@ pub type KeyHandle = Uuid;
...
+    }
+
+    /// Spawns a background reaper thread to walk through the KeyRegistry every 60s and sanitize expired keys.
+    pub fn start_reaper(
+        &self,
+        stop_signal: Arc<std::sync::atomic::AtomicBool>,
+    ) -> std::thread::JoinHandle<()> {
+        let keys_clone = Arc::clone(&self.keys);
+        std::thread::spawn(move || {
+            while !stop_signal.load(std::sync::atomic::Ordering::Relaxed) {
```
clundin25: nit:

Could just be a `loop` and rely on the condition in line 93.

keymanager/km_common/src/key_types.rs
```diff
@@ -50,13 +55,50 @@ pub type KeyHandle = Uuid;
...
+                std::thread::sleep(Duration::from_secs(REAPER_INTERVAL_SECS));
+                if stop_signal.load(std::sync::atomic::Ordering::Relaxed) {
+                    break;
+                }
+                let now = Instant::now();
+                keys_clone
+                    .write()
+                    .unwrap()
```
clundin25: Is it correct that the reaper process should die if an `Arc::write` fails? This error seems retryable, or rather than panic just ignore a failed write?

keymanager/km_common/src/key_types.rs
NilanjanDaw: ack, updated.

keymanager/km_common/src/key_types.rs
NilanjanDaw: Yeah that makes sense. Updated to ignore write fails, we will just retry the same write next time the loop runs.

PR #649, opened 2026-02-08
[KeyManager] Implement FFIs for decap-and-seal and HPKE open operations

NilanjanDaw:
This PR implements the core FFI interfaces for the Key Protection Service (KPS) and Workload Service (WS) to support the secure secret delivery flow using the `decap-and-seal` and  HPKE `open` operations. It also introduces necessary cryptographic utilities to test the end-to-end secret sharing flow.

Key Changes

**km_common (Common Crypto & Utilities)**
   - DHKEM Encapsulation: Added a "test-only" `encap` to support encapsulation operation flows with a pre-shared key. This manually implements the Encap operation as BoringSSL's current FFI lacks a direct entry point encapsulation.
   - Test Infrastructure: Introduced a test-utils feature to allow these cryptographic helpers to be shared across crates for testing purposes without exposing them in production builds.

**kps_key_custody_core (Key Protection Service)**
   - `key_manager_decap_and_seal` FFI: Implemented the decapsulate-and-seal logic. This takes a client's encapsulated key, recovers the shared secret using the KEM private key, and immediately reseals it for the target Workload Service using its binding public key.

**ws_key_custody_core (Workload Service)**
   - `key_manager_open` FFI: Implemented the final decryption step. It retrieves the Binding Private Key (isolated in memfd_secret backed storage) to decrypt the final payload delivered from the KPS.

**Testing Performed**
   - Ran cargo test -p km_common
   - Ran cargo test -p kps_key_custody_core
   - Ran cargo test -p ws_key_custody_core
   - All 40+ unit tests across the workspace are passing.

### DIFF
--- a/keymanager/km_common/src/crypto/x25519.rs
+++ b/keymanager/km_common/src/crypto/x25519.rs
@@ -185,8 +185,14 @@ impl PrivateKeyOps for X25519PrivateKey {
 
 /// LabeledExtract(salt, label, ikm) = HKDF-Extract(salt, "HPKE-v1" || suite_id || label || ikm)
 fn labeled_extract(salt: &[u8], label: &[u8], ikm: &[u8], suite_id: &[u8]) -> hkdf::Prk {
+    // print the params received
+    println!(
+        "labeled_extract: salt={:?}, label={:?}, ikm={:?}, suite_id={:?}",
+        salt, label, ikm, suite_id
+    );
+
     let labeled_ikm = SecretBox::new([b"HPKE-v1".as_slice(), suite_id, label, ikm].concat());
-    hkdf::HkdfSha256::extract(labeled_ikm.as_slice(), hkdf::Salt::NonEmpty(salt))
+    hkdf::HkdfSha256::extract(labeled_ikm.as_slice(), hkdf::Salt::None)
 }
 
 /// LabeledExpand(prk, label, info, L) = HKDF-Expand(prk, "HPKE-v1" || suite_id || label || info, L)
@@ -223,15 +229,12 @@ mod tests {
 
     #[test]
     fn test_decaps_x25519_clamped_vector() {
-        // Since BoringSSL X25519 always clamps the private key, we use vectors that
-        // are consistent with clamping.
-        // Input private key (clamped internally by BoringSSL):
-        let sk_r_hex = "468c86c75053df4d0925e01f5446700e57288f3316c5b610c3b9b94090b8f2cb";
-        let enc_hex = "1b2767097950294d300c2830366c3c58853c83a736466336e392576b9762194d";
+        // Vectors from https://www.rfc-editor.org/rfc/rfc9180.html#appendix-A.1
+        let sk_r_hex = "4612c550263fc8ad58375df3f557aac531d26850903e55a9f23f21d8534e8ac8";
+        let enc_hex = "37fda3567bdbd628e88668c3c8d7e97d1d1253b6d4ea6d44c150f741f1bf4431";
 
-        // This is what we get when we run with clamping:
         let expected_shared_secret_hex =
-            "b1e179eefbcdfe490a1929c3c6e5de6d98f3ed4463b6d94627390119610baa83";
+            "fe0e18c9f024ce43799ae393c7e8fe8fce9d218875e8227b0187c04e7d2ea1fc";
 
         let sk_r_bytes: Vec<u8> = hex::decode(sk_r_hex).unwrap();
         let sk_r = X25519PrivateKey(SecretBox::new(sk_r_bytes));
@@ -275,13 +278,12 @@ mod tests {
 
     #[test]
     fn test_encaps_x25519_clamped_vector() {
-        // Since BoringSSL X25519 always clamps the private key, we use vectors that
-        // are consistent with clamping.
-        let sk_e_hex = "1cbbe38b923164ec953f62fb8d182090a68e7b4d7bf2af0e1315284ed5cd2814";
-        let pk_e_hex = "1c7be5f4cd9e5217cec0af5d53e5512516c23ef232f8e9921c18d4cf9561f42d";
-        let pk_r_hex = "e7a1b035517828a006dcdb862da34f0a7846ba64ba94a066b3bf5f428dfa0c4c";
+        // Vectors from https://www.rfc-editor.org/rfc/rfc9180.html#appendix-A.1
+        let sk_e_hex = "52c4a758a802cd8b936eceea314432798d5baf2d7e9235dc084ab1b9cfa2f736";
+        let pk_e_hex = "37fda3567bdbd628e88668c3c8d7e97d1d1253b6d4ea6d44c150f741f1bf4431";
+        let pk_r_hex = "3948cfe0ad1ddb695d780e59077195da6c56506b027329794ab02bca80815c4d";
         let expected_shared_secret_hex =
-            "8feafe5dfc6bf39bfe8e872f31c36fd66ba9d846e6256c98e0f0f6134f867dfc";
+            "fe0e18c9f024ce43799ae393c7e8fe8fce9d218875e8227b0187c04e7d2ea1fc";
 
         let sk_e_bytes: Vec<u8> = hex::decode(sk_e_hex).unwrap();
         let pk_r_bytes: [u8; 32] = hex::decode(pk_r_hex).unwrap().try_into().unwrap();


keymanager/km_common/src/key_types.rs
```diff
@@ -0,0 +1,343 @@
...
+where
+    F: FnOnce(HpkeAlgorithm, Vec<u8>) -> KeySpec,
+{
+    let (pub_key, mut priv_key) = match KemAlgorithm::try_from(algo.kem)
+        .ok()
+        .and_then(|k| crypto::generate_x25519_keypair(k).ok())
+    {
+        Some(pair) => pair,
+        None => return Err(-1),
+    };
```
bluegate010: Below you use `map_err(|_| -1)?`, could use that here?

```suggestion
    let alg = KemAlgorithm::try_from(algo.kem).map_err(|_| -1)?;
    let (pub_key, mut priv_key) = crypto::generate_x25519_keypair(alg).map_err(|_| -1)?;
```

keymanager/km_common/src/key_types.rs
```diff
@@ -0,0 +1,343 @@
...
+use zeroize::Zeroize;
+
+#[cfg(not(test))]
+const REAPER_INTERVAL_SECS: u64 = 60;
+#[cfg(test)]
+const REAPER_INTERVAL_SECS: u64 = 1;
+
+/// Represents the purpose of the Key and its associated algorithms.
+#[derive(Clone)]
+pub enum KeySpec {
```
bluegate010, line 17: Uber-nit: we have KeySpec, KeyMetadata, and KeyRecord. We could probably do away with one of those, to cut down on cognitive overhead. Consider merging all fields of KeyMetadata into KeyRecord - and perhaps s/KeySpec/KeyMetadata/, since a "key spec" to me seems to imply algorithms and policies but not data about specific keys, and the current fields in KeySpec seem metadata-ish.

keymanager/km_common/src/key_types.rs
```diff
@@ -0,0 +1,343 @@
...
+        binding_public_key: Vec<u8>,
+    },
+}
+
+/// Internal Rust struct to hold the Key Metadata
+#[derive(Clone)]
+pub struct KeyMetadata {
+    /// UUID key handle for internal tracking
+    pub id: Uuid,
+    pub created_at: Instant,
```
bluegate010, line 39: Nit: needed?

keymanager/km_common/src/key_types.rs
```diff
@@ -0,0 +1,343 @@
...
+    ) -> std::thread::JoinHandle<()> {
+        let keys_clone = Arc::clone(&self.keys);
+        std::thread::spawn(move || {
+            while !stop_signal.load(std::sync::atomic::Ordering::Relaxed) {
+                std::thread::sleep(Duration::from_secs(REAPER_INTERVAL_SECS));
+                if stop_signal.load(std::sync::atomic::Ordering::Relaxed) {
+                    break;
+                }
+                let now = Instant::now();
+                // We only expire KemWithBindingPub keys as per requirements.
```
bluegate010: We could just generically expire any key that has a non-zero delete_after field, since otherwise we could accidentally forget to update this logic and think we're erasing keys when we aren't.

In practice we could either (a) use an expiry time of zero for Binding keys, or (b) expire both Binding and KemWithBindingPub keys. Personally I'm kind of leaning towards (b), but only attesting to the delete_after field for the KEM keypair. (I know I gave conflicting feedback before, sorry. Didn't realize we could be consistent on expiry while only attesting to the expiry of the KEM public key. Also didn't realize that if we don't expire Binding keys, we could end up with stale binding keys just cluttering the keys map if the corresponding KEM keys have since expired.)

keymanager/km_common/src/crypto.rs
```diff
@@ -0,0 +1,461 @@
...
+
+        // 7. AEAD Open
+        let aead = aead::Aes256Gcm::new(&key);
+        aead.open(&nonce, ciphertext, aad)
+            .ok_or(Error::HpkeDecryptionError)
+    })
+}
+
+/// Generates an X25519 keypair for the given KEM algorithm.
+pub fn generate_x25519_keypair(algo: KemAlgorithm) -> Result<(Vec<u8>, Vec<u8>), Error> {
```
bluegate010: I recall a [comment](https://github.com/google/go-tpm-tools/pull/643#discussion_r2782477900) in a separate PR where this was renamed to generate_keypair.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,671 @@
...
+                &mut kem_pubkey_len,
+            );
+        }
+
+        // 3. Generate a "client" ciphertext/encapsulation targeting KEM key.
+        let pt = b"ignored_plaintext";
+        let aad = b"test_aad";
+        // We use `hpke_seal` to act as the client to generate a valid encapsulation
+        let (client_enc, client_ct) =
+            km_common::crypto::hpke_seal(&kem_pubkey_bytes, pt, aad, &algo).unwrap();
```
bluegate010: I feel like we shouldn't use a raw KEM public key to do HPKE sealing.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,671 @@
...
+        let expected_shared_secret = km_common::crypto::decaps(
+            key_record.private_key.as_bytes(),
+            &client_enc,
+            KemAlgorithm::DhkemX25519HkdfSha256,
+        )
+        .expect("decaps failed");
+        assert_eq!(
+            recovered_shared_secret, expected_shared_secret,
+            "Recovered secret mismatch"
+        );
```
bluegate010, line 573: `key_manager_decap_and_seal()` invokes `km_common::crypto::decaps()`. Therefore this test ensures that two invocations of `decaps()` match each other, but doesn't demonstrate that either is cryptographically correct. This test should probably replace the call to hpke_seal() with an encaps() step. Which hpke_open_with_shared_secret() actually already has, plus some other stuff that could be dropped from the PR.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -1,14 +1,671 @@
...
+
+        // 6. Verify that this secret correctly decrypts the original client ciphertext
+        // using the shared secret directly instead of the private key.
+        let decrypted_pt = km_common::crypto::hpke_open_with_shared_secret(
+            &recovered_shared_secret,
+            &client_ct,
+            aad,
+            &algo,
+        )
+        .expect("Failed to decrypt client message with shared secret");
```
bluegate010: I wouldn't go to much effort to test this, IMO. This tests one possible way someone could use the API, but as long as we've checked that the shared secret that comes out of hpke_open() is equal to the shared secret that comes out of encaps(), job done.

keymanager/km_common/src/key_types.rs
NilanjanDaw: +1, I also think it would be good practice to clean up both the binding and the KEM keys. I have added a reaper thread for the binding keys too. I am handling these in the [Key Destroy PR](https://github.com/google/go-tpm-tools/pull/647)

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -96,6 +97,130 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+    encapsulated_key: *const u8,
+    encapsulated_key_len: usize,
+    aad: *const u8,
+    aad_len: usize,
+    out_encapsulated_key: *mut u8,
+    out_encapsulated_key_len: usize,
+    out_ciphertext: *mut u8,
+    out_ciphertext_len: usize,
+) -> i32 {
+    if uuid_bytes.is_null()
```
atulpatildbz: wondering if we should wrap the function in `std::panic::catch_unwind` to protect against undefined behaviour

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -258,4 +383,250 @@ mod tests {
...
+        // 2. Generate KEM key
+        let mut uuid_bytes = [0u8; 16];
+        let mut kem_pubkey_bytes = [0u8; 32];
+        let kem_pubkey_len = 32;
+        let algo = HpkeAlgorithm {
+            kem: KemAlgorithm::DhkemX25519HkdfSha256 as i32,
+            kdf: KdfAlgorithm::HkdfSha256 as i32,
+            aead: AeadAlgorithm::Aes256Gcm as i32,
+        };
+        unsafe {
```
atulpatildbz: can we assert that the setup call succeeds. If `key_manager_generate_kem_keypair` fails (e.g., returns -1), `uuid_bytes` will remain all zeros. The subsequent call to `decap_and_seal` would then likely fail with `-1` (Invalid UUID) instead of the expected `-2` (Buffer Too Small), leading to a confusing test failure message.

```rust
let res = unsafe {
    key_manager_generate_kem_keypair(
        algo,
        binding_pk.as_bytes().as_ptr(),
        binding_pk.as_bytes().len(),
        3600,
        uuid_bytes.as_mut_ptr(),
        kem_pubkey_bytes.as_mut_ptr(),
        kem_pubkey_len,
    )
};
assert_eq!(res, 0, "Setup failed: key generation returned error");
```

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -258,4 +383,250 @@ mod tests {
...
+        // We use `hpke_seal` to act as the client to generate a valid encapsulation
+        let kem_pub_key_obj = PublicKey::try_from(kem_pubkey_bytes.to_vec()).unwrap();
+        let pt_box = km_common::crypto::secret_box::SecretBox::new(pt.to_vec());
+        let (client_enc, client_ct) =
+            km_common::crypto::hpke_seal(&kem_pub_key_obj, &pt_box, aad, &algo).unwrap();
+
+        // Step 3: Call `decap_and_seal`.
+        let mut out_enc_key = [0u8; 32];
+        let out_enc_key_len = 32;
+        let mut out_ct = [0u8; 48]; // 32 bytes secret + 16 tag
```
atulpatildbz, line 541: can we define constant for this:
```rs
km_common::crypto::AES_256_GCM_TAG_LEN
```

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -82,6 +83,91 @@ pub unsafe extern "C" fn key_manager_generate_binding_keypair(
...
+///   On success, it will be updated with the actual size of the plaintext.
+///
+/// ## Safety
+/// This function is unsafe because it dereferences raw pointers. The caller must ensure that:
+/// * `uuid_bytes` points to a valid 16-byte buffer.
+/// * `enc` points to a valid buffer of at least `enc_len` bytes.
+/// * `ciphertext` points to a valid buffer of at least `ciphertext_len` bytes.
+/// * `aad` points to a valid buffer of at least `aad_len` bytes.
+/// * `out_plaintext` is either null or points to a valid buffer of at least `*out_plaintext_len` bytes.
+/// * `out_plaintext_len` points to a valid `usize`.
```
atulpatildbz, line 172: `out_plaintext_len` isn't a pointer right?

the code defines it as `out_plaintext_len: usize`

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -82,6 +83,91 @@ pub unsafe extern "C" fn key_manager_generate_binding_keypair(
...
+///
+/// ## Returns
+/// * `0` on success.
+/// * `-1` if arguments are invalid (e.g., key not found, null pointers).
+/// * `-2` if the `out_plaintext` buffer is too small.
+/// * `-3` if decryption failed.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_open(
```
atulpatildbz, line 180: similar suggestion here to wrap in `std::panic::catch_unwind`. WDYT?

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -82,6 +83,91 @@ pub unsafe extern "C" fn key_manager_generate_binding_keypair(
...
+    ciphertext_len: usize,
+    aad: *const u8,
+    aad_len: usize,
+    out_plaintext: *mut u8,
+    out_plaintext_len: usize,
+) -> i32 {
+    if uuid_bytes.is_null()
+        || enc.is_null()
+        || ciphertext.is_null()
+        || aad.is_null()
```
atulpatildbz: i thought `aad` is optional in HPKE? are we intentionally making in mandatory here?

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -203,4 +289,120 @@ mod tests {
...
+    fn test_key_manager_open_buffer_too_small() {
+        let mut uuid_bytes = [0u8; 16];
+        let mut pubkey_bytes = [0u8; 32];
+        let pubkey_len: usize = pubkey_bytes.len();
+        let algo = HpkeAlgorithm {
+            kem: KemAlgorithm::DhkemX25519HkdfSha256 as i32,
+            kdf: KdfAlgorithm::HkdfSha256 as i32,
+            aead: AeadAlgorithm::Aes256Gcm as i32,
+        };
+        unsafe {
```
atulpatildbz: similar check here might be useful `assert_eq!(res, 0);`

keymanager/km_common/src/crypto.rs
NilanjanDaw: Yes it has been updated in a previous PR, this fixed now.

keymanager/km_common/src/key_types.rs
NilanjanDaw: Yeah that makes sense. Let me do this is in a future PR. Filed issue #667 to track this.

keymanager/km_common/src/key_types.rs
NilanjanDaw: I kept it for completeness, incase we need this in future, for example something like `nbf` in JWT tokens.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Sorry I didn't completely get your comment. Are you talking about encoding the public key into base64 or DER?
In any case, I have encoded it into our `PublicKey` struct.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: The initial `hpke_seal` call that I make on the plain-text produces an encapsulated key.
```
let (client_enc, client_ct) =
            km_common::crypto::hpke_seal(&kem_pub_key_obj, &pt_box, aad, &algo).unwrap();
```
I wanted to check that our manually implemented `decap` works against a standard `encap` implementation and can decapsulate the encapped key. Let me know if you feel we should replace it with a encap implementation.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Yes I agree, but I wanted to test that the decap operation works and can be plugged into at least one standard decrypt workflow type (for example HPKE decrypt).

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -132,6 +132,130 @@ pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i
...
+    uuid_bytes: *const u8,
+    encapsulated_key: *const u8,
+    encapsulated_key_len: usize,
+    aad: *const u8,
+    aad_len: usize,
+    out_encapsulated_key: *mut u8,
+    out_encapsulated_key_len: usize,
+    out_ciphertext: *mut u8,
+    out_ciphertext_len: usize,
+) -> i32 {
```
bluegate010, line 237: Similar comment from other PRs - can we have this unsafe function just be responsible for marshalling into safe types, and then have it invoke a safe function to do the actual business logic?

keymanager/key_protection_service/key_custody_core/src/lib.rs
bluegate010: Right, but the key being passed to hpke_seal() in this test is not an HPKE key, nor really a KEM key if you look hard; it's a weird mash-up key that supports encaps() and decaps_and_seal(). Which is why you needed to write the big hpke_open_with_shared_secret() function, which is only needed because we're treating this mash-up key as an HPKE key.

Zooming out, what we're trying to implement is a KEM operation that happens to involve an internal HPKE step:

```
def encaps():
  kem_pub.encaps()

def decaps():
  kem_priv.decaps_and_seal()
  binding_priv.open()
```

So we could test it like that, invoking encaps() and decaps() and ensuring that the shared secrets are the same. If we want to go further and test that the shared secret can be used with a higher-level crypto operation, we could do so with a simple AES or HMAC step, using the shared secret to encrypt or MAC some text.

Feel free to capture this in an issue for future cleanup though.

keymanager/km_common/src/key_types.rs
NilanjanDaw: Resolving since the code block is not valid anymore.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack, done. Separated this function and the `key_manager_open` into unsafe logic and a separate safe business logic.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Makes sense. I cleaned up the code to use a test-only `encap` function to test the `decap_and_seal` logic.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack, correected.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Makes sense. Updated this and other functions.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Makes sense, wrapped this and others with `std::panic::catch_unwind`

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack. Done.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Ah my bad. Changed this to optional aad.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Done.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -117,21 +121,150 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+fn decap_and_seal_internal(
+    uuid: Uuid,
+    encapsulated_key: &[u8],
+    aad: &[u8],
+) -> Result<(Vec<u8>, Vec<u8>), i32> {
+    // Get key record from registry
+    let key_record = match KEY_REGISTRY.get_key(&uuid) {
+        Some(record) => record,
+        None => return Err(-1),
+    };
```
clundin25: nit:
```suggestion
    let Some(key_record) = KEY_REGISTRY.get_key(&uuid) else { Err(-1)?) };
```

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -70,37 +70,41 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
             }
-            out_uuid.copy_from_slice(id.as_bytes());
-            out_pubkey.copy_from_slice(pubkey.as_bytes());
-            0 // Success
+            Err(e) => e,
         }
-        Err(e) => e,
-    }
+    }))
+    .unwrap_or(-1)
```
clundin25, line 120: The FFI API has a pretty opaque error code interface. Any future plans for improving this? 

Once pattern I've seen before is setting a global ERRNO with an error code for the last failure, and always returning `-1` in the FFI API.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -117,21 +121,150 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+    };
+
+    let (hpke_algo, binding_public_key) = match &key_record.meta.spec {
+        KeySpec::KemWithBindingPub {
+            algo,
+            binding_public_key,
+            ..
+        } => (algo, binding_public_key),
+        _ => return Err(-1), // Wrong key type
+    };
```
clundin25: nit:
```suggestion
    let KeySpec::KemWithBindingPub { algo: hpke_algo, binding_public_key, .. } = &key_record.meta.spec else { Err(-1)? };
```

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -117,21 +121,150 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+/// * `uuid_bytes` - A pointer to the 16-byte UUID of the KEM key.
+/// * `encapsulated_key` - A pointer to the encapsulated key bytes (ciphertext from client).
+/// * `encapsulated_key_len` - The length of the encapsulated key.
+/// * `aad` - A pointer to the Additional Authenticated Data (AAD) for the sealing operation.
+/// * `aad_len` - The length of the AAD.
+/// * `out_encapsulated_key` - A pointer to a buffer where the new encapsulated key will be written.
+/// * `out_encapsulated_key_len` - A pointer to a `usize` containing the size of `out_encapsulated_key`.
+///   On success, updated with the actual size.
+/// * `out_ciphertext` - A pointer to a buffer where the sealed ciphertext will be written.
+/// * `out_ciphertext_len` - A pointer to a `usize` containing the size of `out_ciphertext`.
```
clundin25: Looks like this comment is stale.

keymanager/key_protection_service/key_custody_core/Cargo.toml
```diff
@@ -7,3 +7,6 @@ edition = "2024"
 km_common = { path = "../../km_common" }
 uuid = { version = "1.20.0", features = ["v4"] }
 zeroize = "1.8"
+
+[dev-dependencies]
+km_common = { path = "../../km_common", features = ["test-utils"] }
```
clundin25: It might be useful to use a "Cargo workspace" to manage all of these shared crates:

https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Let me do this in a future PR. We will track this in #668

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack done.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Thank you, done.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Oops, updated.

keymanager/key_protection_service/key_custody_core/Cargo.toml
NilanjanDaw: Done, updated the cargo dependencies to a workspace.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Hmm, we are using it only a couple of times I guess so was wondering if we need to define a constant?

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -3,9 +3,9 @@ use km_common::crypto::PublicKey;
 use km_common::key_types::{KeyRecord, KeyRegistry, KeySpec};
 use prost::Message;
 use std::slice;
-use std::sync::atomic::AtomicBool;
```
atulpatildbz, line 6: nit: can you format this file? i think the previous position of this import was correct

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -130,21 +134,149 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+    }))
+    .unwrap_or(-1)
+}
+
+/// Internal function to decapsulate and reseal a shared secret.
+fn decap_and_seal_internal(
+    uuid: Uuid,
+    encapsulated_key: &[u8],
+    aad: &[u8],
+) -> Result<(Vec<u8>, Vec<u8>), i32> {
```
atulpatildbz: should we avoid the intermediate allocation and  accept `&mut [u8]` instead? since caller already provides `out_encapsulated_key`, `out_ciphertext` output buffers

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -130,21 +134,149 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
+    let shared_secret = match km_common::crypto::decaps(&priv_key, encapsulated_key) {
+        Ok(s) => s,
+        Err(_) => return Err(-3),
     };
 
-    match KEY_REGISTRY.remove_key(&uuid) {
-        Some(_) => 0, // Success
-        None => -1,   // Not found
+    // Seal
+    match km_common::crypto::hpke_seal(binding_public_key, &shared_secret, aad, hpke_algo) {
```
atulpatildbz, line 186: we're using `hpke_algo` here. which assumes that binding key always supports the exact same cryptro params (KEM/KDF/AEAD) as the KEM key. 

is your comment of the other PR https://github.com/google/go-tpm-tools/pull/665#discussion_r2832895596 valid here?

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -114,18 +117,121 @@ pub unsafe extern "C" fn key_manager_generate_binding_keypair(
...
+    }))
+    .unwrap_or(-1)
+}
 
-    match KEY_REGISTRY.remove_key(&uuid) {
-        Some(_) => 0, // Success
-        None => -1,   // Not found
+/// Internal function to decrypt a ciphertext using a stored binding key.
+fn open_internal(uuid: Uuid, enc: &[u8], ciphertext: &[u8], aad: &[u8]) -> Result<SecretBox, i32> {
+    let Some(record) = KEY_REGISTRY.get_key(&uuid) else {
```
atulpatildbz: nit: we can use shorter idiom here:
```
let key_record = KEY_REGISTRY.get_key(&uuid).ok_or(-1)?;
```

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -114,18 +117,121 @@ pub unsafe extern "C" fn key_manager_generate_binding_keypair(
...
+/// * `uuid_bytes` - A pointer to a 16-byte buffer containing the key UUID.
+/// * `enc` - A pointer to the encapsulated shared secret.
+/// * `enc_len` - The length of the encapsulated shared secret.
+/// * `ciphertext` - A pointer to the ciphertext to decrypt.
+/// * `ciphertext_len` - The length of the ciphertext.
+/// * `aad` - A pointer to the additional authenticated data.
+/// * `aad_len` - The length of the additional authenticated data.
+/// * `out_plaintext` - A pointer to a buffer where the decrypted plaintext will be written.
+/// * `out_plaintext_len` - The size of `out_plaintext` buffer.
+///   On success, it will be updated with the actual size of the plaintext.
```
atulpatildbz: the signature is :
```
out_plaintext_len: usize, // Passed by value!
```

this statement is misleading: "On success, it will be updated with the actual size of the plaintext". since  it's passed by value. 

check if you need to update the signature to `*mut usize`. or fix the docstring

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: yeah, my vs-code formatter is clashing with `rustfmt`. Its comparing without ignoring case :smh:

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Yeah this is a valid point. I think we need to fix the `generatekem` FFI to also include the Binding key algorithm. That will allow us to decouple the algorithms.

NilanjanDaw: Filed #673 to track the effort.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Done.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack done.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Updated the doc-string. We are not updating the plaintext length anymore.

keymanager/km_common/src/crypto/x25519.rs
```diff
@@ -175,15 +229,12 @@ mod tests {
 
     #[test]
     fn test_decaps_x25519_clamped_vector() {
```
bluegate010, line 231: Here and the next test, is `_clamped_vector` still correct?

bluegate010 (approved):

clundin25 (approved):

keymanager/km_common/src/crypto/x25519.rs
NilanjanDaw: Yeah, I realized late but the RFC also uses clamped vectors for x25519 keys.

PR #650, opened 2026-02-09
[keymanager/wsd] Add /keys:decaps KOL API with DecapAndSeal + Open orchestration

atulpatildbz:
Implements the Go orchestration layer (KOL) for `POST /v1/keys:decap`. This key exchange endpoint allows workloads to recover a shared secret from a KEM encapsulation.

Flow:
1. Workload sends `{key_handle, ciphertext: {algorithm, ciphertext}}` to WSD
2. WSD looks up binding UUID from the KEM→Binding map (populated during `/v1/keys:generate_kem`)
3. WSD calls KPS `DecapAndSeal` — decapsulates the shared secret using the KEM private key, reseals it with the binding public key
4. WSD calls WSD KCC `Open` — unseals the shared secret using the binding private key
5. WSD returns the plaintext shared secret (base64-encoded) to the workload

**Changes:**
- **Endpoint**: Added `POST /v1/keys:decap` (aligned with API contract)
- **JSON**: Used `snake_case` for JSON fields in `DecapsRequest` and `DecapsResponse`
- **C headers**: Added `key_manager_decap_and_seal` (KPS) and `key_manager_open` (WSD) declarations
- **CGO bridges**: Added `DecapAndSeal()` and `Open()` Go wrappers for the Rust FFI functions
- **KPS service**: Extended `Service` with `DecapAndSeal` method
- **WSD server**: Added `DecapSealer`/`Opener` interfaces, `handleDecaps` handler
- **Tests**: Verified with `TestHandleDecapsSuccess` in  and Rust unit tests

## Testing

tested with encap operation manually. 
Testing generate + decap + seal + open
Code for manual testing: https://github.com/atulpatildbz/go-tpm-tools/commit/ac4b94757a7960dcf3303f462408ba92d8168548

PR #651, opened 2026-02-09
[keymanager/wsd] Add /keys:destroy Go KOL handler with KEM + binding key destruction

atulpatildbz:
Implements the Go orchestration layer (KOL) for `POST /v1/keys:destroy`, the key destruction endpoint that allows workloads to explicitly destroy a key pair.

Flow:
1. Workload sends `{key_handle: {handle: ...}}` to WSD
2. WSD looks up binding UUID from the KEM→Binding map
3. WSD calls **KPS** `DestroyKEMKey` — destroys the KEM key from the registry
4. WSD calls **WSD KCC** `DestroyBindingKey` — destroys the binding key from the registry
5. WSD removes the KEM→Binding mapping
6. Returns `204 No Content`

**Changes:**
- **C headers**: Added `key_manager_destroy_kem_key` (KPS) and `key_manager_destroy_binding_key` (WSD) declarations
- **CGO bridges**: Added `DestroyKEMKey()` and `DestroyBindingKey()` Go wrappers for the Rust FFI functions
- **KPS service**: Extended `Service` with `DestroyKEMKey` method and `KEMKeyDestroyer` interface
- **WSD server**: Added `KEMKeyDestroyer`/`BindingKeyDestroyer` interfaces, `DestroyRequest` type, `handleDestroy` handler, `/v1/keys:destroy` route registration
- **Tests**: 7 new destroy handler tests + 2 new KPS service tests + Integration test `TestIntegrationDestroyKey`

## Dependencies

This PR is built on top of:
- **PR #647** (`key_manager_destroy_ffi`) — Rust FFI changes (Base commit: `d1d6bad`)
- **PR #652** (`wsd_generate_key_go`) — Included via merge (Commit: `d7f52c6`)

## What to review

**Please review the following commits:**
1.  `d1d6bad` — `feat(keymanager): Add manual FFI headers for destroy` (C headers for destroy)
2.  `60466aa` — `[keymanager/wsd] Add /keys:destroy endpoint with KEM + binding key destruction` (Go implementation)

All other commits are from dependencies (PRs #647, #652) that are not yet merged to main but required for this build.

### Automated Tests
- Unit tests: `go test ./keymanager/workload_service/...` (Passed)
- Integration tests: `go test -tags=integration ./keymanager/workload_service/...` (Passed)

### Manual Verification (Curl)
Manually verified the destruction workflow locally:
1. **Generate Key**: `POST /v1/keys:generate_kem` -> Returns Handle.
2. **Destroy Key (First)**: `POST /v1/keys:destroy` -> Returns 204.
3. **Destroy Key (Second)**: `POST /v1/keys:destroy` -> Returns 404 (Key mapping removed).

### DIFF
--- a/keymanager/workload_service/server.go
+++ b/keymanager/workload_service/server.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"math"
@@ -396,7 +397,7 @@ func (s *Server) handleDestroy(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Step 1: Look up the binding UUID for this KEM key.
+	// Look up the binding UUID for this KEM key.
 	bindingUUID, ok := s.LookupBindingUUID(kemUUID)
 	if !ok {
 		writeError(w, fmt.Sprintf("KEM key handle not found: %s", kemUUID), http.StatusNotFound)
@@ -406,18 +407,13 @@ func (s *Server) handleDestroy(w http.ResponseWriter, r *http.Request) {
 	errKps := s.keyProtectionService.DestroyKEMKey(kemUUID)
 	errWs := s.workloadService.DestroyBindingKey(bindingUUID)
 
-	// Step 4: Remove the mapping.
+	// Remove the mapping.
 	s.mu.Lock()
 	delete(s.kemToBindingMap, kemUUID)
 	s.mu.Unlock()
 
-	if errKps != nil {
-		writeError(w, fmt.Sprintf("failed to destroy KEM key: %v", errKps), http.StatusInternalServerError)
-		return
-	}
-
-	if errWs != nil {
-		writeError(w, fmt.Sprintf("failed to destroy binding key: %v", errWs), http.StatusInternalServerError)
+	if err := errors.Join(errKps, errWs); err != nil {
+		writeError(w, fmt.Sprintf("failed to destroy keys: %v", err), http.StatusInternalServerError)
 		return
 	}
 


keymanager/workload_service/server.go
```diff
@@ -102,16 +118,19 @@ type Server struct {
 }
 
 // NewServer creates a new WSD server with the given dependencies.
-func NewServer(bindingGen BindingKeyGenerator, kemGen KEMKeyGenerator) *Server {
+func NewServer(bindingGen BindingKeyGenerator, kemGen KEMKeyGenerator, kemKeyDestroyer KEMKeyDestroyer, bindingKeyDestroyer BindingKeyDestroyer) *Server {
```
atulpatildbz: this will need change after #678  is merged.
May have to rebase over that

keymanager/key_protection_service/service.go
```diff
@@ -6,26 +6,45 @@ package keyprotectionservice
...
 	"github.com/google/uuid"
 
-	keymanager "github.com/google/go-tpm-tools/keymanager/km_common/proto"
+	algorithms "github.com/google/go-tpm-tools/keymanager/km_common/proto"
 )
 
 // KEMKeyGenerator generates KEM keypairs linked to a binding public key.
+
+// KEMKeyDestroyer destroys a KEM key by UUID.
+type KEMKeyDestroyer interface {
```
NilanjanDaw: Similar comment to PR #650: https://github.com/google/go-tpm-tools/pull/650/changes#r2872752692

keymanager/workload_service/integration_test.go
```diff
@@ -121,3 +131,72 @@ func TestIntegrationGenerateKeysUniqueMappings(t *testing.T) {
 	t.Logf("E2E uniqueness: KEM1=%s→Binding1=%s, KEM2=%s→Binding2=%s",
 		kemUUIDs[0], binding1, kemUUIDs[1], binding2)
 }
+
+func TestIntegrationDestroyKey(t *testing.T) {
```
NilanjanDaw, line 137: Should we also add an integration test to validate the auto-destroy mechanism with this PR? And also verify that calling a destroy on a auto-destroyed key will not cause problems (since the kem to binding mapping will still exist in the KOL layer).

keymanager/workload_service/server.go
```diff
@@ -255,10 +270,50 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+	}
+
+	if err := s.workloadService.DestroyBindingKey(bindingUUID); err != nil {
+		writeError(w, fmt.Sprintf("failed to destroy binding key: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	// Step 4: Remove the mapping.
+	s.mu.Lock()
+	delete(s.kemToBindingMap, kemUUID)
```
NilanjanDaw, line 412: I am wondering if we should try to remove the `kemUUID` from the `kemToBindingMap` even in case of errors. For example, if for some reason `DestroyKEMKey` or `DestroyBindingKey` fails the delete call will not go through. A subsequent decap call on the same key will pass through since the mapping exist and fail in the Rust layer. Which is okay, but would it better to try to delete the mapping to avoid such a situation?

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -61,3 +61,15 @@ func GenerateKEMKeypair(algo *keymanager.HpkeAlgorithm, bindingPubKey []byte, li
 	copy(pubkey, pubkeyBuf[:pubkeyLen])
 	return id, pubkey, nil
 }
+
+// DestroyKEMKey destroys the KEM key identified by kemUUID via Rust FFI.
+func DestroyKEMKey(kemUUID uuid.UUID) error {
+	uuidBytes := kemUUID[:]
+	rc := C.key_manager_destroy_kem_key(
+		(*C.uint8_t)(unsafe.Pointer(&uuidBytes[0])),
```
pgonda, line 75: if rc:= ...; rc :=0 {
..
}

keymanager/key_protection_service/service.go
atulpatildbz: Done

keymanager/workload_service/integration_test.go
atulpatildbz: Done! I have added a new integration test `TestIntegrationAutoDestroy`

I also verified the behavior offline: when the KEM key and binding key expire and are auto-destroyed in the Rust KCC layer, an explicit call to `/v1/keys:destroy` does not cause any problems. The Rust backend returns success, allowing the Go KOL to cleanly remove the stale mapping from its `kemToBindingMap` and return an `HTTP 204 No Content`. The new integration test ensures this behavior won't regress.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
atulpatildbz: Done

keymanager/workload_service/server.go
atulpatildbz: That makes sense.

I have updated the logic to ensure that both KEM and Binding keys are *always* attempted for destruction, and the mapping is *always* deleted from the map, regardless of individual failure outcomes. I've also updated `server_test.go` to explicitly verify this behavior.

Even if delete fails for some reason, the reaper would eventually delete expired keys right?

keymanager/workload_service/server.go
```diff
@@ -313,11 +329,12 @@ func (s *Server) generateKEMKey(w http.ResponseWriter, req GenerateKeyRequest) {
 	}
 
 	// Generate KEM keypair via KPS KOL, passing the binding public key.
-	kemUUID, _, err := s.keyProtectionService.GenerateKEMKeypair(algo, bindingPubKey, req.Lifespan.Seconds)
+	kemUUID, kemPubKey, err := s.keyProtectionService.GenerateKEMKeypair(algo, bindingPubKey, req.Lifespan.Seconds)
 	if err != nil {
 		writeError(w, fmt.Sprintf("failed to generate KEM keypair: %v", err), http.StatusInternalServerError)
 		return
 	}
+	log.Printf("Generated KEM Public Key (base64): %s", base64.StdEncoding.EncodeToString(kemPubKey))
```
atulpatildbz: need to remove this. left over during manual testing

atulpatildbz: removed

keymanager/workload_service/integration_test.go
NilanjanDaw: Awesome, thank you.

keymanager/workload_service/server.go
```diff
@@ -357,10 +373,53 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+	bindingUUID, ok := s.LookupBindingUUID(kemUUID)
+	if !ok {
+		writeError(w, fmt.Sprintf("KEM key handle not found: %s", kemUUID), http.StatusNotFound)
+		return
+	}
+
+	errKps := s.keyProtectionService.DestroyKEMKey(kemUUID)
+	errWs := s.workloadService.DestroyBindingKey(bindingUUID)
+
+	// Step 4: Remove the mapping.
```
NilanjanDaw: I think this is `Step 2` now

keymanager/workload_service/server.go
```diff
@@ -357,10 +373,53 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+
+	if errKps != nil {
+		writeError(w, fmt.Sprintf("failed to destroy KEM key: %v", errKps), http.StatusInternalServerError)
+		return
+	}
+
+	if errWs != nil {
+		writeError(w, fmt.Sprintf("failed to destroy binding key: %v", errWs), http.StatusInternalServerError)
+		return
+	}
```
NilanjanDaw: This works, but if the `errKps` fails, `errWs` is ignored. We can use `errors.Join()` to join the errors, so that both or either of them are surfaced.

atulpatildbz: Done

keymanager/workload_service/server.go
atulpatildbz: Removed `Step` from the comment as this comment pattern isn't used in our codebase

PR #652, opened 2026-02-09
[keymanager/wsd] Add Go orchestration layer (KOL) for WSD key generation

atulpatildbz:
## Summary
Implement the Go Key Orchestration Layer (KOL) for the Workload Service Daemon (WSD)
Expose POST `/v1/keys:generate_kem` endpoint that orchestrates the full key generation flow:
- WSD KOL calls WSD KCC (Rust FFI) to generate a binding keypair
- WSD KOL passes the binding public key to KPS KOL (direct Go method call)
- KPS KOL calls KPS KCC (Rust FFI) to generate a KEM keypair
- WSD KOL stores the KEM UUID → Binding UUID mapping
- Returns the KEM UUID to the workload

Extend Go CGO wrappers to match updated Rust FFI signatures (KmHpkeAlgorithm, usize)
Align request/response format with API spec (JSON body with algorithm, key_protection_mechanism, lifespan)

## Dependencies
This PR is created on top of #655

## Commits to Review
a8545e1 onwards ("feat(keymanager): orchestrate Key Generation in Go"), which contains all orchestration logic and CGO adaptations.

Details
- CGO bridges (`wskcc`, `kpskcc`): Go wrappers calling Rust FFI via CGO, returning `(uuid.UUID, []byte, error)`
- KPS KOL (`key_protection_service/service.go`): KEMKeyGenerator interface wrapping KPS KCC FFI
- WSD KOL (`workload_service/server.go`): HTTP server with `/v1/keys:generate_kem`, `kemToBindingMap`, dependency-injected interfaces
- Request validation: algorithm (DHKEM_X25519 only), key_protection_mechanism (KEY_PROTECTION_VM only), lifespan > 0

### Component Integration Tests:

```
go test -tags integration ./... passed.
```

Verifies full flow from Go Service layer -> CGO -> Rust Core FFI.
Covers:
- Binding Key Gen (WSD)
- KEM Key Gen (KPS)
- End-to-End Orchestration (WSD -> KPS)

### Manual Verification :

https://github.com/google/go-tpm-tools/commit/1647b5eb6f7cf1e82cfc98e5231f048947d443e4 : Created a manual test binary that initializes the WSD server with Rust FFI bindings and exposed the `generate_kem` endpoint for verification via curl.

```
> go run keymanager/cmd/wsd_manual_test/main.go

> curl --unix-socket /tmp/wsd.sock http://unix/v1/keys:generate_kem \
  -H "Content-Type: application/json" \
  -d '{"algorithm":"DHKEM_X25519_HKDF_SHA256", "key_protection_mechanism":"KEY_PROTECTION_VM", "lifespan":"3600s"}'

{"key_handle":{"handle":"4cbe1d70-d637-44f9-aa53-a08d5ba445e0"}}

```

### DIFF
--- a/keymanager/workload_service/server.go
+++ b/keymanager/workload_service/server.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"net"
 	"net/http"
+	"os"
 	"sync"
 
 	"github.com/google/uuid"
@@ -96,6 +97,7 @@ func NewServer(bindingGen BindingKeyGenerator, kemGen KEMKeyGenerator) *Server {
 
 // Serve starts the HTTP server listening on the given unix socket path.
 func (s *Server) Serve(socketPath string) error {
+	_ = os.Remove(socketPath)
 	ln, err := net.Listen("unix", socketPath)
 	if err != nil {
 		return fmt.Errorf("failed to listen on unix socket %s: %w", socketPath, err)


keymanager/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/google/go-tpm-tools/keymanager
+
+go 1.23.0
```
NilanjanDaw: Do we need this here or can we infer from the global go mod version?

keymanager/key_protection_service/key_custody_core/include/kps_key_custody_core.h
```diff
@@ -0,0 +1,38 @@
...
+typedef struct {
+    int32_t kem;
+    int32_t kdf;
+    int32_t aead;
+} KpsHpkeAlgorithm;
+
+// Algorithm constants matching algorithms.proto enum values.
+#define KPS_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256 1
+#define KPS_KDF_ALGORITHM_HKDF_SHA256 1
+#define KPS_AEAD_ALGORITHM_AES_256_GCM 1
```
NilanjanDaw: Can we re-use the protobuf defined in `km_common` or generate C compatible headers from them instead of re-defining these here?

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	var uuidBytes [16]byte
+	var pubkeyBuf [64]byte
+	pubkeyLen := C.size_t(len(pubkeyBuf))
+
+	algo := C.KpsHpkeAlgorithm{
+		kem:  C.KPS_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256,
+		kdf:  C.KPS_KDF_ALGORITHM_HKDF_SHA256,
+		aead: C.KPS_AEAD_ALGORITHM_AES_256_GCM,
```
NilanjanDaw: Same here, would be good if we can use the protobuf definitions, easier to keep the codes in sync.

keymanager/workload_service/key_custody_core/include/ws_key_custody_core.h
```diff
@@ -0,0 +1,35 @@
...
+typedef struct {
+    int32_t kem;
+    int32_t kdf;
+    int32_t aead;
+} HpkeAlgorithm;
+
+// Algorithm constants matching algorithms.proto enum values.
+#define KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256 1
+#define KDF_ALGORITHM_HKDF_SHA256 1
+#define AEAD_ALGORITHM_AES_256_GCM 1
```
NilanjanDaw: Similar comment as above.

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,186 @@
...
+	if req.Algorithm != 1 {
+		http.Error(w, "unsupported algorithm: only DHKEM_X25519 (1) is supported", http.StatusBadRequest)
+		return
+	}
+
+	// Validate keyProtectionMechanism: only KEY_PROTECTION_VM (2) supported.
+	if req.KeyProtectionMechanism != 2 {
+		http.Error(w, "unsupported keyProtectionMechanism: only KEY_PROTECTION_VM (2) is supported", http.StatusBadRequest)
+		return
+	}
```
NilanjanDaw: Hardcoded values. I think these should be matched against the proto definitions.

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,186 @@
...
+	Handle string `json:"handle"`
+}
+
+// GenerateKemRequest is the JSON body for POST /v1/keys:generate.
+type GenerateKemRequest struct {
+	Algorithm              int    `json:"algorithm"`
+	KeyProtectionMechanism int    `json:"keyProtectionMechanism"`
+	LifespanSecs           uint64 `json:"lifespanSecs"`
+}
+
```
NilanjanDaw, line 40: Was wondering if these should be proto defintions?

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -0,0 +1,59 @@
+//go:build cgo
+
+package kpskcc
+
+/*
+#cgo LDFLAGS: -L${SRCDIR}/../../target/release -lkps_key_custody_core
```
NilanjanDaw: This assumes the KCC is built in release mode, we should probably handle `debug` modes too?

NilanjanDaw (commented):
Thanks for the PR. Some initial comments.

keymanager/go.mod
atulpatildbz: `keymanager` is a separate nested Go module, so it must keep its own `go` directive and cannot inherit from the repo-root `go.mod`.
I removed the explicit toolchain pin and aligned this module to `go 1.22` so we don’t force a specific local toolchain.

keymanager/key_protection_service/key_custody_core/include/kps_key_custody_core.h
atulpatildbz: Good point. I've updated to generate C-compatible headers automatically using `cbindgen`.

The point about re-using protobuf definitions directly:
1. we cannot pass the `prost`-generated struct over the FFI boundary because it is not `#[repr(C)]` in `km_common::ffi`.
2. I added a `cbindgen` config to generate C headers from this type `KmHpkeAlgorithm`
3. FFi functions in keymanager/workload_service/key_custody_core/src/lib.rs now accepts `KmHpkeAlgorithm` and immediately convert it to internal `HpkeAlgorithm` (via `.into()`).

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
atulpatildbz: Addressed. cgo now uses the shared generated FFI constants/types (`C.KmHpkeAlgorithm` and `C.KM_*`) from `km_common/include/km_algorithms.h`, so there are no duplicated algorithm literals in cgo code paths.

keymanager/workload_service/key_custody_core/include/ws_key_custody_core.h
atulpatildbz: Addressed similarly here: WSD header now includes the shared generated `km_common/include/km_algorithms.h` instead of redefining algorithm constants/structs locally.

keymanager/workload_service/server.go
atulpatildbz: fixed. I removed hardcoded numeric checks and added typed enum constants in `proto_enums.go` (`KemAlgorithm*`, `KeyProtectionMechanism*`). `server.go` validation now checks against those named constants so the API contract matches proto-defined values.

keymanager/workload_service/server.go
atulpatildbz: Yes, updated accordingly. These request fields now use typed enum definitions (`KemAlgorithm`, `KeyProtectionMechanism`) that mirror the proto values, rather than untyped ints/magic numbers.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
atulpatildbz: Addressed. cgo now links both debug and release Rust output dirs (`target/debug` and `target/release`). I also corrected the relative path to the module’s actual target location (`../../target`), and validated this with integration tests.

keymanager/go.sum
```diff
@@ -0,0 +1,2 @@
+github.com/google/uuid v1.6.0 h1:NIvaJDMOsjHA8n1jAhLSgzrAzy1Hgr+hNrb57e+94F0=
```
NilanjanDaw, line 1: Does this need to be checked in, or do we `.gitignore`?

keymanager/km_common/build.rs
```diff
@@ -6,7 +6,6 @@ fn main() -> Result<()> {
     }
 
     let mut config = prost_build::Config::new();
-    config.type_attribute("HpkeAlgorithm", "#[repr(C)]");
```
NilanjanDaw, line 10: I see you removed the type_attribute injection in build.rs. Is this because `cbindgen` couldn't pick up the generated struct from `OUT_DIR`?

keymanager/key_protection_service/key_custody_core/include/kps_key_custody_core.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include <stddef.h>
+#include <stdint.h>
+#include "../../../km_common/include/km_algorithms.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+int32_t key_manager_generate_kem_keypair(KmHpkeAlgorithm algo,
```
NilanjanDaw: Can we remove the `km` prefixes from the algorithm definitions, seems a bit weird to have them

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,201 @@
...
+	Handle string `json:"handle"`
+}
+
+// ProtoDuration represents a google.protobuf.Duration in JSON (e.g. "3600s").
+type ProtoDuration struct {
+	Seconds uint64
+}
+
+// UnmarshalJSON parses a proto3 Duration JSON string (e.g. "300s", "1.5s").
+func (d *ProtoDuration) UnmarshalJSON(data []byte) error {
```
NilanjanDaw, line 42: We can probably just assume the duration will be in seconds? Instead of manually parsing the time string?

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,201 @@
...
+	defer s.mu.RUnlock()
+	id, ok := s.kemToBindingMap[kemUUID]
+	return id, ok
+}
+
+func (s *Server) handleGenerateKem(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
```
NilanjanDaw: Does Golang support function decorators? I am wondering if there is a way to handle the list of supported HTTP methods in a cleaner way.

NilanjanDaw: Something like https://go.dev/blog/routing-enhancements maybe?

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,201 @@
...
+
+	var req GenerateKemRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, fmt.Sprintf("invalid request body: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	// Validate algorithm: only DHKEM_X25519_HKDF_SHA256 supported.
+	if req.Algorithm != KemAlgorithmDHKEMX25519HKDFSHA256 {
```
NilanjanDaw: We should check against the `algorithms.proto` or its golang equivalent instead of hardcoding it here, since we will add more algorithms in future.

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,201 @@
...
+		http.Error(
```
NilanjanDaw: Nit: Should we declare a global error handler? Probably would make it easier to handle and return standard HTTP error codes.

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,201 @@
...
+			w,
+			fmt.Sprintf(
+				"unsupported algorithm: only DHKEM_X25519_HKDF_SHA256 (%d) is supported",
+				KemAlgorithmDHKEMX25519HKDFSHA256,
```
NilanjanDaw: Should be a list of supported algorithms coming from `algorithms.proto`

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,201 @@
...
+			),
+			http.StatusBadRequest,
+		)
+		return
+	}
+
+	// Validate keyProtectionMechanism: only KEY_PROTECTION_VM supported.
+	if req.KeyProtectionMechanism != KeyProtectionMechanismVM {
```
NilanjanDaw: Wondering if there is a better way to sanitize the request params. Can something like https://github.com/bufbuild/protovalidate-go for this?

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -0,0 +1,59 @@
...
+	rc := C.key_manager_generate_kem_keypair(
+		algo,
+		(*C.uint8_t)(unsafe.Pointer(&bindingPubKey[0])),
+		C.size_t(len(bindingPubKey)),
+		C.uint64_t(lifespanSecs),
+		(*C.uint8_t)(unsafe.Pointer(&uuidBytes[0])),
+		(*C.uint8_t)(unsafe.Pointer(&pubkeyBuf[0])),
+		pubkeyLen,
+	)
+	if rc != 0 {
```
NilanjanDaw, line 50: Since the rust FFI returns specific error codes, should we handle them and return better error messages to the client?

keymanager/workload_service/key_custody_core/ws_key_custody_core_cgo.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	rc := C.key_manager_generate_binding_keypair(
+		algo,
+		C.uint64_t(lifespanSecs),
+		(*C.uint8_t)(unsafe.Pointer(&uuidBytes[0])),
+		(*C.uint8_t)(unsafe.Pointer(&pubkeyBuf[0])),
+		pubkeyLen,
+	)
+	if rc != 0 {
```
NilanjanDaw, line 43: Similar comment about handling rust FFI error codes.

keymanager/go.sum
atulpatildbz: it's recommended to check-in `go.sum` file for reproducible builds.

https://go.dev/blog/using-go-modules

> Both go.mod and go.sum should be checked into version control

NilanjanDaw: Got it. Thanks.

keymanager/key_protection_service/key_custody_core/include/kps_key_custody_core.h
atulpatildbz: i'm trying to remove `KmHpkeAlgorithm` altogether and reuse `HpkeAlgorithm`. as discussed [here](https://github.com/google/go-tpm-tools/pull/655#discussion_r2811945743)

Keeping this thread open until that is done

keymanager/workload_service/server.go
atulpatildbz: makes sense. handled [here](https://github.com/google/go-tpm-tools/pull/652/changes/c85d54e9f6465a7d35f960fe2cab34fa0772d52b)

keymanager/workload_service/server.go
atulpatildbz: i don't actually see a problem with this. i had referred to the teeserver implementation https://github.com/google/go-tpm-tools/blob/main/launcher/teeserver/tee_server.go#L142

But makes sense to follow Go 1.22's routing. handled [here](https://github.com/google/go-tpm-tools/pull/652/changes/a8667c3cbbd5990cb7911bd48b6dc85517b06c4b)

keymanager/workload_service/server.go
atulpatildbz: thanks for the suggestion. added [here](https://github.com/google/go-tpm-tools/pull/652/changes/d5006e599f62d4d73aa7ccb39f9a8ee3a330df03)

keymanager/workload_service/server.go
atulpatildbz: handled it using a helper function. [here](https://github.com/google/go-tpm-tools/pull/652/changes/d3f9b5ce5d44d9d31c6b84b8ee11f1242afccf1d)

keymanager/workload_service/server.go
atulpatildbz: makes sense. handled [here](https://github.com/google/go-tpm-tools/pull/652/changes/d757900cb4cbef8448499608887e1fd7cecd322b)

keymanager/workload_service/server.go
atulpatildbz: Since `GenerateKemRequest` is a manually defined Go struct (not a generated Protobuf message), protovalidate-go cannot be used directly on it as it requires proto.Message instances

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
atulpatildbz: the current Rust FFI implementation collapses almost all logical errors (like unsupported algorithms or invalid keys) into a generic -1 code. Only buffer size mismatches return a refined code (-2). I agree we should handle this better, but it requires a prerequisite change in the Rust FFI to expose those specific error variants first.
 
 Created an issue to track this https://github.com/google/go-tpm-tools/issues/668

keymanager/workload_service/key_custody_core/ws_key_custody_core_cgo.go
atulpatildbz: tracking here: https://github.com/google/go-tpm-tools/issues/668

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
NilanjanDaw: Sounds good. Lets do this in a future PR.

keymanager/workload_service/key_custody_core/ws_key_custody_core_cgo.go
```diff
@@ -0,0 +1,61 @@
...
+	var uuidBytes [16]byte
+	var pubkeyBuf [32]byte
+	pubkeyLen := C.size_t(len(pubkeyBuf))
+
+	algo := &algorithms.HpkeAlgorithm{
+		Kem:  algorithms.KemAlgorithm_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256,
+		Kdf:  algorithms.KdfAlgorithm_KDF_ALGORITHM_HKDF_SHA256,
+		Aead: algorithms.AeadAlgorithm_AEAD_ALGORITHM_AES_256_GCM,
+	}
+
```
NilanjanDaw: I think it would be better to take the algorithm type from the request instead of hardcoding it here.

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	// Validate lifespan is positive.
+	if req.Lifespan.Seconds == 0 {
+		writeError(w, "lifespan must be greater than 0s", http.StatusBadRequest)
+		return
+	}
+
+	// Step 1: Generate binding keypair via WSD KCC FFI.
+	bindingUUID, bindingPubKey, err := s.bindingGen.GenerateBindingKeypair(req.Lifespan.Seconds)
```
NilanjanDaw: We should pass `req.Algorithm` to the  `GenerateBindingKeypair` function and use it to decide the Algorithm type.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -0,0 +1,68 @@
...
+
+	var uuidBytes [16]byte
+	var pubkeyBuf [32]byte
+	pubkeyLen := C.size_t(len(pubkeyBuf))
+
+	algo := &algorithms.HpkeAlgorithm{
+		Kem:  algorithms.KemAlgorithm_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256,
+		Kdf:  algorithms.KdfAlgorithm_KDF_ALGORITHM_HKDF_SHA256,
+		Aead: algorithms.AeadAlgorithm_AEAD_ALGORITHM_AES_256_GCM,
+	}
```
NilanjanDaw: Same comment about not hardcoding the algorithm here.

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		writeError(w, fmt.Sprintf("failed to generate binding keypair: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	// Step 2: Generate KEM keypair via KPS KOL, passing the binding public key.
+	kemUUID, _, err := s.kemGen.GenerateKEMKeypair(bindingPubKey, req.Lifespan.Seconds)
```
NilanjanDaw: Same comment as above.

NilanjanDaw (changes requested):

keymanager/workload_service/key_custody_core/ws_key_custody_core_cgo.go
atulpatildbz: Done.
`GenerateBindingKeypair`  function now accepts a `*algorithms.HpkeAlgorithm`

keymanager/workload_service/server.go
atulpatildbz: Done. request algorithm is now mapped to a full `HpkeAlgorithm` suite and passed down to the CGO layer

Since GenerateBindingKeypair is a core layer, i thought it'd be better to translate External API into internal `HpkeAlgorithm` proto here and then pass it down

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
atulpatildbz: made change to accept the *algorithms.HpkeAlgorithm argument

keymanager/workload_service/server.go
atulpatildbz: Done. it now uses the parameterized algorithm

keymanager/workload_service/server.go
```diff
@@ -0,0 +1,193 @@
...
+	mux := http.NewServeMux()
+	mux.HandleFunc("POST /v1/keys:generate_kem", s.handleGenerateKem)
+
+	s.httpServer = &http.Server{Handler: mux}
+	return s
+}
+
+// Serve starts the HTTP server listening on the given unix socket path.
+func (s *Server) Serve(socketPath string) error {
+	ln, err := net.Listen("unix", socketPath)
```
NilanjanDaw, line 101: minor nit: If the WSD server crashed previously or was shut down ungracefully, the .sock file will still exist on the filesystem, probably a good idea to preemptively clean it up before starting to Listen?
```
import "os"
// ...
_ = os.Remove(socketPath) // Clean up dead socket file
ln, err := net.Listen("unix", socketPath)
```

atulpatildbz: done https://github.com/google/go-tpm-tools/pull/652/changes/90cd4c6afb5b81b484695ff76509ee474c5126e9

PR #655, opened 2026-02-10
[keymanager] Generate Rust FFI headers with cbindgen

atulpatildbz:
This PR adds C header for the key custody FFI layer and fixes abi mismatch to ensure safe interoperability with C/Go.

### Changes

1.  **Automated FFI Header Generation**:
    *   Added `keymanager/generate_ffi_headers.sh` to generate C headers for all key custody components using `cbindgen`.
    *   Added `cbindgen.toml` configurations for `kps_key_custody_core`, `ws_key_custody_core`, and `km_common`.
    *   Exposed `km_common::ffi` with stable C-ABI structs and constants to ensure compatibility.

2.  **FFI ABI Fix**:
    *   Previous FFI functions (`key_manager_generate_kem_keypair`, `key_manager_generate_binding_keypair`) accepted `HpkeAlgorithm` by value. `HpkeAlgorithm` is a Protobuf-generated struct which is `#[repr(Rust)]` by default. Passing it by value across FFI caused undefined behavior because C callers expected a standard C struct layout, but Rust used its own internal layout.
    *   Updated FFI signatures to accept `KmHpkeAlgorithm`, a dedicated `#[repr(C)]` struct.
    *   Implemented safe conversion (`impl From<KmHpkeAlgorithm> for HpkeAlgorithm`) to bridge the FFI boundary safely.
    *   Added FFI-specific tests covering `KmHpkeAlgorithm` usage.

**List of manual changes**:
*   `keymanager/generate_ffi_headers.sh`: New script to automate header generation.
*   `*/cbindgen.toml`: New configuration files for `cbindgen`.
*   `keymanager/km_common/src/ffi.rs`: New file defining stable `#[repr(C)]` structs (like `KmHpkeAlgorithm`).
*   `keymanager/*/src/lib.rs`: Modified FFI functions to use the safe `KmHpkeAlgorithm` struct and perform conversion.

**Auto-Generated changes**:
*   `keymanager/km_common/include/km_algorithms.h`: Generated C header for common algorithms.
*   `keymanager/key_protection_service/key_custody_core/include/kps_key_custody_core.h`: Generated C header for KPS.
*   `keymanager/workload_service/key_custody_core/include/ws_key_custody_core.h`: Generated C header for WS.

### Verification
*   **Tests**: Verified that `cargo test` passes in `keymanager/`. Added new test cases to `kps_key_custody_core` and `ws_key_custody_core` that invoke the FFI functions with the correct `KmHpkeAlgorithm` struct.
*   **Headers**: Verified that `generate_ffi_headers.sh` produces headers where the C struct definition matches the Rust FFI signature.

### DIFF
--- a/keymanager/workload_service/key_custody_core/src/lib.rs
+++ b/keymanager/workload_service/key_custody_core/src/lib.rs
@@ -1,6 +1,7 @@
 use km_common::algorithms::HpkeAlgorithm;
 use km_common::crypto::PublicKey;
 use km_common::key_types::{KeyRecord, KeyRegistry, KeySpec};
+use prost::Message;
 use std::slice;
 use std::sync::LazyLock;
 use std::time::Duration;
@@ -53,7 +54,6 @@ fn generate_binding_keypair_internal(
 /// * `0` on success.
 /// * `-1` if an error occurred during key generation.
 /// * `-2` if the `out_pubkey` buffer size does not match the key size.
-use prost::Message;
 
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn key_manager_generate_binding_keypair(


keymanager/km_common/Cargo.toml
```diff
@@ -10,7 +10,7 @@ prost = "0.13"
 uuid = { version = "1.20.0", features = ["v4", "serde"] }
 zeroize = { version = "1.8.2", features = ["derive"] }
 thiserror = "2.0"
-bssl-crypto = { path = "../third_party/bssl-crypto" }
+bssl-crypto = { path = "../boringssl/rust/bssl-crypto" }
```
NilanjanDaw: Are we changing the boringssl build system?

keymanager/km_common/src/ffi.rs
```diff
@@ -0,0 +1,30 @@
+use crate::algorithms::{AeadAlgorithm, HpkeAlgorithm, KdfAlgorithm, KemAlgorithm};
+
+// KmHpkeAlgorithm is the stable C ABI representation of algorithms::HpkeAlgorithm.
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct KmHpkeAlgorithm {
+    pub kem: i32,
+    pub kdf: i32,
+    pub aead: i32,
+}
```
NilanjanDaw: Instead of having 2 rust struct definitions, would it be better to use `#[repr(C)]` or something like this: https://github.com/google/go-tpm-tools/blob/e08f358218750c89d992c2f80f3724f4f0cf96da/keymanager/km_common/build.rs#L10 so that the original structs are C compatible?

keymanager/km_common/Cargo.toml
NilanjanDaw: Any reason why we are changing the rust path?

NilanjanDaw (commented):
Thanks for working on this! A few initial comments.

keymanager/km_common/Cargo.toml
atulpatildbz: I've reverted this change. it wasn't needed

atulpatildbz: I had removed this changed in the latest push, not sure why it showed up for your review.

keymanager/km_common/src/ffi.rs
atulpatildbz: `HpkeAlgorithm` generated by prost-build is in `OUT_DIR` which `cbindgen` can't read. 

Based on our offline discussion here are some approaches to try:

1. using [Macro expansion](https://rustc-dev-guide.rust-lang.org/macro-expansion.html) with `parse_expand`
2. Using cbindgen as a Build Script
3. commit the algorithms.rs files that is generated by prost-build. Need to check if this is the recommended way though

I'll post updates on this thread after a few experiments / reading.

atulpatildbz: i tried following approaches.

- Macro expansion: this works but the feature is unstable and needs `RUSTC_BOOTSTRAP=1`v
- cbindgen as a Build Script - works. but also requires `RUSTC_BOOTSTRAP=1`
- commit the generated algorithms.rs - works but need to check if this is acceptable. we're changing the default OUT_DIR in this case
- https://github.com/google/go-tpm-tools/commit/40886ad7221e2dad18afedb4344412831b0ed1d0#diff-11e48d07ec74ec1bb0574c9a80a87d78c174fe48a03034259dd1f860f34b8ba9 - pass pointer + length instead of proto in the argument and let go caller marshal and rust function unmarshal. see `lib.go` files were the args are chagned to receive pointer and length

Will read more and try to finalize one approach tomorrow. Leaving it here in case someone has an opinion. \
option 3 and 4 in the above list are both looking fine to me right now. i might go with one of those

atulpatildbz: went ahead with option 4. [here](https://github.com/google/go-tpm-tools/pull/655/changes/f76535452cccb038e73fe2a31694a84d5026d84e)

keymanager/km_common/src/lib.rs
```diff
@@ -3,5 +3,6 @@ pub mod algorithms {
 }
 
 pub mod crypto;
+
```
NilanjanDaw: nit: delete extra line?

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -53,9 +53,12 @@ fn generate_kem_keypair_internal(
 /// * `0` on success.
 /// * `-1` if an error occurred during key generation or if `binding_pubkey` is null/empty.
 /// * `-2` if the `out_pubkey` buffer size does not match the key size.
+use prost::Message;
```
NilanjanDaw: I think this should be moved to the top of the file.

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -51,23 +51,32 @@ fn generate_binding_keypair_internal(
 /// * `0` on success.
 /// * `-1` if an error occurred during key generation.
 /// * `-2` if the `out_pubkey` buffer size does not match the key size.
+use prost::Message;
```
NilanjanDaw: Same comment as above.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: We need to update the doc-string here.
```
* `algo_ptr` - Pointer to a serialized HpkeAlgorithm protobuf message to use for the keypair.
* `algo_len`- Length of the serialized message.
```

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Same here lets update teh docstring.

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: done [here](https://github.com/google/go-tpm-tools/pull/655/changes/4a0706b4f7872a5680be87fb55f129b1b8a3085e)

keymanager/workload_service/key_custody_core/src/lib.rs
atulpatildbz: done [here](https://github.com/google/go-tpm-tools/pull/655/changes/4a0706b4f7872a5680be87fb55f129b1b8a3085e)

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: done [here](https://github.com/google/go-tpm-tools/pull/655/changes/4a0706b4f7872a5680be87fb55f129b1b8a3085e)

keymanager/km_common/src/lib.rs
atulpatildbz: done [here](https://github.com/google/go-tpm-tools/pull/655/changes/4a0706b4f7872a5680be87fb55f129b1b8a3085e)

keymanager/workload_service/key_custody_core/src/lib.rs
atulpatildbz: done [here](https://github.com/google/go-tpm-tools/pull/655/changes/4a0706b4f7872a5680be87fb55f129b1b8a3085e)

PR #659, opened 2026-02-12
Measure GPU attestation evidence into RTMR with a new CEL event

Sibcgh:
Binding GPU measurements into RTMR. Measurements need to completed before GPU is in ready state to avoid malicious GPU workload from potentially tampering with GPU measurements before CPU can adds measurements into RTMR.


launcher/container_runner.go
```diff
@@ -421,6 +421,17 @@ func (r *ContainerRunner) measureMemoryMonitor() error {
 	return nil
 }
 
+// TODO: Test functionality once GPU support is available.
+// measureGPUAttestationEvidence will measure GPU attestation claims into the COS
+// eventlog in the AttestationAgent.
+func (r *ContainerRunner) measureGPUAttestationEvidence() error {
+	if err := r.attestAgent.MeasureEvent(cel.CosTlv{EventType: cel.GPUDeviceAttestationBindingType, EventContent: []byte{}}); err != nil {
```
yawangwang: Let's define the GPU evidence data structure first because this PR seems to depend on it. 

The GPU evidence will be collected first, translated into the known data structure in launcher/teeserver/models/attestation.go, and then serialized as JSON bytes.

Sibcgh: https://github.com/google/go-tpm-tools/pull/664 Draft PR for that change

go.mod
```diff
@@ -12,9 +12,11 @@ require (
 	github.com/google/go-tdx-guest v0.3.2-0.20250814004405-ffb0869e6f4d
 	github.com/google/go-tpm v0.9.6
 	github.com/google/logger v1.1.1
-	google.golang.org/protobuf v1.35.1
+	google.golang.org/protobuf v1.35.2
 )
 
+require github.com/GoogleCloudPlatform/confidential-space/server v0.0.0-20260306205006-9e9ec5040a5f // indirect
```
jkl73: not sure why it's here instead of in the next section, might want to run go mod tidy

launcher/internal/gpu/driverinstaller.go
```diff
@@ -176,32 +176,61 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
...
+	if ccEnabled == attest.GPUDeviceCCMode_ON {
+		di.logger.Info("Confidential Compute is ON, setting GPU state to READY")
+		setGPUStateCmd := NvidiaSmiOutputFunc("conf-compute", "-srs", "1")
+		if err = setGPUStateToReady(setGPUStateCmd); err != nil {
+			return ccEnabled, fmt.Errorf("failed to set the GPU state to ready: %v", err)
+		}
+	}
+
+	return ccEnabled, nil
+}
```
yawangwang, line 188: why duplicating this code? And is this being referenced by somewhere?

launcher/container_runner.go
```diff
@@ -448,6 +455,51 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+// eventlog in the AttestationAgent.
+func (r *ContainerRunner) measureGPUAttestationEvidence() error {
+	if !r.launchSpec.InstallGpuDriver {
+		return nil
+	}
+
+	// We use an empty nonce for the measurement binding as the measurement itself
+	// is the binding to the TEE's RTMRs.
+	nonce := make([]byte, 32)
+	rng := rand.ChaCha8{}
```
yawangwang: it's safer to use "crypto/rand" when generating random numbers.

launcher/container_runner.go
```diff
@@ -448,6 +455,51 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to collect GPU evidence: %w", err)
+	}
+
+	gpuEvidence, ok := evidence.(*attestationpb.NvidiaAttestationReport)
+	if !ok {
+		return fmt.Errorf("unexpected evidence type: %T", evidence)
+	}
+
+	// Ensure the report reflects the nonce used for the signature.
```
yawangwang: this comment is misleading; the nonce is used to generate a raw attestation report signed by nvidia.

launcher/container_runner.go
```diff
@@ -448,6 +455,51 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+	gpuEvidence.Nonce = nonce
```
yawangwang: You should set the nonce within this func https://github.com/google/go-tpm-tools/blob/c9fbf6f92c713424062c8defcbe0304d0f94f0e4/launcher/internal/gpu/attestation.go#L44

launcher/container_runner.go
```diff
@@ -448,6 +455,51 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+
+	// Measure the evidence into the TEE's RTMR via the Attestation Agent.
+	event := cel.CosTlv{
+		EventType:    cel.GPUDeviceAttestationBindingType,
+		EventContent: []byte(gpuEvidence.String()),
```
yawangwang: you should use `proto.Marshal` to serialize

launcher/container_runner.go
```diff
@@ -448,6 +455,51 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
 	return nil
 }
 
+// measureGPUAttestationEvidence will measure GPU attestation claims into the COS
+// eventlog in the AttestationAgent.
+func (r *ContainerRunner) measureGPUAttestationEvidence() error {
```
yawangwang, line 472: Could you add unit tests? See example in https://github.com/google/go-tpm-tools/blob/c9fbf6f92c713424062c8defcbe0304d0f94f0e4/launcher/container_runner_test.go#L570

server/eventlog.go
```diff
@@ -449,6 +449,7 @@ func getVerifiedCosState(coscel gecel.CEL, registerType gecel.MRType) (*pb.Attes
 				return nil, fmt.Errorf("unknown GPU device CC mode in COS eventlog: %s", string(cosTlv.EventContent))
 			}
 			cosState.GpuDeviceState.CcMode = pb.GPUDeviceCCMode(ccMode)
+		case cel.GPUDeviceAttestationBindingType:
```
yawangwang, line 455: Try to unmarshal the `cosTlv.EventContent`, then return error if deserialization failed.

server/eventlog.go
yawangwang, line 348: Update the unit test w/ the new event type https://github.com/google/go-tpm-tools/blob/c9fbf6f92c713424062c8defcbe0304d0f94f0e4/server/eventlog_test.go#L862

yawangwang (changes requested):

yawangwang:
Please fix the failed CI test failure too.

launcher/internal/gpu/attestation.go
```diff
@@ -104,6 +109,7 @@ func (a *NvidiaAttester) collectAttestationEvidence(handler gpu.NvmlHandler, non
 					GpuQuote: gpuInfos[0],
 				},
 			},
+			Nonce: nonce,
```
yawangwang: Use `nvNonce`

launcher/internal/gpu/attestation.go
```diff
@@ -112,6 +118,7 @@ func (a *NvidiaAttester) collectAttestationEvidence(handler gpu.NvmlHandler, non
 					GpuQuotes: gpuInfos,
 				},
 			},
+			Nonce: nonce,
```
yawangwang: ditto

launcher/container_runner.go
```diff
@@ -448,6 +465,50 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+	if _, err := cryt.Read(nonce); err != nil {
+		return fmt.Errorf("failed to generate random nonce: %v", err)
+	}
+
+	evidence, err := r.gpuAttester.Attest(nonce)
+	if err != nil {
+		return fmt.Errorf("failed to collect GPU evidence: %w", err)
+	}
+
+	// Your existing type assertion logic
```
yawangwang: Remove this comment

launcher/internal/gpu/driverinstaller.go
```diff
@@ -176,29 +176,35 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
...
 		}
 	}
 
+	di.logger.Info("GPU driver installation completed successfully")
+	return nil
+}
+
+// EnableGPUReadyState checks the Confidential Computing mode and transitions
+// the GPU to a READY state if CC is enabled.
+func EnableGPUReadyState() error {
```
yawangwang: there's no need to export it as gpu attester will be the only caller

launcher/container_runner.go
```diff
@@ -305,6 +316,7 @@ func NewRunner(ctx context.Context, cdClient *containerd.Client, token oauth2.To
 		launchSpec,
 		attestAgent,
 		logger,
+		&gpu.NvidiaAttester{},
```
yawangwang: See this comment https://github.com/google/go-tpm-tools/pull/659/changes#r2914416344. Don't blindly create a GPU attester, you should create one only when the "InstallGpuDriver" is true.

launcher/container_runner.go
yawangwang: Could you pass the gpu attester created here to the "ContainerRunner"?

launcher/container_runner.go
```diff
@@ -448,6 +465,50 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
 	return nil
 }
 
+// measureGPUAttestationEvidence will measure GPU attestation claims into the COS
+// eventlog in the AttestationAgent.
+func (r *ContainerRunner) measureGPUAttestationEvidence() error {
+	if !r.launchSpec.InstallGpuDriver {
+		return nil
+	}
```
yawangwang, line 475: We can remove this check since GPU attester will be created only when "InstallGpuDriver" is set as true.

And we need to add the nil check for "gpuAttester"

launcher/container_runner.go
```diff
@@ -52,9 +56,16 @@ type ContainerRunner struct {
 	launchSpec    spec.LaunchSpec
 	attestAgent   agent.AttestationAgent
 	logger        logging.Logger
+	gpuAttester   GPUAttester
 	serialConsole *os.File
 }
 
+// GPUAttester defines the interface for GPU attestation.
+type GPUAttester interface {
```
yawangwang: Do not export this interface as this is solely for testing.

launcher/container_runner.go
```diff
@@ -448,6 +469,44 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
 
+// measureGPUAttestationEvidence will measure GPU attestation claims into the COS
+// eventlog in the AttestationAgent.
+func (r *ContainerRunner) measureGPUAttestationEvidence() error {
+	nonce := make([]byte, 32)
+	if _, err := cryt.Read(nonce); err != nil {
+		return fmt.Errorf("failed to generate random nonce: %v", err)
+	}
+
+	evidence, err := r.gpuAttester.Attest(nonce)
```
yawangwang, line 482: You should first check if `r.gpuAttester` is nil. If it's nil (meaning InstallGpuDriver is false) then it should be a no-op operation.

launcher/container_runner.go
```diff
@@ -381,6 +395,13 @@ func (r *ContainerRunner) measureCELEvents(ctx context.Context) error {
 	if err := r.measureContainerClaims(ctx); err != nil {
 		return fmt.Errorf("failed to measure container claims: %v", err)
 	}
+
+	if r.launchSpec.InstallGpuDriver {
```
yawangwang: Remove this check

launcher/container_runner.go
```diff
@@ -3,6 +3,7 @@ package launcher
 
 import (
 	"context"
+	cryt "crypto/rand"
```
alexmwu, line 6: why do you need to rename the import?

launcher/container_runner.go
```diff
@@ -52,9 +56,16 @@ type ContainerRunner struct {
 	launchSpec    spec.LaunchSpec
 	attestAgent   agent.AttestationAgent
 	logger        logging.Logger
+	gpuAttester   gpuAttester
 	serialConsole *os.File
 }
 
+// gpuAttester defines the interface for GPU attestation.
+type gpuAttester interface {
```
alexmwu, line 64: This should not live in container_runner. ContainerRunner should call the attestation agent to get GPU attestation, but this can be a follow up PR

launcher/container_runner.go
Sibcgh: math/rand is used [elsewhere](https://github.com/Sibcgh/go-tpm-tools/blob/c5f201dca41dbfba84281336089b3d8f16a8158d/launcher/container_runner.go#L572-L573) in this function so used crypt alias for `"crypto/rand" on [this line ](https://github.com/Sibcgh/go-tpm-tools/blob/c5f201dca41dbfba84281336089b3d8f16a8158d/launcher/container_runner.go#L478)

launcher/container_runner.go
Sibcgh: Will create a bug for this tomorrow

PR #665, opened 2026-02-15
[KeyManager] FFI for Enumerate keys

atulpatildbz:
## Summary
Implements FFI for the Enumerate Keys API in the Key Custody Core (KCC). It exposes the necessary Rust functions to allow the Go Workload Service to list active KEM keys.

## Dependency
- Created on top of PR #652 (Key Generation KOL changes)

## Key Changes
1.  **Key Custody Core (`key_custody_core`)**:
    -   Added `key_manager_enumerate_kem_keys` FFI function.
    -   Implemented `enumerate_kem_keys` in `KeyRegistry` to safely expose key metadata.
    -   Added `KmHpkeAlgorithm` struct for C compatibility.

2.  **C Header (`kps_key_custody_core.h`)**:
    -   Exported `key_manager_enumerate_kem_keys` definition.

## Verification
-   **Rust Tests**:
    -   Ran `cargo check` and `cargo test` in `keymanager/key_protection_service/key_custody_core`.
    -   Verified that the new FFI function compiles and integrates with the existing system.


keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -145,6 +145,81 @@ pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i
...
+    if out_entries.is_null() || out_count.is_null() {
+        return -1;
+    }
+
+    let metas = KEY_REGISTRY.enumerate_keys();
+    let now = Instant::now();
+    let mut count = 0usize;
+
+    for meta in &metas {
+        if count >= max_entries {
```
NilanjanDaw: Do we need to enable pagination in future otherwise we will be limited to the first `max_entries` keys.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -145,6 +145,81 @@ pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i
     }
 }
 
+#[repr(C)]
+pub struct KpsKeyInfo {
+    pub uuid: [u8; 16],
+    pub algorithm: [u8; 64],
```
NilanjanDaw: The KEM keys and the Binding keys will have two different sets of algorithms. How are you planning to map them?

keymanager/km_common/src/key_types.rs
```diff
@@ -86,19 +86,22 @@ impl KeyRegistry {
...
+                break;
+            }
+            let now = Instant::now();
+            if let Ok(mut keys) = keys_clone.write() {
+                keys.retain(|_, key| key.meta.delete_after > now);
             }
         })
     }
+
+    pub fn enumerate_keys(&self) -> Vec<KeyMetadata> {
```
NilanjanDaw: If we are sending `max_entries` through the FFI interface, we should probably use it here to limit the number of keys cloned instead of cloning the entire keyset.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -145,6 +145,81 @@ pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i
...
+            binding_public_key,
+        } = &meta.spec
+        {
+            let algo_bytes = algo.encode_to_vec();
+            if kem_public_key.as_bytes().len() > 64
+                || binding_public_key.as_bytes().len() > 64
+                || algo_bytes.len() > 64
+            {
+                continue;
+            }
```
NilanjanDaw: This will silently discard large keys, for example for PQC algorithms like ML-KEM the keys can be hundreds of bytes long. We should either add an error or a warning log here.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: We would also need an `offset` in that case.

atulpatildbz: Added `offset` and `max_entries` params to support pagination. The underlying `list_kem_keys` implementation in the registry now uses these to skip and limit results

keymanager/km_common/src/key_types.rs
atulpatildbz: Fixed. We now pass `max_entries` down to `KEY_REGISTRY.list_kem_keys`, which filters, skips, and takes before cloning the metadata. This ensures we only clone the number of keys requested

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: Added an explicit size check. If a key or algorithm encoding exceeds the buffer limits (2048 bytes for key, 128 for algo), we now log a warning to stderr (eprintln!) and skip that key to prevent partial/corrupt data

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: As per discussion, we've removed binding key from enumerate.

but this is a valid point. we currently enforce that if the KEM is X25519, the binding key must also be a valid 32-byte X25519 public key. If we support other algorithms in future, we'll have to change the checks here i think: https://github.com/google/go-tpm-tools/blob/main/keymanager/km_common/src/key_types.rs#L124

keymanager/km_common/src/key_types.rs
```diff
@@ -246,6 +294,63 @@ mod tests {
...
+        let page1 = registry.list_kem_keys(0, 1);
+        assert_eq!(page1.len(), 1);
+        let page2 = registry.list_kem_keys(1, 1);
+        assert_eq!(page2.len(), 1);
+        assert_ne!(page1[0].id, page2[0].id);
+
+        // Verify order (record1 created before record2)
+        // Note: created_at might be identical if executed very fast, so sort checks ID too.
+        // But record1 and record2 are created with explicit calls, so created_at likely differs slightly.
+        // Either way, stability is guaranteed.
```
NilanjanDaw: Do you want to add some asserts here? Otherwise lets remove this comment.

keymanager/km_common/src/key_types.rs
```diff
@@ -86,19 +86,40 @@ impl KeyRegistry {
...
+    /// Lists only KEM keys with pagination support.
+    pub fn list_kem_keys(&self, offset: usize, limit: usize) -> Vec<KeyMetadata> {
+        let keys = self.keys.read().unwrap();
+        let mut refs: Vec<&Arc<KeyRecord>> = keys
+            .values()
+            .filter(|k| matches!(k.meta.spec, KeySpec::KemWithBindingPub { .. }))
+            .collect();
+
+        // Sort for stable pagination: created_at, then id
+        refs.sort_by(|a, b| {
```
NilanjanDaw, line 108: This is a `O(NlogN)` sort for each offset look-up. We should probably maintain a `BTreeSet` index ordered by `(created_at, id)` for efficient iteration. But this is probably an over optimisation since the number of keys are probably gonna be small.

This can be looked after in future.

keymanager/km_common/src/key_types.rs
atulpatildbz: Thanks for pointing it out. added.

keymanager/km_common/src/key_types.rs
atulpatildbz: Ack. I'll create an issue to evaluate if this needs to be optimized https://github.com/google/go-tpm-tools/issues/674

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -145,6 +145,75 @@ pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i
...
+                .copy_from_slice(kem_public_key.as_bytes());
+            entry.kem_pub_key_len = kem_public_key.as_bytes().len();
+
+            entry.remaining_lifespan_secs = remaining;
+
+            count += 1;
+        }
+    }
+
+    unsafe {
```
bluegate010: The whole function is marked as unsafe; is this needed?

Also: would it make sense to fold out the business logic into a safe function, or would that just add overhead?

keymanager/km_common/src/key_types.rs
```diff
@@ -186,6 +207,33 @@ mod tests {
     use super::*;
     use crate::algorithms::{AeadAlgorithm, KdfAlgorithm, KemAlgorithm};
 
+    fn create_key_record<F>(
```
bluegate010, line 216: Should this function be marked as test-only?

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -145,6 +145,75 @@ pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i
...
+    pub remaining_lifespan_secs: u64,
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_enumerate_kem_keys(
+    out_entries: *mut KpsKeyInfo,
+    max_entries: usize,
+    offset: usize,
+    out_count: *mut usize,
+) -> i32 {
```
bluegate010: Should there be an output parameter indicating whether there are more entries the caller should ask for?

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -145,6 +145,75 @@ pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i
...
+                    meta.id,
+                    algo_bytes.len(),
+                    kem_public_key.as_bytes().len()
+                );
+                continue;
+            }
+
+            let remaining = meta.delete_after.saturating_duration_since(now).as_secs();
+
+            let entry = unsafe { &mut *out_entries.add(count) };
```
zhalvorsen: I'd recommend using the pattern used in previous FFI functions for this array. It will avoid unnecessarily using unsafe code here and allows the compiler to give more guarantees about buffer overflows.

```Rust
let entries = unsafe { slice::from_raw_parts_mut(out_entries, max_entries) };
```

This has the added benefit of being able to use the `zip` function on the two iterators so you can traverse both at the same time.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -145,6 +145,75 @@ pub unsafe extern "C" fn key_manager_destroy_kem_key(uuid_bytes: *const u8) -> i
...
+    pub kem_pub_key_len: usize,
+    pub remaining_lifespan_secs: u64,
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_enumerate_kem_keys(
+    out_entries: *mut KpsKeyInfo,
+    max_entries: usize,
+    offset: usize,
+    out_count: *mut usize,
```
zhalvorsen: When working with pointers for an individual item (i.e. not an array), you can use this cool "optimization" to avoid using unsafe with this parameter https://doc.rust-lang.org/nomicon/ffi.html#the-nullable-pointer-optimization.

In other words you can use an option here instead of a pointer and the check for `Some` will automatically and safely check for null. I'm not sure if cbindgen will be able to handle this correctly though.

zhalvorsen: I'm not sure if this breaks the pattern for the other APIs you have, but you could also change this parameter to be the function output value. Negative numbers indicate an error and positive number indicate the number of entries written to the array.

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: Good idea. I've extracted the business logic into a new safe internal function `enumerate_kem_keys_internal`. ffi wrapper `key_manager_enumerate_kem_keys` is still kept unsafe.

Thanks

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: Thanks for the tip!
`cbindgen` handles `Option<&mut T>` perfectly, generating the identical `T *` C-signature. applied this change to a new `out_has_more` parameter (`Option<&mut bool>`).

Also updated the return type to return the count of items written on success, and left negative values for errors.

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: Yes, good catch. I've added an `out_has_more` parameter to the FFI signature to indicate if there are remaining keys.

also added test `test_enumerate_kem_keys_has_more`

keymanager/km_common/src/key_types.rs
atulpatildbz: I thought it isn't necessary since `create_key_record` is defined inside `#[cfg(test)] mod tests { ... }` block?

let me know if my understanding is wrong.

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: thanks for the recommendation. updated the implementation to use `slice::from_raw_parts_mut` and updated the loop to safely map over the entries and the valid metas using `.zip()`. ([ref](https://github.com/google/go-tpm-tools/pull/665/changes#diff-6490673ad3110050364077e4c88a2a98ea3fc15881bf91104be14baba465fc6bR198))

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -134,151 +130,112 @@ pub unsafe extern "C" fn key_manager_generate_kem_keypair(
...
-    // Seal
-    match km_common::crypto::hpke_seal(binding_public_key, &shared_secret, aad, hpke_algo) {
-        Ok((enc, ct)) => {
-            if out_encapsulated_key.len() != enc.len() || out_ciphertext.len() != ct.len() {
-                return Err(-2);
-            }
-            out_encapsulated_key.copy_from_slice(&enc);
-            out_ciphertext.copy_from_slice(&ct);
-            Ok(())
+        if kem_public_key.as_bytes().len() > 2048 || algo_bytes.len() > 128 {
```
clundin25: Would this be an implementation error?

atulpatildbz: Yes. if generated key ever exceeds the sized defined in [KpsKeyInfo](https://github.com/google/go-tpm-tools/pull/665/changes#diff-6490673ad3110050364077e4c88a2a98ea3fc15881bf91104be14baba465fc6bR149-R156), it means there's an implementation error.

i've added a `debug_assert!` here. let me know if you have any other suggestions.

keymanager/km_common/src/key_types.rs
bluegate010: Oh I missed that part, Github hid those lines.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -196,6 +196,105 @@ fn decap_and_seal_internal(
...
+    let metas = KEY_REGISTRY.list_kem_keys(offset, entries.len() + 1);
+    let mut has_more = false;
+    let limit = entries.len();
+
+    let process_metas = if metas.len() > limit {
+        has_more = true;
+        &metas[..limit]
+    } else {
+        &metas[..]
+    };
```
bluegate010:
```suggestion
    let (process_metas, has_more) = if metas.len() > limit {
        (&metas[..limit], true)
    } else {
        (&metas[..], false)
    };
```

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -196,6 +196,105 @@ fn decap_and_seal_internal(
...
+            kem_public_key,
+            ..
+        } = &meta.spec
+        {
+            (algo.encode_to_vec(), kem_public_key.clone())
+        } else {
+            return None;
+        };
+
+        if kem_public_key.as_bytes().len() > 2048 || algo_bytes.len() > 128 {
```
bluegate010: How was 2048 chosen? Let's make it a named constant?

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -196,6 +196,105 @@ fn decap_and_seal_internal(
...
+    pub uuid: [u8; 16],
+    pub algorithm: [u8; 128],
+    pub algorithm_len: usize,
+    pub kem_pub_key: [u8; 2048],
+    pub kem_pub_key_len: usize,
+    pub remaining_lifespan_secs: u64,
+}
+
+fn enumerate_kem_keys_internal(entries: &mut [KpsKeyInfo], offset: usize) -> (usize, bool) {
+    let metas = KEY_REGISTRY.list_kem_keys(offset, entries.len() + 1);
```
bluegate010: Why are we adding one to entries.len()?

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -196,6 +196,105 @@ fn decap_and_seal_internal(
...
+        entry.kem_pub_key[..kem_public_key.as_bytes().len()]
+            .copy_from_slice(kem_public_key.as_bytes());
+        entry.kem_pub_key_len = kem_public_key.as_bytes().len();
+
+        entry.remaining_lifespan_secs = remaining;
+
+        count += 1;
+    }
+
+    (count, has_more)
```
bluegate010: > count

Nit: this could just be `valid_metas.len()` right? (cuts down on stateful variables)

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -196,6 +196,105 @@ fn decap_and_seal_internal(
...
+    let limit = entries.len();
+
+    let process_metas = if metas.len() > limit {
+        has_more = true;
+        &metas[..limit]
+    } else {
+        &metas[..]
+    };
+
+    let now = Instant::now();
```
bluegate010: Nit: suggest moving this to closer where it's used.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -196,6 +196,105 @@ fn decap_and_seal_internal(
...
+        }
+    });
+
+    let mut count = 0;
+    for (entry, (meta, algo_bytes, kem_public_key)) in entries.iter_mut().zip(valid_metas) {
+        let remaining = meta.delete_after.saturating_duration_since(now).as_secs();
+
+        entry.uuid.copy_from_slice(meta.id.as_bytes());
+
+        entry.algorithm = [0u8; 128];
```
bluegate010: Instead of zero-initializing these particular fields, could this just do `entry = KpsKeyInfo::default()` up front?

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -196,6 +196,105 @@ fn decap_and_seal_internal(
...
+        has_more = true;
+        &metas[..limit]
+    } else {
+        &metas[..]
+    };
+
+    let now = Instant::now();
+
+    let valid_metas = process_metas.iter().filter_map(|meta| {
+        let (algo_bytes, kem_public_key) = if let KeySpec::KemWithBindingPub {
```
bluegate010: Here we're filtering out Binding key entries. Which is correct if we're trying to just surface KEM keys, but it seems like we should surface all kinds of keys. (Seems like we could just s/kem_pub_key/pub_key/ in KpsKeyInfo.)

keymanager/key_protection_service/key_custody_core/src/lib.rs
bluegate010: I feel like we should not skip these keys, but fail loudly if this occurs. If panicking in production builds is not desired, then at least an error code.

atulpatildbz: makes sense. added logic to fail in this case. [here](https://github.com/google/go-tpm-tools/pull/665/changes/ffaa09e687cf752777ca7a7534142e1751d30071)

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: 2048 was chosen as a generous maximum buffer limit for vector allocation over the FFI boundary. which is enough for current `X25519` algorithm and has enough headroom for larger keys in the future.

i've added constants `MAX_KEM_PUBLIC_KEY_LEN` and `MAX_ALGORITHM_LEN` [here](https://github.com/google/go-tpm-tools/pull/665/changes/b4b5cfcd08b9aa1bd2390d4863647c44599c9c1b)

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: [done](https://github.com/google/go-tpm-tools/pull/665/changes/507cf70c7544203068056d246aa86b4c814b3b2f)

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: values of `valid_metas` are moved due to `into_iter`, so we can't use `valid_metas.len()` at the end of the function.
But i've removed the `count += 1` inside the loop and just stored the vector length in the count variable once before the loop begins. ([here](https://github.com/google/go-tpm-tools/pull/665/changes/507cf70c7544203068056d246aa86b4c814b3b2f))

Thanks for pointing that out!

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: [done](https://github.com/google/go-tpm-tools/pull/665/changes/fb6b4949ce60f815c1c021aabe4b9362f930b74c)

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: I thought currently we are only supposed to return KEM keys from enumerate API? let me know if my understanding is correct.

also cc: @NilanjanDaw

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: for making it easier to determine `has_more`
Basically looking at 1 extra item to determine if there are more entries in the next page.

Alternative was to get the `total_count` of KEM keys and using that to calculate has_more
```
has_more = (offset + metas.len()) < total_count
```
that would require changing `fn list_kem_keys` to return total_count. 

Now that i think about it. the alternative of `total_count` might be simpler to read. I'll make this change shortly.

keymanager/key_protection_service/key_custody_core/src/lib.rs
bluegate010: Will defer to Nilanjan. I see a few API surfaces at play:

- The Rust FFI exposed from within KPS
- The Rust FFI exposed from within WSD
- The RPC service exposed from KPS to WSD
- The RPC service exposed from WSD to the workload container

Seems to me that all APIs should enumerate all keys, except for the RPC service exposed from WSD to the workload container, which should not enumerate binding keys. (Binding keys only need to be exposed as part of attestation, i.e. if I attest to a KEM key held in KPS then I also need to attest to the associated binding key held in WSD.)

atulpatildbz: Discussed offline. Went with the approach you suggested.

1. KPS FFI -> returns Bounded kem key
2. WSD FFI -> return Binding keys

And i'll handle the logic to filter the KEMs from the BoundKEM struct in the RPC in separate PR.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: I think we should move this much further down (or at least within the loop below)

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: Changed the logic to use total count for has_more check

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -229,6 +229,115 @@ pub unsafe extern "C" fn key_manager_open(
...
+    let limit = entries.len();
+
+    let (process_metas, has_more) = if metas.len() > limit {
+        (&metas[..limit], true)
+    } else {
+        (&metas[..], false)
+    };
+
+    let count = process_metas.len();
+    let now = std::time::Instant::now();
```
NilanjanDaw: Same here, lets move this closer to the point of use.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -196,6 +196,123 @@ fn decap_and_seal_internal(
...
+    Ok((count, has_more))
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn key_manager_enumerate_kem_keys(
+    out_entries: *mut KpsKeyInfo,
+    max_entries: usize,
+    offset: usize,
+    out_has_more: Option<&mut bool>,
+) -> i32 {
```
NilanjanDaw, line 297: Should we wrap the function in a `catch_unwind` to prevent propagation of Rust panics?

keymanager/workload_service/key_custody_core/cbindgen.toml
```diff
@@ -15,7 +15,7 @@ parse_deps = false
 clean = true
 
 [export]
```
NilanjanDaw, line 17: Lets add s`tyle = "type"` to the WS `cbindgen.toml` so that we have a consistent C interface across for both the `KpsKeyInfo` and `WsKeyInfo` (allowing callers to use `WsKeyInfo` instead of `struct WsKeyInfo`).

atulpatildbz: done

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: done

keymanager/workload_service/key_custody_core/src/lib.rs
atulpatildbz: done

PR #666, opened 2026-02-15
[keymanager]: implement GetCapabilities API

atulpatildbz:
Implement Get Capabilities API

This PR implements the `GET /v1/capabilities` endpoint in the Workload Service.
It returns the supported KEM algorithms and key protection mechanisms, allowing clients to discover available features.

**Dependencies:**
- Dependent on PR #652 

Tested response:
```json
{
  "supported_algorithms": [
    {
      "algorithm": {
        "type": "kem",
        "params": {
          "kem_id": "DHKEM_X25519_HKDF_SHA256"
        }
      }
    }
  ]
}
```

### DIFF
--- a/keymanager/workload_service/server.go
+++ b/keymanager/workload_service/server.go
@@ -202,8 +202,7 @@ func (s *Server) handleGenerateKem(w http.ResponseWriter, r *http.Request) {
 	resp := GenerateKemResponse{
 		KeyHandle: KeyHandle{Handle: kemUUID.String()},
 	}
-	w.Header().Set("Content-Type", "application/json")
-	json.NewEncoder(w).Encode(resp)
+	writeJSON(w, resp, http.StatusOK)
 }
 
 func (s *Server) handleGetCapabilities(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


keymanager/workload_service/server.go
```diff
@@ -184,12 +206,36 @@ func (s *Server) handleGenerateKem(w http.ResponseWriter, r *http.Request) {
...
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(resp)
+}
+
 // writeError writes a JSON error response.
 func writeError(w http.ResponseWriter, message string, code int) {
 	w.Header().Set("Content-Type", "application/json")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 	w.WriteHeader(code)
 	json.NewEncoder(w).Encode(map[string]string{"error": message})
```
NilanjanDaw: nit: since these are expected to be used for every API response, should we separate this into a separate function of its own? Something like `createResponse` which will automatically add the necessary headers and also JSON encode the response.

keymanager/workload_service/server.go
```diff
@@ -184,12 +206,36 @@ func (s *Server) handleGenerateKem(w http.ResponseWriter, r *http.Request) {
...
+				},
+			},
+		})
+	}
+
+	resp := GetCapabilitiesResponse{
+		SupportedAlgorithms: supportedAlgos,
+	}
+
+	w.Header().Set("Content-Type", "application/json")
```
NilanjanDaw: Do we need to add `w.Header().Set("X-Content-Type-Options", "nosniff")` here?

keymanager/workload_service/proto_enums.go
```diff
@@ -62,12 +63,12 @@ const (
 
 var (
 	keyProtectionMechanismToString = map[KeyProtectionMechanism]string{
-		KeyProtectionMechanismDefault: "KEY_PROTECTION_DEFAULT",
```
NilanjanDaw, line 65: I am assuming you will add the `KEY_PROTECTION_EMULATED` mode later.

NilanjanDaw (approved):
Thanks for the PR! Some minor nits.

keymanager/workload_service/proto_enums.go
atulpatildbz: planning to do is in another PR: https://github.com/google/go-tpm-tools/pull/677

keymanager/workload_service/server.go
atulpatildbz: [added](https://github.com/google/go-tpm-tools/pull/666/changes#diff-d977b51b4c6a39e550deb8e3aa85a1f379c1d189c96da32490ea61e2027a0d48R238)

keymanager/workload_service/server.go
atulpatildbz: Done. Refactored the JSON response generation into a shared  `writeJSON` helper function. Updated the key generation and capabilities endpoint handlers, as well as the error writer, to utilize this new function

let me know if you had the same idea, or something else

PR #672, opened 2026-02-20
[launcher] add flag to disable GCA refresh

jkl73:
The new flag allow user to disable the default initial and hourly GCA attestation token refresh.

### DIFF
--- a/launcher/spec/launch_spec.go
+++ b/launcher/spec/launch_spec.go
@@ -94,6 +94,7 @@ const (
 	cgroupNS                   = "tee-cgroup-ns"
 	gcaServiceEnv              = "gca-service-env"
 	installGpuDriver           = "tee-install-gpu-driver"
+	disableGcaRefreshKey       = "tee-disable-gca-refresh"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


launcher/spec/launch_spec.go
```diff
@@ -306,6 +307,13 @@ func (s *LaunchSpec) UnmarshalJSON(b []byte) error {
 		}
 	}
 
+	if val, ok := unmarshaledMap[disableGcaRefreshKey]; ok && val != "" {
+		var err error
+		if s.DisableGcaRefresh, err = strconv.ParseBool(val); err != nil {
+			return fmt.Errorf("invalid value for %v (not a boolean): %w", disableGcaRefreshKey, err)
+		}
+	}
```
yawangwang, line 315: Could you add unit tests?

yawangwang (approved):

jessieqliu (approved):
Should we remove the mounting of the token file if there's no GCA refresh (meaning nothing is written to the file)?

launcher/spec/launch_spec.go
jkl73: added

PR #676, opened 2026-02-24
Integrate go-nvtrust to collect GPU attestation

yawangwang:
Integrate the OSS library [go-nvtrust ](https://github.com/confidentsecurity/go-nvtrust/tree/main/pkg) into tdx attester to collect confidential GPU attestation.


launcher/agent/agent.go
```diff
@@ -404,6 +409,10 @@ func (t *tpmAttestRoot) ComputeNonce(challenge []byte, extraData []byte) []byte
 	return finalNonce[:]
 }
 
+func (t *tpmAttestRoot) GetDeviceAttestation(_ []byte) (*verifier.DeviceAttestation, error) {
+	return nil, nil
```
alexmwu: Why would we exclude this if we have it for TDX?

launcher/agent/agent.go
```diff
@@ -459,6 +474,21 @@ func (t *tdxAttestRoot) ComputeNonce(challenge []byte, extraData []byte) []byte
 	return finalNonce[:]
 }
 
+func (t *tdxAttestRoot) GetDeviceAttestation(nonce []byte) (*verifier.DeviceAttestation, error) {
```
alexmwu: I wonder if we should have a separate device RoT. It seems strange to have this be a method under the TDX attestation root.

launcher/agent/agent.go
```diff
@@ -459,6 +474,21 @@ func (t *tdxAttestRoot) ComputeNonce(challenge []byte, extraData []byte) []byte
...
+	deviceAttestation := &verifier.DeviceAttestation{}
+	// Try get GPU device attestation if gpu device is detected
+	gpuType, _ := deviceinfo.GetGPUTypeInfo()
```
alexmwu: The launcher could instantiate detected devices and then pass only the supported ones into the attestation agent.

launcher/internal/gpu/attestation.go
```diff
@@ -0,0 +1,116 @@
...
+	// MPT indicates Nvidia's multi GPU secure passthrough attestation
+	MPT
+)
+
+// Stub this func for testing purpose.
+var getGpuTypeInfo = deviceinfo.GetGPUTypeInfo
+
+// CollectAttestationEvidence assumes CC GPU devices are in place w/ driver support
+// and will try to collect raw attestation evidence and convert it to known data models.
+func CollectAttestationEvidence(handler gpu.NvmlHandler, nonce []byte) (*models.NvidiaAttestation, error) {
```
alexmwu: This will need a bit of refactoring to use the protos, but that can be done in a follow-up CL

launcher/agent/agent.go
alexmwu: The logic in my mind goes like:
1. Launcher already detects supported devices due to drivers.
2. Launcher installs drivers but does not put GPU into ready state.
3.  Launcher calls MeasureEvent with the GPU's DeviceAttestation proto. 
4. Launcher puts GPUs into ready state.
5. Collect attestation evidence for runtime (tbd if needed) in the Attest call

launcher/cloudbuild.yaml
```diff
@@ -37,7 +37,7 @@ steps:
     - -c
     - |
       cd launcher/launcher
-      CGO_ENABLED=0 go build -o ../image/launcher -ldflags="-X 'main.BuildCommit=${SHORT_SHA}'"
+      go build -o ../image/launcher -ldflags="-extldflags=-Wl,-z,lazy -X 'main.BuildCommit=${_SHORT_SHA}'"
```
jkl73: are you sure it is ${_SHORT_SHA} instead of ${SHORT_SHA}

launcher/internal/gpu/attestation.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, fmt.Errorf("unsupported GPU attestation")
+	}
+}
+
+// determineAttesationType auto-detects the GPU attestation type.
+// The current implementations "guess" the attestation type.
+// Further improvement should be made to parse GPU attesation report to get the actual attestation type.
+func determineAttestationType(gpuInfos []models.GPUInfo) attestationType {
+	gpuType, _ := getGpuTypeInfo()
+	if gpuType != deviceinfo.H100 && gpuType != deviceinfo.B200 {
```
jkl73, line 121: is it possible to just use deviceInfo.Arch() to determine the behavior, instead of using cos's deviceinfo which is an additional dependency.

launcher/internal/gpu/attestation.go
```diff
@@ -0,0 +1,116 @@
...
+		if ret != nvml.SUCCESS {
+			return nil, fmt.Errorf("failed to get GPU VBIOS version: %v", nvml.ErrorString(ret))
+		}
+
+		driverVersion, ret := handler.SystemGetDriverVersion()
+		if ret != nvml.SUCCESS {
+			return nil, fmt.Errorf("failed to get GPU driver version: %v", nvml.ErrorString(ret))
+		}
+
+		base64PEM, err := deviceInfo.Certificate().EncodeBase64()
```
jkl73, line 77: are sure it's a PEM format? (start with header -----BEGIN)

launcher/agent/agent.go
yawangwang: Removed

launcher/agent/agent.go
yawangwang: That's also my understanding. Updated the PR to have launcher pass only the supported ones into the attestation agent.

launcher/agent/agent.go
yawangwang: Done.

launcher/internal/gpu/attestation.go
yawangwang: Agreed

launcher/cloudbuild.yaml
yawangwang: Good catch, done.

launcher/internal/gpu/attestation.go
yawangwang: That's also my initial thoughts of NOT depending on cos's deviceinfo. 

However, I'm trying to avoid hardcoding those GPU arch because the source code returns the hardcoded strings not an enum type https://github.com/confidentsecurity/go-nvtrust/blob/33efe3bbb4ffa918cdfadb30193b12fba624d979/pkg/gonvtrust/gpu/gpu_admin.go#L38. My concern is if the upstream code accidently changes, it will break our GPU attestation collection flow.

 Note that cos's deviceinfo is not actually an additional dependency - attestation.go shares the same package with driverinstaller which also depends on cos's deviceinfo.

launcher/internal/gpu/attestation.go
yawangwang: From source code i think so https://github.com/confidentsecurity/go-nvtrust/blob/33efe3bbb4ffa918cdfadb30193b12fba624d979/pkg/gonvtrust/certs/cert_chain.go#L79

launcher/internal/gpu/attestation.go
jkl73: sgtm, I think it might be easier to do unit test without the package. But I think it's okay for now, we can refactor later

launcher/agent/agent.go
```diff
@@ -71,6 +71,14 @@ type attestRoot interface {
 	Attest(nonce []byte) (any, error)
 	// ComputeNonce hashes the challenge and extraData using the algorithm preferred by the attestation root.
 	ComputeNonce(challenge []byte, extraData []byte) []byte
+	// AddDeviceRoTS adds detected device RoTs(root of trust).
+	AddDeviceRoTS([]DeviceRoT)
```
alexmwu: AddDeviceRoTs or AddDeviceROTs

verifier/client.go
```diff
@@ -58,6 +58,11 @@ type TDCCELAttestation struct {
 	// still needs following two for GCE info
 	AkCert            []byte
 	IntermediateCerts [][]byte
+	DeviceAttestation *DeviceAttestation
+}
+
+type DeviceAttestation struct {
```
alexmwu: Why a DeviceAttestation vs just using the NvidiaAttestation directly? For VerifyCS, we expose the NvidiaAttestation without wrapping it

launcher/internal/gpu/attestation.go
```diff
@@ -0,0 +1,128 @@
...
+	SPT
+	// MPT indicates Nvidia's multi GPU secure passthrough attestation
+	MPT
+)
+
+// Stub this func for testing purpose.
+var getGpuTypeInfo = deviceinfo.GetGPUTypeInfo
+
+// Attester is responsible for collecting GPU attestation.
+type Attester struct{}
```
alexmwu: NvidiaAttester

yawangwang: Done

verifier/client.go
yawangwang: My initial thought is to create an abstraction layer to host any device attestation (gpu for now, but tpu may come in the future). But we can also keep parity with VerifyCS request, removed.

launcher/agent/agent.go
yawangwang: AddDeviceRoTs was the initial naming but triggered linter check failures. I accepted the `AddDeviceRoTS` as what linter suggested but `AddDeviceROTs` makes more sense here. Updated

PR #678, opened 2026-02-25
[launcher][KeyManager] Integrate KeyManager into the Confidential Spa…

NilanjanDaw:
…ce image build

This commit integrates the KeyManager Workload Service into the Confidential Space launcher and ensures it is properly built and included in the resulting images.

Key changes:
- Corrected the argument order in `NewServer` calls within `keymanager/workload_service/server_test.go` and `integration_test.go` to match the expected interface definitions.
- Updated `launcher/cloudbuild.yaml` to install Rust, CMake, and the C++ toolchain in the `golang:1.24` build environment, enabling the `CGO_ENABLED=1` build of the launcher to successfully link the KeyManager Rust libraries via FFI.
    - **Note.** We need to downgrade the current `golang:1.24` to `golang:1.24-bullseye` to fix `glibc` incompatibility with the `cos-125-lts` image.
- Hooked the KeyManager Server into `launcher/container_runner.go`, configured to start listening on `kmaserver.sock` when the `EnableKeyManager` experiment is active.
- Added the `EnableKeyManager` flag to the `Experiments` struct and updated relevant tests and configurations to support the new feature toggle.

Tested:
- All `cloudbuild` tests are passing.
- Instantiated a CS VM with the `confidential-space-vg-debug-0-presubmit` and queried the `capabilities` API which returned the supported algorithms list.
```bash
curl --unix-socket /tmp/container_launcher/kmaserver.sock http://localhost/v1/capabilities
{"supported_algorithms":[{"algorithm":{"type":"kem","params":{"kem_id":"DHKEM_X25519_HKDF_SHA256"}}}]}
```

### DIFF
--- a/launcher/cloudbuild.yaml
+++ b/launcher/cloudbuild.yaml
@@ -46,6 +46,10 @@ steps:
       export PATH="/opt/cmake-3.28.3-linux-x86_64/bin:$$PATH"
       rm cmake.tar.gz
 
+      # Install bindgen-cli (needed by BoringSSL's CMake build)
+      cargo install bindgen-cli
+      export PATH="$$HOME/.cargo/bin:$$PATH"
+
       # Build the Rust keymanager libraries
       cd keymanager
       cargo build --release


.goreleaser.yaml
```diff
@@ -7,7 +7,7 @@ builds:
       - amd64
       - arm64
     env:
-      - CGO_ENABLED=0
+      - CGO_ENABLED=1
```
jkl73: it's probably not needed here, since this is for the cmd tool

go.work.sum
atulpatildbz, line 1: this file can be added to .gitignore

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_stub.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+
+	algorithms "github.com/google/go-tpm-tools/keymanager/km_common/proto"
+)
+
+// GenerateKEMKeypair is a stub for architectures where the Rust library is not supported.
+func GenerateKEMKeypair(algo *algorithms.HpkeAlgorithm, bindingPubKey []byte, lifespanSecs uint64) (uuid.UUID, []byte, error) {
```
atulpatildbz: parameters are named but they are never used. we can replace them with black identifiers

```suggestion
func GenerateKEMKeypair(_ *algorithms.HpkeAlgorithm, _ []byte, _ uint64) (uuid.UUID, []byte, error) {
```

launcher/cloudbuild.yaml
```diff
@@ -30,14 +30,36 @@ steps:
     echo "base image:" ${base_image} "project:" ${BASE_IMAGE_PROJECT} 
     echo ${base_image} > /workspace/base_image.txt
 
-- name: golang:1.24
+- name: golang:1.24-bullseye
```
atulpatildbz: was this done to build against older version of `glibc`?

launcher/cloudbuild.yaml
```diff
@@ -30,14 +30,36 @@ steps:
...
   entrypoint: /bin/bash
   args:
     - -c
     - |
+      set -exuo pipefail
+      # Install required build tools (excluding the old cmake) and Rust toolchain
+      apt-get update && apt-get install -y clang pkg-config libssl-dev build-essential curl
+      
+      # Install a newer CMake (need 3.22+, Bullseye has 3.18)
+      curl -sSL https://github.com/Kitware/CMake/releases/download/v3.28.3/cmake-3.28.3-linux-x86_64.sh -o cmake-install.sh
```
atulpatildbz: Running curl or downloading a .sh file and executing it directly is generally discouraged in CI/CD

check if we can add the official Kitware APT repository and use `apt-get install cmake`

This can be taken as an improvement in a follow up PR.

launcher/cloudbuild.yaml
```diff
@@ -30,14 +30,36 @@ steps:
...
+      
+      # Build the Rust keymanager libraries
+      cd keymanager
+      cargo build --release
+      cd ..
+      
+      # Build the launcher with CGO enabled
       cd launcher/launcher
-      CGO_ENABLED=0 go build -o ../image/launcher -ldflags="-X 'main.BuildCommit=${SHORT_SHA}'"
+      export CGO_LDFLAGS="-L/workspace/keymanager/target/release -L/workspace/keymanager/target/debug"
```
atulpatildbz: for CI pipeline, debug is unnecessary right?

```suggestion
      export CGO_LDFLAGS="-L/workspace/keymanager/target/release"
```

launcher/cloudbuild.yaml
jkl73: Rust building can be a separate step with a proper rust building image, doesn't need to reuse golang building step here

.goreleaser.yaml
```diff
@@ -8,6 +8,11 @@ builds:
       - arm64
     env:
       - CGO_ENABLED=0
+    overrides:
```
jkl73: probably no need to change here, since it's for the gotpm cmd package

jkl73 (approved):
launcher integration lgtm

launcher/cloudbuild.yaml
jkl73: took a look at the built binary, might be libc, might be gcc:

	linux-vdso.so.1 (0x00007fd74e2f2000)
	libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007fd74e2bc000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fd74e2b7000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fd74e2b2000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fd74e281000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fd74e08b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd74e2f4000)

go.work.sum
NilanjanDaw: Hmm this has been committed for a long time now. Not sure if we should change it.

launcher/cloudbuild.yaml
NilanjanDaw: Yeah this was done to handle a compatibility issue with `glibc`. The LauncherCompiler was using `golang:1.24` which uses `GLIBC_2.39`. This is a newer version than what is available with the `cos-125-lts` image. I downgraded the `golang:1.24` to `golang:1.24-bullseye` which is the standard debian distribution. This resolved the `glibc` version available with `cos-125-lts`.

go.work.sum
atulpatildbz: Okay i thought you created and pushed it. We can remove this later.

launcher/cloudbuild.yaml
atulpatildbz: Okay. which means we can revert this once we move to COS 129.
Is it possible to add a short comment here?

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_stub.go
NilanjanDaw: Thanks for edit. Added.

launcher/cloudbuild.yaml
NilanjanDaw: Ack. Added a comment to revert back to golang:1.24 for COS 129.

launcher/cloudbuild.yaml
NilanjanDaw: The problem is `debian 11` bullseye currently supports cmake `3.18` while `bssl-crypto` needs `3.22+` which is not available with available with `apt-get`. I have updated the script to custom install the cmake using tar.gz instead of runing the custom .sh file.

launcher/cloudbuild.yaml
NilanjanDaw: Ack removed debug.

.goreleaser.yaml
NilanjanDaw: Makes sense removed.

launcher/cloudbuild.yaml
NilanjanDaw: Ack makes sense, divided into the build into two steps.

.goreleaser.yaml
NilanjanDaw: Ack thanks. Removed.

atulpatildbz (approved):

PR #671, opened 2026-02-20
[KeyManager] Add KeyClaims proto and generation script.

atulpatildbz:
adds the `KeyClaims` proto definition in `keymanager/km_common/proto/` which includes `VmAttestation`, `KeyAttestation`, and related messages for key protection and attestation.

It also adds `proto/gen_keymanager.sh` to generate Go code for keymanager protos.

PR #681, opened 2026-02-27
[KeyManager] Add key claims retrieval to KPS and WSD

NilanjanDaw:
This commit introduces new functionality for retrieving key claims and provides
more comprehensive key management operations across both the Key Protection
Service (KPS) and Workload Service Daemon (WSD).

Key changes include:
- **KPS (key_custody_core):**
  - Added FFI definitions and implementations for `key_manager_get_kem_key`.
  - Implemented `GetKemKey` in `kps_key_custody_core_cgo.go` for Go-FFI bridging.
  - Added integration tests for KEM key retrieval.
- **WSD (workload_service):**
  - Added FFI definitions and implementations for `key_manager_get_binding_key`.
  - Implemented `GetBindingKey` in `ws_key_custody_core_cgo.go` for Go-FFI bridging.
  - Added integration tests for binding key retrieval.
  - Introduced key claims processing mechanism:
    - Added `KeyClaimsProvider` interface.
    - `Server` now includes a `claimsChan` and a processClaims` goroutine.
    - Implemented `GetBindingKeyClaims` and `GetKemKeyClaims` for generating protobuf claims.
  - Updated server constructors and integration tests to support the new claims and interfaces.
- Added comprehensive integration tests and unit tests for the new key management
  and claims functionalities in both KPS and WSD.

This PR depends on #671. Please review it before this one.

### DIFF
--- a/keymanager/workload_service/server.go
+++ b/keymanager/workload_service/server.go
@@ -252,6 +252,12 @@ type Server struct {
 	// todo: add logging mechanism here
 }
 
+var (
+	// ClaimsResponseTimeout is the maximum time to wait for the caller to receive
+	// the result of a GetKeyClaims request before timing out.
+	ClaimsResponseTimeout = 5 * time.Second
+)
+
 // New creates a new WSD Server listening on the given unix socket path.
 func New(_ context.Context, socketPath string) (*Server, error) {
 	return NewServer(&keyProtectionService{}, &workloadService{}, socketPath)
@@ -513,8 +519,8 @@ func (s *Server) handleDestroy(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusNoContent)
 }
 
-// GetBindingKeyClaims returns the claims for a binding key identified by its KEM UUID.
-func (s *Server) GetBindingKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error) {
+// handleGetBindingKeyClaims returns the claims for a binding key identified by its KEM UUID.
+func (s *Server) handleGetBindingKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error) {
 	// Key ID Lookup. The orchestration layer will look-up the key_handle
 	// in its ActiveKeyRegistry to find the Binding Key ID.
 	bindingID, ok := s.LookupBindingUUID(id)
@@ -542,8 +548,8 @@ func (s *Server) GetBindingKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error
 	return claims, nil
 }
 
-// GetKEMKeyClaims returns the claims for a KEM key identified by its UUID.
-func (s *Server) GetKEMKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error) {
+// handleGetKEMKeyClaims returns the claims for a KEM key identified by its UUID.
+func (s *Server) handleGetKEMKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error) {
 	// Key Metadata Lookup.
 	kemPubKey, bindingPubKey, algo, remainingLifespanSecs, err := s.keyProtectionService.GetKEMKey(id)
 	if err != nil {
@@ -572,37 +578,45 @@ func (s *Server) GetKEMKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error) {
 	return claims, nil
 }
 
+// processClaims is a background worker that processes key claims requests from claimsChan.
 func (s *Server) processClaims() {
 	for call := range s.claimsChan {
-		req := call.Request
-		keyHandle := req.GetKeyHandle().GetHandle()
-		keyType := req.GetKeyType()
+		result := s.handleGetClaims(call.Request)
 
-		id, err := uuid.Parse(keyHandle)
-		if err != nil {
-			call.RespChan <- &ClaimsResult{Err: fmt.Errorf("failed to retrieve key claims: %w", err)}
-			continue
+		select {
+		case call.RespChan <- result:
+		case <-time.After(ClaimsResponseTimeout):
+			log.Printf("processClaims: timed out sending response for key %s", call.Request.GetKeyHandle().GetHandle())
 		}
-		var claims *keymanager.KeyClaims
-		switch keyType {
-		case keymanager.KeyType_KEY_TYPE_VM_PROTECTION_BINDING:
-			claims, err = s.GetBindingKeyClaims(id)
-			if err != nil {
-				call.RespChan <- &ClaimsResult{Err: fmt.Errorf("failed to retrieve binding key claims: %w", err)}
-				continue
-			}
-
-		case keymanager.KeyType_KEY_TYPE_VM_PROTECTION_KEY:
-			claims, err = s.GetKEMKeyClaims(id)
-			if err != nil {
-				call.RespChan <- &ClaimsResult{Err: fmt.Errorf("failed to retrieve VM protection key claims: %w", err)}
-				continue
-			}
-		default:
-			call.RespChan <- &ClaimsResult{Err: fmt.Errorf("unsupported key type: %v", keyType)}
-			continue
-		}
-
-		call.RespChan <- &ClaimsResult{Reply: claims}
 	}
+}
+
+// handleGetClaims processes a single GetKeyClaimsRequest and returns the result.
+func (s *Server) handleGetClaims(req *keymanager.GetKeyClaimsRequest) *ClaimsResult {
+	keyHandle := req.GetKeyHandle().GetHandle()
+	keyType := req.GetKeyType()
+
+	id, err := uuid.Parse(keyHandle)
+	if err != nil {
+		return &ClaimsResult{Err: fmt.Errorf("failed to retrieve key claims: %w", err)}
+	}
+
+	var claims *keymanager.KeyClaims
+	switch keyType {
+	case keymanager.KeyType_KEY_TYPE_VM_PROTECTION_BINDING:
+		claims, err = s.handleGetBindingKeyClaims(id)
+		if err != nil {
+			return &ClaimsResult{Err: fmt.Errorf("failed to retrieve binding key claims: %w", err)}
+		}
+
+	case keymanager.KeyType_KEY_TYPE_VM_PROTECTION_KEY:
+		claims, err = s.handleGetKEMKeyClaims(id)
+		if err != nil {
+			return &ClaimsResult{Err: fmt.Errorf("failed to retrieve VM protection key claims: %w", err)}
+		}
+	default:
+		return &ClaimsResult{Err: fmt.Errorf("unsupported key type: %v", keyType)}
+	}
+
+	return &ClaimsResult{Reply: claims}
 }
--- a/keymanager/workload_service/server_test.go
+++ b/keymanager/workload_service/server_test.go
@@ -1024,3 +1024,50 @@ func TestHandleDecapsUnsupportedAlgorithm(t *testing.T) {
 		t.Fatalf("expected status 400, got %d", w.Code)
 	}
 }
+
+func TestProcessClaimsTimeout(t *testing.T) {
+	oldTimeout := ClaimsResponseTimeout
+	ClaimsResponseTimeout = 10 * time.Millisecond
+	defer func() { ClaimsResponseTimeout = oldTimeout }()
+
+	srv := newTestServer(t, &mockKeyProtectionService{}, &mockWorkloadService{})
+	// processClaims is already started in newTestServer -> NewServer -> New
+
+	respChan1 := make(chan *ClaimsResult) // Unbuffered
+	req1 := &keymanager.GetKeyClaimsRequest{
+		KeyHandle: &keymanager.KeyHandle{Handle: uuid.New().String()},
+		KeyType:   keymanager.KeyType_KEY_TYPE_VM_PROTECTION_BINDING,
+	}
+
+	// 1. Send first request and DO NOT read from it.
+	// This should timeout in 10ms.
+	srv.claimsChan <- &ClaimsCall{Request: req1, RespChan: respChan1}
+
+	// 2. Send second request and read from it.
+	// We need a valid UUID in the map for this to succeed easily.
+	kemUUID := uuid.New()
+	bindingUUID := uuid.New()
+	srv.mu.Lock()
+	srv.kemToBindingMap[kemUUID] = bindingUUID
+	srv.mu.Unlock()
+
+	respChan2 := make(chan *ClaimsResult, 1)
+	req2 := &keymanager.GetKeyClaimsRequest{
+		KeyHandle: &keymanager.KeyHandle{Handle: kemUUID.String()},
+		KeyType:   keymanager.KeyType_KEY_TYPE_VM_PROTECTION_BINDING,
+	}
+
+	// Give it a bit of time for the first one to timeout
+	time.Sleep(20 * time.Millisecond)
+
+	srv.claimsChan <- &ClaimsCall{Request: req2, RespChan: respChan2}
+
+	select {
+	case res := <-respChan2:
+		if res.Err != nil {
+			t.Errorf("expected no error for second request, got: %v", res.Err)
+		}
+	case <-time.After(1 * time.Second):
+		t.Fatal("timed out waiting for second request response - background worker might be blocked!")
+	}
+}


keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -400,6 +400,105 @@ pub unsafe extern "C" fn key_manager_decap_and_seal(
...
+            || out_binding_pubkey_len == 0
+            || out_delete_after.is_null()
+        {
+            return -1;
+        }
+
+        // Convert to Safe Types
+        let uuid_slice = unsafe { std::slice::from_raw_parts(uuid_bytes, 16) };
+        let out_kem_pubkey_slice =
+            unsafe { std::slice::from_raw_parts_mut(out_kem_pubkey, out_kem_pubkey_len) };
```
atulpatildbz, line 474: we're checking the `out_kem_pubkey_len` value much later after we're already using it to create a slice.

from_raw_parts_mut asserts to the compiler that valid `len` bytes exists. And this could result in undefined behaviour.

check the length first and create the slice only after we know the length is correct.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -400,6 +400,105 @@ pub unsafe extern "C" fn key_manager_decap_and_seal(
...
+
+        let uuid = match Uuid::from_slice(uuid_slice) {
+            Ok(u) => u,
+            Err(_) => return -1,
+        };
+
+        // Call Safe Internal Function
+        match get_kem_key_internal(uuid) {
+            Ok((kem_pubkey, binding_pubkey, delete_after)) => {
+                if out_kem_pubkey_len != kem_pubkey.as_bytes().len()
```
atulpatildbz: we've followed `!=` size enforcement at many places which is fine for now since we own the cgo layer too.

But would `<` be a more generalized check? that way caller can pass larger buffer too if they don't know the exact size.

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -992,4 +1091,78 @@ mod tests {
...
+                uuid_bytes.as_ptr(),
+                kem_pubkey_bytes.as_mut_ptr(),
+                kem_pubkey_bytes.len(),
+                binding_pubkey_bytes.as_mut_ptr(),
+                binding_pubkey_bytes.len(),
+                &mut delete_after,
+            )
+        };
+        assert_eq!(result, -1);
+    }
```
atulpatildbz, line 1198: can we add a test `test_get_kem_key_invalid_buffer_len` that passes `retrieved_kem_pubkey_bytes.len() + 1` or  - 1 to ensure the C-API correctly rejects it with a -2 status code

keymanager/key_protection_service/key_custody_core/integration_test.go
```diff
@@ -61,3 +61,49 @@ func TestIntegrationGenerateKEMKeypairUniqueness(t *testing.T) {
...
+	retrievedKemPK, retrievedBindingPK, deleteAfter, err := GetKemKey(id)
+	if err != nil {
+		t.Fatalf("GetKemKey failed: %v", err)
+	}
+
+	if len(retrievedKemPK) != len(pubKey) {
+		t.Fatalf("expected KEM pubkey length %d, got %d", len(pubKey), len(retrievedKemPK))
+	}
+	for i := range pubKey {
+		if pubKey[i] != retrievedKemPK[i] {
```
atulpatildbz: instead of iterating and comparing, we can use `bytes.Equal`

```go
import "bytes"

// ...

if !bytes.Equal(pubKey, retrievedKemPK) {
	t.Fatalf("KEM pubkey mismatch: expected %x, got %x", pubKey, retrievedKemPK)
}

if !bytes.Equal(bindingPK, retrievedBindingPK) {
	t.Fatalf("binding pubkey mismatch: expected %x, got %x", bindingPK, retrievedBindingPK)
}

```

keymanager/key_protection_service/key_custody_core/integration_test.go
```diff
@@ -61,3 +61,49 @@ func TestIntegrationGenerateKEMKeypairUniqueness(t *testing.T) {
...
+	if deleteAfter == 0 {
+		t.Fatal("expected non-zero deleteAfter timestamp")
+	}
+}
+
+func TestIntegrationGetKemKeyNotFound(t *testing.T) {
+	_, _, _, err := GetKemKey(uuid.New())
+	if err == nil {
+		t.Fatal("expected error for non-existent UUID")
+	}
```
atulpatildbz, line 107: can we also check the expected error string:
```go
expectedErrMsg := "key_manager_get_kem_key failed with code -1"
if !strings.Contains(err.Error(), expectedErrMsg) {
    t.Fatalf("expected error containing %q, got: %v", expectedErrMsg, err)
}
```

if you decide to export typed sentinel errors `ErrKeyNotFound` then you could also use `errors.Is` but comparing string is also fine.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -61,3 +61,34 @@ func GenerateKEMKeypair(algo *algorithms.HpkeAlgorithm, bindingPubKey []byte, li
 	copy(pubkey, pubkeyBuf[:pubkeyLen])
 	return id, pubkey, nil
 }
+
+// GetKemKey retrieves KEM and binding public keys and delete_after timestamp via Rust FFI.
+func GetKemKey(id uuid.UUID) ([]byte, []byte, uint64, error) {
+	uuidBytes, err := id.MarshalBinary()
```
atulpatildbz: Since we know the size of uuid, should we create a fixed size array?

```go
var uuidBytes [16]byte
copy(uuidBytes[:], id[:])
```

keymanager/key_protection_service/service_test.go
```diff
@@ -39,12 +54,44 @@ func TestServiceGenerateKEMKeypairSuccess(t *testing.T) {
...
+	kemPubKey, bindingPubKey, deleteAfter, err := svc.GetKemKey(keyID)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if len(kemPubKey) != 32 {
+		t.Fatalf("expected 32-byte KEM public key, got %d", len(kemPubKey))
+	}
+	if len(bindingPubKey) != 32 {
+		t.Fatalf("expected 32-byte binding public key, got %d", len(bindingPubKey))
+	}
```
atulpatildbz: We should have just checked the length of the keys, we should compare it against expectedKemPubKey and expectedBindingPubKey, right?

Also it would be good to have non-zero values in expectedKemPubKey and expectedBindingPubKey

something like: 
```go
for i := range expectedKemPubKey {
    expectedKemPubKey[i] = byte(i + 10)
}
for i := range expectedBindingPubKey {
    expectedBindingPubKey[i] = byte(i + 20)
}

// Check equality of the actual sequence of bytes
if !bytes.Equal(kemPubKey, expectedKemPubKey) {
    t.Fatalf("KEM public key mismatch")
}
if !bytes.Equal(bindingPubKey, expectedBindingPubKey) {
    t.Fatalf("binding public key mismatch")
}

```

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -400,6 +400,105 @@ pub unsafe extern "C" fn key_manager_decap_and_seal(
...
+/// * `out_binding_pubkey_len` - The size of `out_binding_pubkey` buffer.
+/// * `out_delete_after` - A pointer to a u64 where the UNIX expiration timestamp will be written.
+///
+/// ## Safety
+/// This function is unsafe because it dereferences raw pointers.
+///
+/// ## Returns
+/// * `0` on success.
+/// * `-1` if arguments are invalid or key is not found.
+/// * `-2` if either public key buffer is too small.
```
atulpatildbz: we're checking != so it fails with -2 even if buffer is too large

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -331,6 +331,67 @@ pub unsafe extern "C" fn key_manager_enumerate_binding_keys(
...
+        }
+
+        // Convert to Safe Types
+        let uuid_slice = unsafe { std::slice::from_raw_parts(uuid_bytes, 16) };
+        let out_pubkey_slice =
+            unsafe { std::slice::from_raw_parts_mut(out_pubkey, out_pubkey_len) };
+
+        let uuid = match Uuid::from_slice(uuid_slice) {
+            Ok(u) => u,
+            Err(_) => return -1,
```
atulpatildbz, line 389: is this dead code? `from_raw_parts` guarantees slice is exactly 16 bytes. so `Uuid::from_slice` will never fail

keymanager/workload_service/key_custody_core/src/lib.rs
```diff
@@ -670,4 +731,58 @@ mod tests {
...
+        let result = unsafe {
+            key_manager_get_binding_key(
+                uuid_bytes.as_ptr(),
+                retrieved_pubkey_bytes.as_mut_ptr(),
+                retrieved_pubkey_bytes.len(),
+            )
+        };
+
+        assert_eq!(result, 0);
+        assert_eq!(generated_pubkey_bytes, retrieved_pubkey_bytes);
```
atulpatildbz, line 808: can we also destroy for cleanup?

if we don't cleanup, enumerate tests may become flaky `test_key_manager_enumerate_binding_keys_success` since KEY_REGISTRY sits in a global LazyLock

keymanager/workload_service/key_custody_core/integration_test.go
```diff
@@ -46,3 +46,32 @@ func TestIntegrationGenerateBindingKeypairUniqueness(t *testing.T) {
 		t.Fatal("expected non-empty public keys")
 	}
 }
+
+func TestIntegrationGetBindingKey(t *testing.T) {
+	id, pubKey, err := GenerateBindingKeypair(defaultAlgo, 3600)
```
atulpatildbz, line 52: similarly here, destroy the keys when the test completes

```go
	defer func() {
		_ = DestroyBindingKey(id)
	}()
```

keymanager/workload_service/key_custody_core/integration_test.go
```diff
@@ -46,3 +46,32 @@ func TestIntegrationGenerateBindingKeypairUniqueness(t *testing.T) {
...
+
+	if len(retrievedPubKey) != len(pubKey) {
+		t.Fatalf("expected pubkey length %d, got %d", len(pubKey), len(retrievedPubKey))
+	}
+
+	for i := range pubKey {
+		if pubKey[i] != retrievedPubKey[i] {
+			t.Fatalf("mismatch at index %d: expected %d, got %d", i, pubKey[i], retrievedPubKey[i])
+		}
+	}
```
atulpatildbz, line 72: you can use bytes.Equal here
```go
	if !bytes.Equal(pubKey, retrievedPubKey) {
		t.Fatalf("retrieved public key does not match generated public key.\nExpected: %x\nGot:      %x", pubKey, retrievedPubKey)
	}
```

keymanager/workload_service/integration_test.go
```diff
@@ -121,3 +143,97 @@ func TestIntegrationGenerateKeysUniqueMappings(t *testing.T) {
...
 		kemUUIDs[0], binding1, kemUUIDs[1], binding2)
 }
+
+func TestIntegrationKeyClaims(t *testing.T) {
+	kpsSvc := kps.NewService(&realKCC{})
+	srv, err := NewServer(kpsSvc, &realWorkloadService{}, filepath.Join(t.TempDir(), "test.sock"))
+	if err != nil {
+		t.Fatalf("failed to create server: %v", err)
+	}
+
```
atulpatildbz, line 263:
```suggestion

	t.Cleanup(func() {
		close(srv.claimsChan)
	})

```

keymanager/workload_service/server.go
```diff
@@ -255,10 +289,116 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+	// Step 3: Create KeyClaims
+	claims := &keymanager.KeyClaims{
+		Claims: &keymanager.KeyClaims_VmBindingClaims{
+			VmBindingClaims: &keymanager.KeyClaims_VmProtectionBindingClaims{
+				BindingPubKey: &keymanager.HpkePublicKey{
+					Algorithm: &keymanager.HpkeAlgorithm{
+						Kem:  keymanager.KemAlgorithm_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256,
+						Kdf:  keymanager.KdfAlgorithm_KDF_ALGORITHM_HKDF_SHA256,
+						Aead: keymanager.AeadAlgorithm_AEAD_ALGORITHM_AES_256_GCM,
+					},
```
atulpatildbz: what are your plans when we support multiple algorithms?

keymanager/workload_service/server.go
```diff
@@ -255,10 +289,116 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get binding key: %w", err)
+	}
+
+	// Step 3: Create KeyClaims
+	claims := &keymanager.KeyClaims{
+		Claims: &keymanager.KeyClaims_VmBindingClaims{
+			VmBindingClaims: &keymanager.KeyClaims_VmProtectionBindingClaims{
+				BindingPubKey: &keymanager.HpkePublicKey{
+					Algorithm: &keymanager.HpkeAlgorithm{
```
atulpatildbz: nit: can we define a package-level variable for this

```
var defaultHpkeAlgo = &keymanager.HpkeAlgorithm{
	Kem:  keymanager.KemAlgorithm_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256,
	Kdf:  keymanager.KdfAlgorithm_KDF_ALGORITHM_HKDF_SHA256,
	Aead: keymanager.AeadAlgorithm_AEAD_ALGORITHM_AES_256_GCM,
}
```

keymanager/workload_service/server_test.go
```diff
@@ -431,6 +443,182 @@ func TestHandleGetCapabilities(t *testing.T) {
...
+
+	ws := &mockWorkloadService{uuid: bindingUUID, pubKey: bindingPubKey}
+	kps := &mockKeyProtectionService{
+		uuid:          kemUUID,
+		pubKey:        kemPubKey,
+		bindingPubKey: bindingPubKey,
+		deleteAfter:   uint64(time.Now().Add(1 * time.Hour).Unix()),
+	}
+
+	srv, err := NewServer(kps, ws, filepath.Join(t.TempDir(), "test.sock"))
```
atulpatildbz: can you add cleanup?
`t.Cleanup(func() { close(srv.claimsChan) })`

keymanager/workload_service/server_test.go
```diff
@@ -431,6 +443,182 @@ func TestHandleGetCapabilities(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to create test server: %v", err)
+	}
+
+	t.Run("BindingClaims", func(t *testing.T) {
+		respChan := make(chan *ClaimsResult)
```
atulpatildbz: let's use buffered channel.

the test fails for an unrelated reason before reaching the select block, it could deadlock the test suite.

```suggestion
		respChan := make(chan *ClaimsResult, 1)
```

keymanager/workload_service/server_test.go
atulpatildbz, line 30: actually this might be a good place to add t.Cleanup

```
	t.Cleanup(func() {
		srv.listener.Close() // Or srv.Shutdown(context.Background())
		close(srv.claimsChan) // Stop the background processClaims goroutine
	})
```

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: We check that `out_kem_pubkey_len` is valid ( != 0) in the if condition above, so at this point we should be safe to create the slice. 

However, at this point we don't know the exact size of the out_kem_key so we can't check against the size.  Thats why before the final key write, we do a second check below. Let me know if you think otherwise.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: I would prefer keeping the size check to be strict. Allowing size overcommit can open us to all sorts of abuse. In most of these cases we should know the exact size of the output we are dealing with so I think its better to keep the check strict.

I initially thought about allowing it but then decided against to protect from bugs: https://github.com/google/go-tpm-tools/pull/645#discussion_r2800990184

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: yes, we are keeping the checks strict as of now. We can look into in future if a need arises. For more context: https://github.com/google/go-tpm-tools/pull/645#discussion_r2800990184

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: Ack added.

keymanager/key_protection_service/key_custody_core/integration_test.go
NilanjanDaw: Makes sense. Done.

keymanager/key_protection_service/key_custody_core/integration_test.go
NilanjanDaw: Ack, done.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
NilanjanDaw: thanks that makes sense. Updated.

keymanager/key_protection_service/service_test.go
NilanjanDaw: Hmm yeah that's better. Added.

keymanager/workload_service/server_test.go
NilanjanDaw: Thanks of pointing this out. Added the cleanup function.

keymanager/workload_service/server_test.go
NilanjanDaw: Ack, done.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: I think its just some defensive coding, directly adding the data to the slice without an error handling makes me a bit nervous.

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Hmm that makes sense, since cargo tests run in parallel, other tests can cause stale keys to remain the `KEY_REGISTRY`. However, the exact opposite problem also comes up with a clean up. Cleaning up the keys instantly wipes out keys needed by other tests running in parallel which causes them to fail.

The way we are testing the enumerate, we are checking if keys >= 1 should not fail even if stale keys are present so I think the tests should be fine for now. What do you think?

keymanager/workload_service/key_custody_core/integration_test.go
NilanjanDaw: Similar comment to above.

keymanager/workload_service/integration_test.go
NilanjanDaw: Ack done.

keymanager/workload_service/key_custody_core/integration_test.go
NilanjanDaw: Done updated.

keymanager/workload_service/server.go
NilanjanDaw: I think currently we are using the full `HpkeAlgorithm` only twice. Probably add it in future?

keymanager/workload_service/server_test.go
NilanjanDaw: Done added to the newTestServer

keymanager/workload_service/server.go
NilanjanDaw: Hmm good point. I modified the interfaces to now pass the algorithms directly through the FFI.

pgonda (approved):

pgonda (approved):

keymanager/workload_service/server_test.go
```diff
@@ -403,9 +403,11 @@ func TestToHpkeAlgorithm(t *testing.T) {
 }
 
 func TestHandleGetCapabilities(t *testing.T) {
+	bindingGen := &mockWorkloadService{}
```
pgonda, line 466: Is the code broken before this change?

If so please fix earlier commit so bisect continues to work

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -6,7 +6,7 @@ package kpskcc
 /*
 #cgo LDFLAGS: -L${SRCDIR}/../../target/release -L${SRCDIR}/../../target/debug -lkps_key_custody_core
 #cgo LDFLAGS: -lcrypto -lssl
-#cgo LDFLAGS: -lpthread -ldl -lm -lstdc++
+#cgo LDFLAGS: -lpthread -ldl -lm -static
```
pgonda: Commit description should say why

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -6,7 +6,7 @@ package kpskcc
 /*
 #cgo LDFLAGS: -L${SRCDIR}/../../target/release -L${SRCDIR}/../../target/debug -lkps_key_custody_core
 #cgo LDFLAGS: -lcrypto -lssl
-#cgo LDFLAGS: -lpthread -ldl -lm -static
+#cgo LDFLAGS: -lpthread -ldl -lm -lstdc++
```
pgonda, line 9: Why does this commit undo the last one?

.goreleaser.yaml
```diff
@@ -7,7 +7,12 @@ builds:
       - amd64
       - arm64
     env:
-      - CGO_ENABLED=1
+      - CGO_ENABLED=0
+    overrides:
+      - goos: linux
+        goarch: amd64
+        env:
+          - CGO_ENABLED=1
```
pgonda: Didn't we just change this a few commits go?

Please do this all at once in this series

keymanager/workload_service/server_test.go
```diff
@@ -405,7 +405,7 @@ func TestToHpkeAlgorithm(t *testing.T) {
 func TestHandleGetCapabilities(t *testing.T) {
 	bindingGen := &mockWorkloadService{}
 	kemGen := &mockKeyProtectionService{}
-	srv := NewServer(
+	srv := newTestServer(t,
```
pgonda: The Commit description needs more detail

keymanager/workload_service/server_test.go
```diff
@@ -405,7 +405,7 @@ func TestToHpkeAlgorithm(t *testing.T) {
 func TestHandleGetCapabilities(t *testing.T) {
 	bindingGen := &mockWorkloadService{}
 	kemGen := &mockKeyProtectionService{}
-	srv := newTestServer(t, 
+	srv := newTestServer(t,
```
pgonda, line 468: Again please rollup fix commits in the series

launcher/image/vgexperiment.json
```diff
@@ -1 +1 @@
-{"EnableB200DriverInstallation": false,"EnableKeyManager": true}
+{"EnableB200DriverInstallation": true,"EnableKeyManager": true}
```
pgonda, line 1: Should the commits editing these files even be in the series?

keymanager/key_protection_service/key_custody_core/include/kps_key_custody_core.h
```diff
@@ -52,6 +52,13 @@ int32_t key_manager_decap_and_seal(const uint8_t *uuid_bytes,
                                    uint8_t *out_ciphertext,
                                    size_t out_ciphertext_len);
 
+int32_t key_manager_get_kem_key(const uint8_t *uuid_bytes,
+                                uint8_t *out_kem_pubkey,
```
pgonda, line 56: Not a huge fan of annotating in the name in and out, but that ship has sailed.

keymanager/key_protection_service/key_custody_core/integration_test.go
pgonda: I don't see this done?

keymanager/key_protection_service/key_custody_core/integration_test.go
```diff
@@ -61,3 +61,49 @@ func TestIntegrationGenerateKEMKeypairUniqueness(t *testing.T) {
...
+		}
+	}
+
+	if deleteAfter == 0 {
+		t.Fatal("expected non-zero deleteAfter timestamp")
+	}
+}
+
+func TestIntegrationGetKemKeyNotFound(t *testing.T) {
+	_, _, _, err := GetKemKey(uuid.New())
```
pgonda: Combine into 

if _,_,err := .. {
..
}

Statement if you are not keeping any values.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
```diff
@@ -61,3 +61,34 @@ func GenerateKEMKeypair(algo *keymanager.HpkeAlgorithm, bindingPubKey []byte, li
...
 }
+
+// GetKemKey retrieves KEM and binding public keys and delete_after timestamp via Rust FFI.
+func GetKemKey(id uuid.UUID) ([]byte, []byte, uint64, error) {
+	uuidBytes, err := id.MarshalBinary()
+	if err != nil {
+		return nil, nil, 0, fmt.Errorf("failed to marshal UUID: %v", err)
+	}
+
+	var kemPubkeyBuf [32]byte
```
pgonda: Please move these magics into consts.

keymanager/workload_service/server_test.go
```diff
@@ -435,6 +447,182 @@ func TestHandleGetCapabilities(t *testing.T) {
...
+			if res.Err == nil {
+				t.Fatal("expected error for KEM key not found")
+			}
+			if !strings.Contains(res.Err.Error(), "failed to get KEM key") {
+				t.Errorf("expected error to contain 'failed to get KEM key', got %v", res.Err)
+			}
+		case <-time.After(2 * time.Second):
+			t.Fatal("timed out waiting for response")
+		}
+	})
```
pgonda, line 694: This commit is huge and I imagine could be split a bit.

keymanager/key_protection_service/service_test.go
```diff
@@ -9,16 +9,16 @@ import (
 	keymanager "github.com/google/go-tpm-tools/keymanager/km_common/proto"
 )
 
-type mockKCC struct {
+type mockKeyProtectionService struct {
```
pgonda: Just roll this commit into the original one.

keymanager/workload_service/key_custody_core/include/ws_key_custody_core.h
```diff
@@ -48,6 +48,10 @@ int32_t key_manager_enumerate_binding_keys(WsKeyInfo *out_entries,
                                            size_t offset,
                                            bool *out_has_more);
 
+int32_t key_manager_get_binding_key(const uint8_t *uuid_bytes,
```
pgonda, line 51: Again fix the lint errors in the actual commit they are seen in.

keymanager/key_protection_service/service.go
```diff
@@ -9,23 +9,30 @@ import (
 	keymanager "github.com/google/go-tpm-tools/keymanager/km_common/proto"
 )
 
-// KEMKeyGenerator generates KEM keypairs linked to a binding public key.
-type KEMKeyGenerator interface {
-	GenerateKEMKeypair(algo *keymanager.HpkeAlgorithm, bindingPubKey []byte, lifespanSecs uint64) ([]byte, []byte, error)
+// KeyProtectionService defines the interface for the underlying Key Custody Core operations.
+type KeyProtectionService interface {
+	GenerateKEMKeypair(algo *keymanager.HpkeAlgorithm, bindingPubKey []byte, lifespanSecs uint64) (uuid.UUID, []byte, error)
+	GetKemKey(id uuid.UUID) ([]byte, []byte, *keymanager.HpkeAlgorithm, uint64, error)
```
alexmwu: GetKEMKey

keymanager/km_common/proto/algorithms.proto
```diff
@@ -24,7 +24,7 @@ enum AeadAlgorithm {
 
 // Composite HPKE Algorithm suite
 message HpkeAlgorithm {
-  KemAlgorithm kem = 1;
-  KdfAlgorithm kdf = 2;
+  KemAlgorithm  kem  = 1;
```
alexmwu, line 27: Here and elsewhere, any reason you're changing the indentation? I'm not sure this is in the styleguide and it's also inconsistent with other proto files.

keymanager/km_common/proto/key_claims.proto
alexmwu, line 1: Weird proto indentation in this file as well

keymanager/workload_service/key_custody_core/src/lib.rs
alexmwu: Drive by comment, but why can't we create separate key registries for separate tests?

keymanager/key_protection_service/service.go
```diff
@@ -9,23 +9,30 @@ import (
 	keymanager "github.com/google/go-tpm-tools/keymanager/km_common/proto"
 )
 
-// KEMKeyGenerator generates KEM keypairs linked to a binding public key.
-type KEMKeyGenerator interface {
-	GenerateKEMKeypair(algo *keymanager.HpkeAlgorithm, bindingPubKey []byte, lifespanSecs uint64) ([]byte, []byte, error)
+// KeyProtectionService defines the interface for the underlying Key Custody Core operations.
+type KeyProtectionService interface {
+	GenerateKEMKeypair(algo *keymanager.HpkeAlgorithm, bindingPubKey []byte, lifespanSecs uint64) (uuid.UUID, []byte, error)
```
alexmwu, line 29: Please document these a bit better, especially regarding the expected outputs.

keymanager/workload_service/server.go
```diff
@@ -23,23 +26,47 @@ import (
 // WorkloadService defines the interface for generating binding keypairs.
 type WorkloadService interface {
 	GenerateBindingKeypair(algo *keymanager.HpkeAlgorithm, lifespanSecs uint64) (uuid.UUID, []byte, error)
```
alexmwu, line 43: Please document the methods on this interface

keymanager/workload_service/server.go
```diff
@@ -257,10 +291,108 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
 func writeError(w http.ResponseWriter, message string, code int) {
 	writeJSON(w, map[string]string{"error": message}, code)
 }
+
+// GetBindingKeyClaims returns the claims for a binding key identified by its UUID.
+func (s *Server) GetBindingKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error) {
+	// Step 1: Key ID Lookup. The orchestration layer will look-up the key_handle
+	// in its ActiveKeyRegistry to find the Binding Key ID.
+	bindingID := id
+	if bid, ok := s.LookupBindingUUID(id); ok {
```
alexmwu: I'm not sure what this does vs workloadService.GetBindingKey, but should we handle `!ok`?

keymanager/workload_service/server.go
```diff
@@ -257,10 +291,108 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+					PublicKey: pubKey,
+				},
+			},
+		},
+	}
+	return claims, nil
+}
+
+// GetKemKeyClaims returns the claims for a KEM key identified by its UUID.
+func (s *Server) GetKemKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error) {
```
alexmwu: GetKEMKeyClaims

keymanager/key_protection_service/key_custody_core/src/lib.rs
```diff
@@ -400,6 +400,118 @@ pub unsafe extern "C" fn key_manager_decap_and_seal(
...
+        KeySpec::KemWithBindingPub {
+            algo,
+            kem_public_key,
+            binding_public_key,
+            ..
+        } => {
+            let remaining = record
+                .meta
+                .delete_after
+                .saturating_duration_since(std::time::Instant::now());
```
alexmwu, line 416: Instant is monotonic but SystemTime is subject to clock drift

keymanager/workload_service/server.go
```diff
@@ -102,29 +129,33 @@ type GetCapabilitiesResponse struct {
...
 
 // NewServer creates a new WSD server with the given dependencies.
-func NewServer(keyProtectionService KeyProtectionService, workloadService WorkloadService, socketPath string) (*Server, error) {
+func NewServer(keyProtectionService kps.KeyProtectionService, workloadService WorkloadService, socketPath string) (*Server, error) {
 	s := &Server{
 		keyProtectionService: keyProtectionService,
 		workloadService:      workloadService,
 		kemToBindingMap:      make(map[uuid.UUID]uuid.UUID),
 		mu:                   sync.RWMutex{},
+		claimsChan:           make(chan *ClaimsCall, 100),
```
alexmwu: Do we really need a size 100 buffered chan? Calls to GetKeyEndorsement should be rare. 2 might be a better number since they'll get both the binding and protection key claims.

alexmwu: Do we really need a buffered chan? Calls to GetKeyEndorsement should be rare

keymanager/workload_service/server.go
```diff
@@ -257,10 +291,108 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+		if err != nil {
+			call.RespChan <- &ClaimsResult{Err: fmt.Errorf("failed to retrieve key claims: %w", err)}
+			continue
+		}
+		var claims *keymanager.KeyClaims
+		switch keyType {
+		case keymanager.KeyType_KEY_TYPE_VM_PROTECTION_BINDING:
+			claims, err = s.GetBindingKeyClaims(id)
+			if err != nil {
+				call.RespChan <- &ClaimsResult{Err: fmt.Errorf("failed to retrieve key claims: %w", err)}
```
alexmwu: failed to retrieve binding key claims

keymanager/workload_service/server.go
```diff
@@ -257,10 +291,108 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+				continue
+			}
+
+		case keymanager.KeyType_KEY_TYPE_VM_PROTECTION_KEY:
+			claims, err = s.GetKemKeyClaims(id)
+			if err != nil {
+				call.RespChan <- &ClaimsResult{Err: fmt.Errorf("failed to retrieve key claims: %w", err)}
```
alexmwu: failed to retrieve VM protection key claims

keymanager/key_protection_service/key_custody_core/include/kps_key_custody_core.h
NilanjanDaw: these are autogenerated header files by `cbindgen`, and generally follows the naming convention for these kind of glue files.

keymanager/km_common/proto/algorithms.proto
NilanjanDaw: We recently added a protobuf linter which is making these changes.

keymanager/km_common/proto/key_claims.proto
NilanjanDaw: Same comment as above, done by protobuf linters.

keymanager/workload_service/key_custody_core/include/ws_key_custody_core.h
NilanjanDaw: Sorry didn't get your comment here?

keymanager/workload_service/server_test.go
NilanjanDaw: As mentioned offline, the intermediate commits will be squashed and merged. So these won't show up post merge (or break bisect operations)

keymanager/workload_service/server_test.go
NilanjanDaw: similar comment as above.

keymanager/key_protection_service/key_custody_core/src/lib.rs
NilanjanDaw: simplified the time management, using Instant for all now.

keymanager/key_protection_service/key_custody_core/integration_test.go
NilanjanDaw: I think you're seeing an outdated view here, https://github.com/NilanjanDaw/go-tpm-tools/blob/b8efe7144e578f7ee8db2dab3cc9d955728a92c3/keymanager/key_protection_service/key_custody_core/integration_test.go#L83-L90 for the latest view.

keymanager/workload_service/server_test.go
NilanjanDaw: I squashed all the commits to a single one now. So should be easier to review.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
NilanjanDaw: these were intermediate commits, and can be ignored. Also this were from a different PR. Resolving, feel free to reopen.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
NilanjanDaw: these were intermediate commits, and can be ignored. Also this were from a different PR. Resolving feel free to reopen

.goreleaser.yaml
NilanjanDaw: these were intermediate commits, and can be ignored. Also this were from a different PR. Resolving feel free to reopen

keymanager/workload_service/server_test.go
NilanjanDaw: these were intermediate commits, and can be ignored. Also this were from a different PR. I have squashed and merged all of them for easy reviewing. Resolving feel free to reopen.

keymanager/workload_service/server_test.go
NilanjanDaw: these were intermediate commits, and can be ignored. Also this were from a different PR. Resolving feel free to reopen

launcher/image/vgexperiment.json
NilanjanDaw: these were intermediate commits, and can be ignored. Also this were from a different PR. Resolving feel free to reopen

keymanager/key_protection_service/key_custody_core/integration_test.go
NilanjanDaw: We use the err a few lines below to check for the expected error string, so I think its better to store the err.
```go
if !strings.Contains(err.Error(), expectedErrMsg) {
		t.Fatalf("expected error containing %q, got: %v", expectedErrMsg, err)
	}
```

keymanager/key_protection_service/service_test.go
NilanjanDaw: Commit amends are generally discouraged since it rewrites the git history. It additionally breaks stacked PRs and causes weird conflicts. However, I have squashed all commits into one for now for easy reviewing.

keymanager/key_protection_service/service.go
NilanjanDaw: Added descriptive comments to the interface. Thanks for pointing this out.

keymanager/workload_service/server.go
NilanjanDaw: Done added documentation.

keymanager/workload_service/server.go
NilanjanDaw: Good catch, thanks. Updated to handle `!ok`.

keymanager/workload_service/server.go
NilanjanDaw: Ack updated.

keymanager/workload_service/server.go
NilanjanDaw: Hmm makes sense let me reduce it to 2 (maybe 4)? Just in case?

NilanjanDaw: Yeah, but I feel its still better to have a buffered channel just in case usage sees a spike.

keymanager/workload_service/server.go
NilanjanDaw: Done.

keymanager/workload_service/server.go
NilanjanDaw: Done.

keymanager/key_protection_service/key_custody_core/kps_key_custody_core_cgo.go
NilanjanDaw: Done updated. commit [#b89ddb](https://github.com/google/go-tpm-tools/pull/681/changes/b89ddb9bcc1ab9d491c5de9c481957236600ff55)

keymanager/key_protection_service/service.go
NilanjanDaw: Done.

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: Makes sense.

keymanager/workload_service/server.go
atulpatildbz: SGTM

keymanager/key_protection_service/key_custody_core/src/lib.rs
atulpatildbz: Got it

keymanager/workload_service/key_custody_core/src/lib.rs
atulpatildbz: > Hmm that makes sense, since cargo tests run in parallel, other tests can cause stale keys to remain the `KEY_REGISTRY`. However, the exact opposite problem also comes up with a clean up. Cleaning up the keys instantly wipes out keys needed by other tests running in parallel which causes them to fail.
> 
> The way we are testing the enumerate, we are checking if keys >= 1 should not fail even if stale keys are present so I think the tests should be fine for now. What do you think?

That is true if we wipe the entire registry during cleanup.
But we can do a targeted destroy every test, right? `key_manager_destroy_binding_key(uuid_bytes.as_ptr())`
deleting only the specific UUID it just generated



> Drive by comment, but why can't we create separate key registries for separate tests?

I think this will need some architectural change. Currently we follow a global singleton pattern, We would have to change it to some kind of object/handle based design.

keymanager/workload_service/server.go
```diff
@@ -159,6 +237,7 @@ func NewServer(keyProtectionService KeyProtectionService, workloadService Worklo
 		workloadService:      workloadService,
 		kemToBindingMap:      make(map[uuid.UUID]uuid.UUID),
 		mu:                   sync.RWMutex{},
+		claimsChan:           make(chan *ClaimsCall, 4),
```
jkl73, line 273: any particular reason to pick 4?

keymanager/workload_service/server.go
```diff
@@ -364,3 +446,97 @@ func writeJSON(w http.ResponseWriter, v any, code int) {
...
+			if err != nil {
+				call.RespChan <- &ClaimsResult{Err: fmt.Errorf("failed to retrieve VM protection key claims: %w", err)}
+				continue
+			}
+		default:
+			call.RespChan <- &ClaimsResult{Err: fmt.Errorf("unsupported key type: %v", keyType)}
+			continue
+		}
+
+		call.RespChan <- &ClaimsResult{Reply: claims}
```
jkl73: RespChan might get blocked (if the caller messed up the channel, which I guess it should be fine here since the attestation agent is the caller?)

jkl73 (approved):
lgtm on the server/service code side

atulpatildbz (approved):

keymanager/workload_service/key_custody_core/src/lib.rs
NilanjanDaw: Makes sense, added a drop function to clean up keys after the tests exit.

Also +1 to @atulpatildbz comment, the current implementation is dependent on a single global registry for simplicity. We would need to re-architect a bit to handle separate registries.

keymanager/workload_service/server.go
NilanjanDaw: No particular reason, I initially had it as 100. But @alexmwu suggested that since attestations are going to be rarely used, 100 is too large of a number and is just over-commitment of resources. So I reduced it to 4.

keymanager/workload_service/server.go
NilanjanDaw: Ah yes I missed that! Thanks for pointing this out. Updated the implementation to add a timeout of 5 seconds to handle any channel freezes.

keymanager/workload_service/server.go
```diff
@@ -572,37 +578,45 @@ func (s *Server) GetKEMKeyClaims(id uuid.UUID) (*keymanager.KeyClaims, error) {
...
-		keyType := req.GetKeyType()
+		result := s.handleGetClaims(call.Request)
 
-		id, err := uuid.Parse(keyHandle)
-		if err != nil {
-			call.RespChan <- &ClaimsResult{Err: fmt.Errorf("failed to retrieve key claims: %w", err)}
-			continue
+		select {
+		case call.RespChan <- result:
+		case <-time.After(ClaimsResponseTimeout):
```
atulpatildbz, line 588: Can we use `time.NewTimer`?

timer created with `time.After` would stay in memory until time passes. 

time.NewTimer gives us control to stop the timer early if we get the result.

```go
timer := time.NewTimer(5 * time.Second)

select {
case call.RespChan <- result:
	timer.Stop() 
case <-timer.C:
	log.Printf("timed out")
}
```

keymanager/workload_service/server_test.go
```diff
@@ -1024,3 +1024,50 @@ func TestHandleDecapsUnsupportedAlgorithm(t *testing.T) {
 		t.Fatalf("expected status 400, got %d", w.Code)
 	}
 }
+
+func TestProcessClaimsTimeout(t *testing.T) {
+	oldTimeout := ClaimsResponseTimeout
+	ClaimsResponseTimeout = 10 * time.Millisecond
+	defer func() { ClaimsResponseTimeout = oldTimeout }()
```
atulpatildbz, line 1031: this is fine for now but if we ever add `t.Parallel()` it could result in flaky tests.

In future we can make this configurable field in `Server` struct

NilanjanDaw: Thanks for pointing this out. Let me fix this is in a follow up PR. Filed Issue #692

PR #691, opened 2026-03-04
migrate models.NvidiaAttestation to attestation proto library

yawangwang:
With the attestation proto library available [here](https://github.com/GoogleCloudPlatform/confidential-space/blob/main/server/proto/gen/attestation/attestation.pb.go), gpu attestation collection should be consistent with that.


launcher/internal/gpu/attestation_test.go
```diff
@@ -124,5 +125,37 @@ func TestDetermineAttestationType(t *testing.T) {
...
+			wantArch: attestationpb.GpuArchitectureType_GPU_ARCHITECTURE_TYPE_HOPPER,
+		},
+		{
+			name:     "BLACKWELL",
+			arch:     "BLACKWELL",
+			wantArch: attestationpb.GpuArchitectureType_GPU_ARCHITECTURE_TYPE_BLACKWELL,
+		},
+		{
+			name:     "UNSPECIFIED",
+			arch:     "UNSPECIFIED",
```
jessieqliu, line 144: nit: if these are always the same value then there's no need to have both fields in the struct

yawangwang:
> We also need to remove https://github.com/GoogleCloudPlatform/confidential-space/blob/main/server/proto/attestation.proto#L117-L119 and setting it in this API.

Are you suggesting moving this field (not string type but `DeviceAttestationReport`) to TdCcelQuote https://github.com/GoogleCloudPlatform/confidential-space/blob/8e0b07455f87ce15515f8dbb29d86c65166181cb/server/proto/attestation.proto#L77 and export it in https://github.com/google/go-tpm-tools/blob/262ec22c2f0141738f2360d79bd32b3dd32414db/launcher/agent/agent.go#L336-L339?

launcher/internal/gpu/attestation_test.go
yawangwang: Done.

PR #694, opened 2026-03-06
Add EnableRuntimeGPUAttestation opts to AttestationEvidence API

yawangwang:
By default, runtime GPU attestation is disabled during raw evidence collection. It is enabled only when the /v1/evidence API is called

### DIFF
--- a/launcher/agent/agent.go
+++ b/launcher/agent/agent.go
@@ -87,7 +87,12 @@ type DeviceROT interface {
 // AttestAgentOpts contains user generated options when calling the
 // VerifyAttestation API
 type AttestAgentOpts struct {
-	TokenOptions                *models.TokenOptions
+	TokenOptions *models.TokenOptions
+	*DeviceReportOpts
+}
+
+// DeviceReportOpts contains options for runtime device attestations.
+type DeviceReportOpts struct {
 	EnableRuntimeGPUAttestation bool
 }
 
@@ -355,6 +360,9 @@ func (a *agent) AttestationEvidence(_ context.Context, challenge []byte, extraDa
 }
 
 func (a *agent) attestDeviceROTs(nonce []byte, opts AttestAgentOpts) ([]*attestationpb.DeviceAttestationReport, error) {
+	if opts.DeviceReportOpts == nil {
+		return nil, nil
+	}
 	deviceROTs, err := a.avRot.AttestDeviceROTs(nonce)
 	if err != nil {
 		return nil, err
@@ -364,7 +372,7 @@ func (a *agent) attestDeviceROTs(nonce []byte, opts AttestAgentOpts) ([]*attesta
 	for _, dr := range deviceROTs {
 		switch v := dr.(type) {
 		case *attestationpb.NvidiaAttestationReport:
-			if opts.EnableRuntimeGPUAttestation {
+			if opts.DeviceReportOpts.EnableRuntimeGPUAttestation {
 				deviceReports = append(deviceReports, &attestationpb.DeviceAttestationReport{
 					Report: &attestationpb.DeviceAttestationReport_NvidiaReport{
 						NvidiaReport: v,
@@ -457,15 +465,7 @@ func (t *tpmAttestRoot) AttestDeviceROTs(nonce []byte) ([]any, error) {
 	t.tpmMu.Lock()
 	defer t.tpmMu.Unlock()
 
-	var deviceReports []any
-	for _, deviceROT := range t.deviceROTs {
-		deviceReport, err := deviceROT.Attest(nonce)
-		if err != nil {
-			return nil, err
-		}
-		deviceReports = append(deviceReports, deviceReport)
-	}
-	return deviceReports, nil
+	return doAttestDeviceROTs(t.deviceROTs, nonce)
 }
 
 type tdxAttestRoot struct {
@@ -517,20 +517,7 @@ func (t *tdxAttestRoot) AttestDeviceROTs(nonce []byte) ([]any, error) {
 	t.tdxMu.Lock()
 	defer t.tdxMu.Unlock()
 
-	var deviceReports []any
-	for _, deviceRoT := range t.deviceROTs {
-		att, err := deviceRoT.Attest(nonce)
-		if err != nil {
-			return nil, err
-		}
-		switch v := att.(type) {
-		case *attestationpb.NvidiaAttestationReport:
-			deviceReports = append(deviceReports, v)
-		default:
-			return nil, fmt.Errorf("unknown device attestation type: %T", v)
-		}
-	}
-	return deviceReports, nil
+	return doAttestDeviceROTs(t.deviceROTs, nonce)
 }
 
 func (t *tdxAttestRoot) ComputeNonce(challenge []byte, extraData []byte) []byte {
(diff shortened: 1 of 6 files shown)


launcher/teeserver/tee_server.go
```diff
@@ -157,7 +158,7 @@ func (a *attestHandler) getAttestationEvidence(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	evidence, err := a.attestAgent.AttestationEvidence(a.ctx, req.Challenge, nil)
+	evidence, err := a.attestAgent.AttestationEvidence(a.ctx, req.Challenge, nil, agent.AttestAgentOpts{EnableRuntimeGPUAttestation: true})
```
alexmwu: This should be false getAttestationEvidence and true for getToken

launcher/agent/agent.go
```diff
@@ -85,7 +86,8 @@ type DeviceROT interface {
 // AttestAgentOpts contains user generated options when calling the
 // VerifyAttestation API
 type AttestAgentOpts struct {
-	TokenOptions *models.TokenOptions
+	TokenOptions                *models.TokenOptions
+	EnableRuntimeGPUAttestation bool
```
alexmwu, line 96: EnableRuntimeDeviceAttestation? This will probably be true for TPUs and GPUs

launcher/agent/agent.go
```diff
@@ -340,6 +342,18 @@ func (a *agent) AttestationEvidence(_ context.Context, challenge []byte, extraDa
 				TdQuote:           v.TdQuote,
 			},
 		}
+		if opts.EnableRuntimeGPUAttestation && v.NvidiaAttestation != nil {
```
alexmwu: nit: v.NvidiaAttestation won't work for TPM or non-GPUs. Can we abstract this a little better? Perhaps we can migrate to using the AttestationEvidence API protos directly instead of having another layer at the verifier level. Then, this code would just set the device_reports field in attestation.proto (we should change that field back to type DeviceAttestationReport)

launcher/teeserver/tee_server.go
yawangwang: My understanding is that `getAttestationEvidence` is primarily exposed via the /v1/evidence API, which is limited to regular CS images requiring runtime attestation (for prod release tests purpose), so this should be `true` for `getAttestationEvidence`.

Additionally, since `getToken` should always enable runtime GPU attestation by default, I’m considering flipping the logic to a `DisableRuntimeGPUAttestation` flag. This ensures the capability for runtime attestation is on by default for most verifiers, requiring the Key Manager API to explicitly disable it when calling `AttestationEvidence`, WDYT?

alexmwu: > My understanding is that getAttestationEvidence is primarily exposed via the /v1/evidence API, which is limited to regular CS images requiring runtime attestation (for prod release tests purpose), so this should be true for getAttestationEvidence.
Ah, that's right. Since it's just for internal tests, can we control this with a partial response and fields query parameter? https://google.aip.dev/157. Here's an example: https://developers.google.com/workspace/drive/api/guides/fields-parameter

> flipping the logic to a DisableRuntimeGPUAttestation flag
In general, opt-in is safer than opt-out.

launcher/agent/agent.go
alexmwu: Could probably also drop the Enable prefix

launcher/teeserver/tee_server.go
yawangwang: Done

launcher/agent/agent.go
yawangwang: I realized if we migrate to using the AttestationEvidence API protos directly it will be a larger refactor work. 

However, I figured out an alternative approach by adding a new interface method "AttestDeviceROTs(nonce []byte) ([]any, error)" for better abstraction.

launcher/agent/agent.go
```diff
@@ -343,9 +344,38 @@ func (a *agent) AttestationEvidence(_ context.Context, challenge []byte, extraDa
...
+	if err != nil {
+		return nil, err
+	}
+	attestation.DeviceReports = deviceReports
+
 	return attestation, nil
 }
 
+func (a *agent) attestDeviceROTs(nonce []byte, opts AttestAgentOpts) ([]*attestationpb.DeviceAttestationReport, error) {
+	deviceROTs, err := a.avRot.AttestDeviceROTs(nonce)
```
alexmwu, line 366: The check for RuntimeGPUAttestation should happen before we call actually call attest to avoid wasteful attestation requests

yawangwang: I'm thinking about the corner case if we do an early exit on "RuntimeGPUAttestation", this flow will break the attestation flow for more device RoTs. 

So I'm adding a new struct "DeviceReportOpts" that includes "EnableRuntimeGPUAttestation" as its sub field. That way we can do the nil check for DeviceReportOpts to avoid wasteful attestation requests while being future compatible with more device RoTs.

launcher/agent/agent.go
```diff
@@ -461,26 +514,31 @@ func (t *tdxAttestRoot) Attest(nonce []byte) (any, error) {
...
+	t.tdxMu.Lock()
+	defer t.tdxMu.Unlock()
+
+	var deviceReports []any
 	for _, deviceRoT := range t.deviceROTs {
 		att, err := deviceRoT.Attest(nonce)
 		if err != nil {
 			return nil, err
 		}
 		switch v := att.(type) {
```
alexmwu: Why is there a type check here but not in (t *tpmAttestRoot) AttestDeviceROTs?

launcher/agent/agent.go
```diff
@@ -461,26 +514,31 @@ func (t *tdxAttestRoot) Attest(nonce []byte) (any, error) {
...
+		CcelData:      ccelData,
+		TdQuote:       rawQuote,
+	}, nil
+}
+
+func (t *tdxAttestRoot) AttestDeviceROTs(nonce []byte) ([]any, error) {
+	t.tdxMu.Lock()
+	defer t.tdxMu.Unlock()
+
+	var deviceReports []any
```
alexmwu: (t *tdxAttestRoot) AttestDeviceROTs and (t *tpmAttestRoot) AttestDeviceROTs are functionally the same. Should they call the same helper method?

launcher/teeserver/tee_server.go
```diff
@@ -169,13 +173,20 @@ func (a *attestHandler) getAttestationEvidence(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	evidence, err := a.attestAgent.AttestationEvidence(a.ctx, req.Challenge, nil)
+	fullEvidence, err := a.attestAgent.AttestationEvidence(a.ctx, req.Challenge, nil, agent.AttestAgentOpts{DeviceReportOpts: &agent.DeviceReportOpts{EnableRuntimeGPUAttestation: true}})
```
alexmwu: You're always setting runtime attestation to true. Instead, we should only set it if fields == '*' or fields contains 'deviceReports'

launcher/agent/agent.go
yawangwang: My original thought was the lower-level implementations for "AttestDeviceROTs" under different CPU RoTs would divert. For example, (t *tpmAttestRoot) doesn't support GPU attestation so it shouldn't share the same logics with (t *tdxAttestRoot).

But now since attestation agent will do the type check for GPU attestation at a higher-level, it makes sense to have them share the same helper method.

launcher/teeserver/tee_server.go
yawangwang: Done

PR #695, opened 2026-03-06
Send SPT attestation to GCA

yawangwang:


launcher/agent/agent.go
```diff
@@ -280,6 +280,19 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 		return nil, fmt.Errorf("received an unsupported attestation type! %v", v)
 	}
 
+	deviceReports, err := a.attestDeviceROTs(challenge.Nonce, opts)
```
alexmwu: This should go in the experiment check. We only have NV support, and we should gate all new behavior

yawangwang: Done.

PR #707, opened 2026-03-11
Refactor ContainerRunner to call the attestation agent for GPU Attestation

Sibcgh:
### Summary
Refactors attached accelerator attestation into a generic `device.ROTManager` (`agent/device`), completely decoupling hardware devices (GPUs, TPUs, etc.) from `AttestationAgent`. Addresses [#659 feedback](https://github.com/google/go-tpm-tools/pull/659#discussion_r2915996384) to remove hardcoded `GPUAttester` logic from `ContainerRunner`.

### Key Changes
* **`agent/device`**:
  * Introduces `device.ROT` (`Attest`, `Vendor`) and optional `device.ReadyStateEnabler` (`EnableReadyState`).
  * Introduces `ROTManager` to handle both runtime reports (`AttestDeviceROTs`) and boot-time container measurement (`MeasureDeviceEvidence`).
* **Decoupled Architecture**: Removed all GPU methods from `AttestationAgent` (`agent/agent.go`) and updated `ContainerRunner` to delegate accelerator measurement directly to `ROTManager`.

### Credit
Adopts and builds upon the `ROTManager` architectural pattern prototyped in #718.


agent/agent.go
```diff
@@ -99,6 +102,12 @@ type DeviceROT interface {
 	Attest(nonce []byte) (any, error)
 }
 
+// GPUAttester defines the interface for GPU attestation.
+type GPUAttester interface {
+	Attest(nonce []byte) (any, error)
```
susienme: Consider embedding the interface.

I think the idea is some of deviceROTs will implement this interface as well as `DeviceROT`. If `DeviceROT` changes you also need to change this interface. To avoid, embed the interface here.

agent/agent.go
```diff
@@ -210,6 +220,13 @@ func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher
 
 	// Add deviceRoTs to the CPU attestation root.
 	attestAgent.avRot.AddDeviceROTs(deviceROTs)
+
+	for _, dr := range deviceROTs {
+		if nv, ok := dr.(GPUAttester); ok {
+			attestAgent.nvidiaAttester = nv
```
susienme: How many elements do you expect from `deviceROTs`? Consider breaking after this.

Can there be multiple elements implementing this interface? If so, is it okay use the last one?

agent/agent.go
```diff
@@ -803,6 +820,24 @@ func convertToTPMQuote(v *pb.Attestation) *attestationpb.TpmQuote {
...
 
+func (a *agent) HasGPU() bool {
+	return a.nvidiaAttester != nil
+}
+
+func (a *agent) EnableGPUReadyState() error {
+	if a.nvidiaAttester == nil {
+		return fmt.Errorf("GPU attester not initialized")
+	}
+	return a.nvidiaAttester.EnableReadyState()
```
susienme: Is `EnableReadyState()` idempotent? I.e., what are the consequences if this function is called multiple times? If this function should not be called multiple times, consider using a flag.

agent/agent.go
```diff
@@ -803,6 +820,24 @@ func convertToTPMQuote(v *pb.Attestation) *attestationpb.TpmQuote {
...
+}
+
+func (a *agent) GetGPUAttestation(nonce []byte) (any, error) {
+	if a.nvidiaAttester == nil {
+		return nil, fmt.Errorf("GPU attester not initialized")
+	}
+	return a.nvidiaAttester.Attest(nonce)
```
susienme: Should this function be called only after `EnableReadyState()`? If so, consider using a flag.

agent/agent.go
Sibcgh: We can break off the first one

agent/agent.go
Sibcgh: That is a good idea, implemented the fix

agent/agent.go
Sibcgh: Added flag to avoid setting to ready state multiple times

susienme (approved):
LGTM

agent/agent.go
```diff
@@ -73,6 +73,9 @@ type AttestationAgent interface {
 	AttestWithClient(ctx context.Context, opts AttestAgentOpts, client verifier.Client) ([]byte, error)
 	AttestationEvidence(ctx context.Context, challenge []byte, extraData []byte, opts AttestAgentOpts) (*attestationpb.VmAttestation, error)
 	Refresh(context.Context) error
+	HasGPU() bool
+	EnableGPUReadyState() error
+	GetGPUAttestation(nonce []byte) (any, error)
```
yawangwang: `AttestationAgent` is a general, high-level interface representing the workload's attestation subsystem (managing TPM/TDX quotes, CEL event logs, and verifier tokens). A GPU is a specific, optional hardware accelerator. Baking GPU-specific methods into AttestationAgent couples the generic interface directly to NVIDIA hardware. If tomorrow another device is supported (e.g., TPU, SmartNIC, AMD GPU), will we add HasTPU(), EnableTPUReadyState(), and GetTPUAttestation()? That would bloat the interfaces especially when multiple devices are added.

There're two possible ways for better refactoring: 1. abstract these three specific GPU methods into one generic method, and replace "measureGPUAttestationEvidence" from launcher w/ it 2. Create a generic device manager for handling attestation collections from all attached devices, and inject that dependency into launcher. I've done some early improvements in https://github.com/google/go-tpm-tools/pull/718, which might be helpful for your reference.

yawangwang (changes requested):

agent/agent.go
Sibcgh: Thanks for the feedback and the reference in #718! I agree that `AttestationAgent` should remain strictly a hardware-agnostic platform attestation interface.

I went with **Option 2** (building on your `ROTManager` pattern from #718):
1. **100% Decoupled `AttestationAgent`**: Removed `HasGPU()`, `EnableGPUReadyState()`, and `GetGPUAttestation()` completely from `AttestationAgent`.
2. **Introduced `device.ROTManager` (`agent/device`)**: Manages attached hardware Root-of-Trust devices cleanly for both runtime attestation reports and boot-time event log measurement (`MeasureDeviceEvidence`).
3. **Dedicated `ReadyStateEnabler` Interface**: Kept `device.ROT` minimal (`Attest` + `Vendor`). Hardware that requires a runtime ready-state transition after measurement (like NVIDIA GPUs) additionally implements the optional `ReadyStateEnabler` interface (`EnableReadyState()`).

agent/device/device.go
```diff
@@ -0,0 +1,134 @@
...
+
+// NewROTManager creates a new ROTManager.
+func NewROTManager(rots []ROT) *ROTManager {
+	return &ROTManager{
+		rots: rots,
+	}
+}
+
+// AttestDeviceROTs fetches attestation reports from all detected device ROTs based on the provided options.
+func (m *ROTManager) AttestDeviceROTs(nonce []byte, opts ReportOpts) ([]any, error) {
```
susienme: Since opts is currently only used to access EnableRuntimeGPUAttestation, would it be simpler to just pass the boolean directly? Or is this struct expecting more fields in the future?

agent/device/device.go
```diff
@@ -0,0 +1,134 @@
...
+	m.deviceMu.Lock()
+	defer m.deviceMu.Unlock()
+
+	var deviceReports []any
+	var err error
+	for _, deviceROT := range m.rots {
+		if opts.EnableRuntimeGPUAttestation && deviceROT.Vendor() == NvidiaGPU {
```
susienme: Since `opts.EnableRuntimeGPUAttestation` is checked inside the loop, the loop will run unnecessarily if this flag is false. Consider checking it before the loop and returning `nil, nil` early to save meaningless iterations.

launcher/internal/gpu/attestation.go
```diff
@@ -30,21 +30,29 @@ var getGpuTypeInfo = deviceinfo.GetGPUTypeInfo
...
 // NvidiaAttester is responsible for collecting GPU attestation.
-type NvidiaAttester struct{}
+type NvidiaAttester struct {
+	readyStateEnabled       bool
+	gpuAttestationCollected bool
+}
 
 // NewNvidiaAttester returns a new NvidiaAttester if installGpuDriver is true, otherwise nil.
-func NewNvidiaAttester(installGpuDriver bool) Attester {
+func NewNvidiaAttester(installGpuDriver bool) device.ROT {
```
susienme: I understand the previous code returned an interface, but the standard Go best practice we follow is to return concrete types. Consider returning `*NvidiaAttester` instead. I will send a link via chat.

launcher/container_runner.go
```diff
@@ -564,33 +563,11 @@ func (r *ContainerRunner) measureGPUAttestationEvidence() error {
 		return fmt.Errorf("failed to generate random nonce: %v", err)
 	}
```
susienme: `Read fills b with cryptographically secure random bytes. It never returns an error, and always fills b entirely.` from https://pkg.go.dev/crypto/rand#Read

Error checking is unnecessary. I know this is now within this PR, but please remove the unnecessary check.

susienme (approved):

agent/device/device.go
Sibcgh: We anticipate having different options to pass through in the future for different accelerator types in the future so using an options struct prevents future API breaking changes.

agent/device/device.go
Sibcgh: Done! I added an early return `if !opts.EnableRuntimeGPUAttestation { return nil, nil }` before the loop. I also added a check right before it (`validateROTs()`) to return an error if multiple different accelerator types are attached in the same boot sequence, and added unit tests for this behavior.

agent/device/device.go
susienme: I see. If you see more devices with the same type, I think it might be better to keep your original code but use 
```
switch { 
  case <your original condition>: 
}
```
so that you can easily plug-in more cases.
For now, there is a single type, so it is up to you.

I mark this unresolved on to get your attention. So, please feel free to resolve this.

susienme (approved):

launcher/container_runner.go
```diff
@@ -552,45 +551,21 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
 // eventlog in the AttestationAgent.
-func (r *ContainerRunner) measureGPUAttestationEvidence() error {
-	if r.gpuAttester == nil {
+func (r *ContainerRunner) measureDeviceAttestationEvidence() error {
+	if r.deviceROTManager == nil {
 		return nil
 	}
 
 	nonce := make([]byte, 32)
-	if _, err := cryt.Read(nonce); err != nil {
```
yawangwang, line 375: why remove the error check?

agent/device/device.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (m *ROTManager) validateROTs() error {
+	if len(m.rots) <= 1 {
+		return nil
+	}
+	firstVendor := m.rots[0].Vendor()
+	for _, rot := range m.rots[1:] {
+		if rot.Vendor() != firstVendor {
+			return fmt.Errorf("multiple different device ROT vendors detected (%v and %v): only one accelerator type per boot sequence is supported", firstVendor, rot.Vendor())
```
yawangwang: why adding this check? Supporting different device ROTs is the goal of cs launcher.

agent/device/device.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	if !opts.EnableRuntimeGPUAttestation {
+		return nil, nil
+	}
+
+	var deviceReports []any
+	var err error
+	for _, deviceROT := range m.rots {
+		if deviceROT.Vendor() == NvidiaGPU {
```
yawangwang: why specifically checking device vendor here, gpu attester should already implement the `Attest` interface.

To make code more readable, i'd suggest filtering deviceROT based on the opts before flowing to the attestation collection.

agent/device/device.go
```diff
@@ -0,0 +1,161 @@
...
+// AttestDeviceROTs fetches attestation reports from all detected device ROTs based on the provided options.
+func (m *ROTManager) AttestDeviceROTs(nonce []byte, opts ReportOpts) ([]any, error) {
+	m.deviceMu.Lock()
+	defer m.deviceMu.Unlock()
+
+	if err := m.validateROTs(); err != nil {
+		return nil, err
+	}
+
+	if !opts.EnableRuntimeGPUAttestation {
```
yawangwang: let's avoid adding this check here - if we want to expand the opts to support TPU, this will return nil deviceRoT.

agent/device/device.go
```diff
@@ -0,0 +1,161 @@
...
+				deviceReports = append(deviceReports, deviceReport)
+			}
+		}
+	}
+	return deviceReports, err
+}
+
+// MeasureDeviceEvidence collects attestation evidence from all attached ROTs, measures them into
+// the event log via measurer, and transitions the devices to their ready states.
+func (m *ROTManager) MeasureDeviceEvidence(nonce []byte, measurer EventMeasurer) error {
```
yawangwang: Let's remove MeasureDeviceEvidence from agent/device/device.go and keep measureGpuEvidence inside launcher/container_runner.go. agent/device should be a generic package without dependencies on cel, gecel, or container launch workflows, while ContainerRunner should handle CEL event measurement during container startup.

launcher/container_runner.go
```diff
@@ -401,7 +400,7 @@ func NewRunner(ctx context.Context, cfg *RunnerConfig) (*ContainerRunner, error)
 		attestAgent,
 		logger,
 		workloadLogger,
-		nvidiaAttester,
+		device.NewROTManager(deviceROTs),
```
yawangwang: This will create two separate ROTManagers managing the same resource (nvidia attester), this creates mutex isolation. Could you pass single one deviceROT around?

agent/device/device.go
```diff
@@ -0,0 +1,161 @@
...
+	EnableReadyState() error
+}
+
+// EventMeasurer defines the interface for measuring events into an event log (CEL).
+type EventMeasurer interface {
+	MeasureEvent(event gecel.Content) error
+}
+
+// ROTManager manages the attestation of all attached device ROTs.
+type ROTManager struct {
```
yawangwang, line 34: it would be nicer to expose a `Lookup` function for ROTManager to return different types of ROT based on key (Vendor in this case).

launcher/container_runner.go
Sibcgh: Removed because checking `err != nil` on `crypto/rand.Read` was failing presubmit lint checks (in Go 1.24+, Read is guaranteed never to return an error, so the check is flagged as unreachable/dead code).

agent/device/device.go
Sibcgh: We should only allow one ROT vendor per container launch. While the codebase supports various vendors, this check ensures that all attached devices on a VM belong to the same vendor and prevents unsupported mixed-vendor configurations.

agent/device/device.go
Sibcgh: I've added a `filteredROTs(opts)` helper to filter the ROTs before iterating, so `AttestDeviceROTs` now calls `Attest()` without any vendor check in the loop.

agent/device/device.go
Sibcgh: Removed.

agent/device/device.go
Sibcgh: Refactored as you suggested. Now `agent/device` so it's now a generic package with zero CEL or proto dependencies, while `ContainerRunner` handles CEL event measurement and ready-state transitions during container startup.

launcher/container_runner.go
Sibcgh: Done

agent/device/device.go
Sibcgh: Done. I've updated `ROTManager` to store devices in a `map[Vendor][]ROT` and exposed a `Lookup(v Vendor)` method for O(1) time access by vendor key. Let me know if this is not effective.

susienme (approved):

agent/device/device.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+	return m
+}
+
+// ValidateROTs ensures that all attached devices belong to the same vendor.
+// While CS Launcher supports various ROT vendors, a single container launch only
+// supports one accelerator vendor type at a time (for example, multiple NVIDIA GPUs
+// are allowed, but mixing different vendor types is not supported).
+func (m *ROTManager) ValidateROTs() error {
```
yawangwang: CS launcher can support various vendor types at a time - there could be nvidia gpus, NIC devices from other vendors. We don't necessarily need strict checks against vendors.

agent/agent.go
```diff
@@ -233,7 +226,10 @@ func createBCAgent(principalFetcher principalIDTokenFetcher, sigsFetcher Signatu
 		return nil, fmt.Errorf("running in BC mode but TDX not supported")
 	}
 
-	baseAgent.avRot.AddDeviceROTs(deviceROTs)
+	if deviceROTManager == nil {
+		deviceROTManager = device.NewROTManager(nil)
+	}
```
yawangwang: is this check necessary? The implementations for deviceROTManager seem to be nil-safe.

launcher/container_runner.go
```diff
@@ -364,45 +365,63 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
 	return nil
 }
 
-// measureGPUAttestationEvidence will measure GPU attestation claims into the COS
+// measureDeviceAttestationEvidence will measure device attestation claims into the COS
 // eventlog in the AttestationAgent.
-func (r *ContainerRunner) measureGPUAttestationEvidence() error {
-	if r.gpuAttester == nil {
+func (r *ContainerRunner) measureDeviceAttestationEvidence() error {
```
yawangwang: let's keep the `measureGPUAttestationEvidence` because this func is coupled w/ nvidiaGPU.

launcher/container_runner.go
```diff
@@ -364,45 +365,63 @@ func (r *ContainerRunner) measureContainerClaims(ctx context.Context) error {
...
+
+		if enabler, ok := rot.(device.ReadyStateEnabler); ok {
+			if err := enabler.EnableReadyState(); err != nil {
+				return fmt.Errorf("failed to enable ready state for device %v: %w", rot.Vendor(), err)
+			}
+		}
 	}
-	r.logger.Info("Successfully measured GPU device attestation binding event and set GPU state to ready")
+
+	r.logger.Info("Successfully measured device attestation binding events and enabled ready states")
```
yawangwang, line 424: Could you expand the existing GPU driver installation test (see example https://github.com/google/go-tpm-tools/pull/730) to make sure gpu attestation is collected?

launcher/internal/gpu/attestation.go
```diff
@@ -30,21 +30,29 @@ var getGpuTypeInfo = deviceinfo.GetGPUTypeInfo
...
-	EnableReadyState() error
+	device.ROT
+	device.ReadyStateEnabler
 }
 
 // NvidiaAttester is responsible for collecting GPU attestation.
-type NvidiaAttester struct{}
+type NvidiaAttester struct {
+	readyStateEnabled       bool
+	gpuAttestationCollected bool
```
yawangwang: `NvidiaAttester` can be stateless as container runner will collect NV attestation anyway, measure it into CEL, and then enable it anyway.

yawangwang (approved):

launcher/container_runner.go
Sibcgh: Done, added a new check to see if attestation collection was written to console output.

PR #708, opened 2026-03-11
fix nil interface assertion

yawangwang:


launcher/internal/gpu/attestation.go
jkl73, line 53: Should do a nil check at beginning of all *NvidiaAttester functions

launcher/container_runner.go
```diff
@@ -64,6 +63,7 @@ type ContainerRunner struct {
 type gpuAttester interface {
 	Attest(nonce []byte) (any, error)
 	EnableReadyState() error
+	IsNil() bool
```
alexmwu: You can avoid this by having a NvidiaAttester constructor that just returns a non-typed nil. Arguably, the gpuAttester interface belongs in the GPU package anyway

yawangwang: Done, I was basically following the existing pattern for a quick fix, but your suggestion makes sense.

launcher/internal/gpu/attestation.go
yawangwang: Done.

PR #727, opened 2026-03-26
[KeyManager] Protoc to Buf

atulpatildbz:
use external protovalidate and remove vendored file


keymanager/gen_keymanager.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Re-generate the Go code for keymanager protos.
+
+cd "$(dirname "$0")" || exit 1
+
+go run github.com/bufbuild/buf/cmd/buf@latest generate . --template buf.gen.yaml
```
NilanjanDaw: Is this using the buf from the latest version? Should we pin it?

keymanager/buf.lock
```diff
@@ -0,0 +1,6 @@
+# Generated by buf. DO NOT EDIT.
```
NilanjanDaw, line 1: Move this file and the one below to KeyManager directory?

keymanager/km_common/proto/algorithms.pb.go
```diff
@@ -1,8 +1,8 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.36.11
-// 	protoc        v3.21.12
-// source: keymanager/km_common/proto/algorithms.proto
+// 	protoc        (unknown)
```
NilanjanDaw, line 4: These are showing up as unknown is that okay?

keymanager/gen_keymanager.sh
atulpatildbz: Good catch — pinned to buf@v1.68.2 in commit 675a16be5c9b.

keymanager/buf.lock
atulpatildbz: Done in 675a16be5c9b. Moved `buf.lock` and `buf.yaml` under keymanager/ and updated modules.path from keymanager to .. Re-generation produces byte-identical .pb.go files.

keymanager/km_common/proto/algorithms.pb.go
atulpatildbz: Yeah. Buf sets it to unknown: https://github.com/bufbuild/buf/blob/main/CHANGELOG.md#v100-rc9---2021-12-15

> Always set compiler_version parameter in the CodeGeneratorRequest to "(unknown)".

keymanager/gen_keymanager.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Re-generate the Go code for keymanager protos.
+
+cd "$(dirname "$0")" || exit 1
+
+go run github.com/bufbuild/buf/cmd/buf@v1.68.2 generate . --template buf.gen.yaml
```
NilanjanDaw, line 6: While this works, I am bit concerned about introducing a 3P build chain directly. Is it not possible to install buf and then run it? Instead of downloading it everytime?

NilanjanDaw (approved):
LGTM. Minor comment.

keymanager/gen_keymanager.sh
atulpatildbz: this is used only on the developer's machine for creating pb.go files on proto file edits, so we should be fine right?

Also, `go run ...@v1.68.2` populates the Go module and build caches on first invocation, so subsequent runs on the same machine are effectively local.

If we wire this up with a CI in future, we'd switch to https://github.com/bufbuild/buf-setup-action which pulls pre-built binary in github actions

PR #732, opened 2026-04-01
Add container cleanup service on stop

susienme:
Create a systemd service to send SIGTERM to running containerd tasks during shutdown/service stop to allow for graceful termination.


launcher/image/container-cleanup.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Container Cleanup on Service Stop
+Wants=containerd.service container-runner.service
+After=containerd.service container-runner.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/sh -c "echo 'Container cleanup service started'"
```
alexmwu: /bin/echo

launcher/image/container-cleanup.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Container Cleanup on Service Stop
+Wants=containerd.service container-runner.service
```
alexmwu: Should this be Requires=containerd.service container-runner.service

launcher/image/container-cleanup.service
alexmwu, line 1: Any reason we can't do this from the launcher? https://github.com/containerd/containerd/blob/v1.7.30/process.go#L33-L54

alexmwu (commented):
This needs image tests. See https://github.com/google/go-tpm-tools/blob/main/launcher/image/test/README.md.

launcher/image/container-cleanup.service
susienme: We decided to not rely on the launcher because "container process and the runner process are mostly detached, we want the container process to continue running even the runner process get terminated for whatever reasons." (from the internal bug)

In other words, we consider the case when the launcher is not running.

launcher/image/container-cleanup.service
susienme: Done.

launcher/image/container-cleanup.service
susienme: I think this should be `Wants` because we assume that the launcher could be stopped.

According to the `Requires=` section in `man systemd.unit`
> "this unit will be stopped (or restarted) if one of the other units is explicitly stopped (or restarted)."

susienme: We have changed the file to `launcher/image/power-button-listener.service` which has `After=` to configure ordering dependencies. `systemd` will stop the power-button-listener before it stops containerd and the container runner.

susienme:
re: the comment about the image test

During end-to-end testing, we found that the previous approach (solely using an ExecStop script) did not work in the hardened image because nothing was watching for power button press events. This is typically handled by `systemd-logind` (including in the debug image), but that service is not running in the hardened image. 

To address this issue, we introduced the power button listener service which identifies the correct event device, listens to it, and sends a power-off event to systemd so that it can propagate the shutdown. The listener service also receives this event from systemd and executes the ExecStop script, which in turn sends SIGTERM to all containers.

PR #743, opened 2026-04-14
[KeyManager] Introduce KPS operation mode

meetrajvala:
This PR introduces modes for calling Key Protection Service interfaces. It introduces following modes:

- KEY_PROTECTION_VM_EMULATED
- KEY_PROTECTION_VM

In KEY_PROTECTION_VM_EMULATED mode, the local Key Protection Service interfaces from KPS modules will be called directly. In KEY_PROTECTION_VM mode, the Key protection interfaces will be called over network (Implementation of these interfaces will be in follow-up CL). 

As currently we only KEY_PROTECTION_VM_EMULATED mode implemented in place, it is explicitly being passed while launching server, later this value will be derived from VM image and passed while launching WSD server.

Proto enum for these modes was already defined under `key_claims.proto` which is being moved to dedicated `modes.proto` file. PR also introduces `SERVICE_ROLE` enum which will be used for launching the same single KPS binary for different setups.

**Testing:** Existing unit tests passes.


keymanager/workload_service/server.go
```diff
@@ -184,8 +216,14 @@ var (
 )
 
 // New creates a new WSD Server listening on the given unix socket path.
-func New(_ context.Context, socketPath string) (*Server, error) {
-	return NewServer(&keyProtectionService{}, &workloadService{}, socketPath)
+func New(_ context.Context, socketPath string, mode KeyProtectionMode) (*Server, error) {
+	var kps KeyProtectionService
+	if mode == KeyProtectionVMEmulated {
```
atulpatildbz, line 221: should we have explicit switch statement? with default that returns and error.

That way if the mode is unknown, it doesn't default to remove service.

keymanager/workload_service/server.go
```diff
@@ -30,6 +30,16 @@ import (
 	wskcc "github.com/google/go-tpm-tools/keymanager/workload_service/key_custody_core"
 )
 
+// KeyProtectionMode defines the mode for the Key Protection Service implementation.
+type KeyProtectionMode string
+
+const (
+	// KeyProtectionVMEmulated uses the default local Key Protection Service implementation.
+	KeyProtectionVMEmulated KeyProtectionMode = "KEY_PROTECTION_VM_EMULATED"
```
atulpatildbz, line 38: this is fine, but do you think it's better to  use iota instead of string?

atulpatildbz (approved):
Minor comments.

keymanager/workload_service/server.go
```diff
@@ -30,6 +30,16 @@ import (
 	wskcc "github.com/google/go-tpm-tools/keymanager/workload_service/key_custody_core"
 )
 
+// KeyProtectionMode defines the mode for the Key Protection Service implementation.
+type KeyProtectionMode string
+
+const (
```
NilanjanDaw: This should be a proto so that it can be used across locations. We should also add the SERVICE_ROLE

keymanager/workload_service/server.go
atulpatildbz: Resolving this. @NilanjanDaw's comment https://github.com/google/go-tpm-tools/pull/743#discussion_r3101138501 is more relevant, we should move this to proto

keymanager/workload_service/server.go
meetrajvala: Updated it to use [KeyProtectionMechanism](https://github.com/google/go-tpm-tools/blob/main/keymanager/km_common/proto/key_claims.proto#L28) proto. As we will mainly have `KEY_PROTECTION_VM_EMULATED` and `KEY_PROTECTION`, so I am marking the `DEFAULT` for deprecation (as it is same as `KEY_PROTECTION_VM_EMULATED`). we will add `SERVICE_ROLE` support in follow-up PR.

keymanager/workload_service/server.go
meetrajvala: Using [KeyProtectionMechanism](https://github.com/google/go-tpm-tools/blob/main/keymanager/km_common/proto/key_claims.proto#L28) proto, now it will fail if key protection mechanism is not specified or mode is unknown.

keymanager/km_common/proto/key_claims.proto
```diff
@@ -28,30 +28,31 @@ message GetKeyClaimsRequest {
...
+  // The key is held by the Key Protection Services VM, and is endorsed by a
+  // pair of CVM attestations, from the Workload Services Daemon and Key
+  // Protection VM.
   KEY_PROTECTION_VM = 2;
 
-  // An interim solution where the key is held by the Workload Services Daemon emulating
-  // the Key Protection Service. It is endorsed by a single CVM attestation.
+  // An interim solution where the key is held by the Workload Services Daemon
+  // emulating the Key Protection Service. It is endorsed by a single CVM
+  // attestation.
```
NilanjanDaw: I feel `key_claims.proto` is not the right location to define this since its not just key_claims that is dependent on this. @atulpatildbz what do you think?

keymanager/workload_service/server.go
NilanjanDaw: `DEFAULT` mode will be used in future, it is not the same as `KEY_PROTECTION_VM_EMULATED`. Please don't deprecate it.

keymanager/km_common/proto/key_claims.proto
NilanjanDaw: I think we need a keymanager proto or something like that to handle protobufs concerning the entire system.

meetrajvala: if we plan to have more such system wide common stuff then let's define keymanager.proto else we can have mdoes.proto capturing these mechanisms and service roles.

keymanager/workload_service/server.go
meetrajvala: Ok, removing the deprecation tag. For now, we will not handle this mode in KPS initialization. Will address it as part of future KPS implementation.

keymanager/workload_service/server.go
```diff
@@ -30,6 +30,9 @@ import (
 	wskcc "github.com/google/go-tpm-tools/keymanager/workload_service/key_custody_core"
 )
 
+// KeyProtectionMode defines the mode for the Key Protection Service implementation.
+type KeyProtectionMode string
+
```
atulpatildbz: dead code?

keymanager/km_common/proto/key_claims.proto
atulpatildbz: I see we decided to go with modes.proto. i'm okay with this

keymanager/workload_service/server.go
meetrajvala: removed it.

keymanager/km_common/proto/key_claims.proto
```diff
@@ -25,33 +25,16 @@ message GetKeyClaimsRequest {
   KeyType key_type = 2;
 }
 
-enum KeyProtectionMechanism {
```
NilanjanDaw, line 28: Are there any side-effects to removing this from the key_claims.proto? Are we breaking any already existing APIs?

meetrajvala: There are references in `GenerateKeyResponse` api but there is no side-effect or breaking change for that because it is still under same package (under different proto file).

NilanjanDaw (approved):

PR #756, opened 2026-04-23
Implement graceful shutdown on power button event.

susienme:
This change enables the launcher to monitor for hardware power button events (ACPI shutdown) and propagate a SIGTERM signal to the container workload for a graceful shutdown.

- Added a Linux input event listener to detect power button presses.
- Integrated the listener into the launcher.
- Added automated tests including a test workload to verify that graceful shutdown triggers the expected cleanup messages in logs.

### DIFF
--- a/launcher/cloudbuild.yaml
+++ b/launcher/cloudbuild.yaml
@@ -520,7 +520,21 @@ steps:
     exit
 
 - name: 'gcr.io/cloud-builders/gcloud'
-  id: WorkloadCleanupTests
+  id: DebugWorkloadCleanupTests
+  waitFor: ['DebugImageBuild']
+  env:
+  - 'OUTPUT_IMAGE_NAME=${_OUTPUT_IMAGE_PREFIX}-debug-${_OUTPUT_IMAGE_SUFFIX}'
+  - 'PROJECT_ID=$PROJECT_ID'
+  script: |
+    #!/usr/bin/env bash
+    cd launcher/image/test
+    echo "running debug workload cleanup tests on ${OUTPUT_IMAGE_NAME}"
+    gcloud builds submit --config=test_workloadcleanup_cloudbuild.yaml --region us-west1 \
+      --substitutions _IMAGE_NAME=${OUTPUT_IMAGE_NAME},_IMAGE_PROJECT=${PROJECT_ID}
+    exit
+
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: HardenedWorkloadCleanupTests
   waitFor: ['HardenedImageBuild']
   env:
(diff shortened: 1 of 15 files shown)


launcher/container_runner.go
```diff
@@ -770,6 +771,23 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 	}
 	defer task.Delete(ctx)
 
+	// A go-routine will monitor the power button and send a SIGTERM to the container uppon a button press.
+	if pwr, err := newPowerButtonListener(r.logger); err != nil {
+		r.logger.Error(err.Error())
+	} else {
+		go func() {
```
jkl73: we might want to make sure the goroutine not getting leaked, like when the task exit normally.

(it's not huge issue in this case, but I think it's good practice)

launcher/cloudbuild.yaml
```diff
@@ -519,6 +519,20 @@ steps:
...
+  id: WorkloadCleanupTests
+  waitFor: ['HardenedImageBuild']
+  env:
+  - 'OUTPUT_IMAGE_NAME=${_OUTPUT_IMAGE_PREFIX}-hardened-${_OUTPUT_IMAGE_SUFFIX}'
+  - 'PROJECT_ID=$PROJECT_ID'
+  script: |
+    #!/usr/bin/env bash
+    cd launcher/image/test
+    echo "running workload cleanup tests on ${OUTPUT_IMAGE_NAME}"
+    gcloud builds submit --config=test_workloadcleanup_cloudbuild.yaml --region us-west1 \
```
jkl73, line 546: can we run debug image as well? want to make sure they have the same behavior

launcher/image/logind_override.conf
```diff
@@ -0,0 +1,2 @@
+[Login]
+HandlePowerKey=ignore
```
jkl73: hmm, why is this needed here? I thought for debug image, we will just let the logind do the job?

launcher/cloudbuild.yaml
susienme: Added `DebugWorkloadCleanupTests` and renamed previous one to `HardenedWorkloadCleanupTests`

launcher/container_runner.go
susienme: Done in commit [6633b8e](https://github.com/google/go-tpm-tools/pull/756/commits/6633b8e27eb6ee7d2d4a08da503d21a45594bc8b). When `ContainerRunner.Close()` is called, the Go-routine will stop.

launcher/image/logind_override.conf
susienme: Removed. Now the launcher relays SIGTERM from `logind` to the main workload in [02d1f83](https://github.com/google/go-tpm-tools/pull/756/commits/02d1f8313353ced1dc23ac66330c7950b7ef33ca)

jkl73 (approved):

PR #762, opened 2026-04-27
[KeyManager] - WSD API migration to gRPC

atharvasaraf123:
Migrates the Key Manager Daemon (WSD) from a manual REST-only HTTP multiplexer to a fully typed, high-performance Dual-Socket Architecture supporting both gRPC and REST JSON clients simultaneously using grpc-gateway.

Changes Made:
1) Dual Unix Domain Socket Support: The daemon now securely listens on two distinct endpoints (-grpc.sock and standard .sock), allowing native gRPC traffic to process without overhead while routing conventional REST requests via a reverse proxy.
2) Backward-Compatible JSON: Configured runtime.JSONPb with UseProtoNames: true to guarantee pristine snake_case key serialization matching legacy API outputs.
3) Robust Error & Validation Handling: Upgraded endpoint errors to status.Errorf to correctly map bad requests to HTTP 400/404 REST codes.

New Autogenerated files:
1) api_grpc.pb.go - Provided with RegisterWorkloadServiceServer
2) api.pb.gw.go - Provided with RegisterWorkloadServiceHandler

Testing:
- Added UTs for GRPC apis
- All existing UTs along with workload passes


keymanager/workload_service/server_test.go
```diff
@@ -391,12 +391,13 @@ func TestHandleGenerateKeyBadRequest(t *testing.T) {
...
 
 			if tc.name == "unsupported algorithm" {
-				var resp map[string]string
+				var resp map[string]interface{}
 				if err := json.Unmarshal(w.Body.Bytes(), &resp); err != nil {
 					t.Fatalf("failed to decode response: %v", err)
 				}
 				expectedSubstr := "Supported algorithms: KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256"
-				if errMsg, ok := resp["error"]; !ok || !strings.Contains(errMsg, expectedSubstr) {
+				errMsg, _ := resp["message"].(string)
```
atulpatildbz: the error format has been changed. 

old:
```json
{"error": "KEM key handle not found: 6e2..."}
```

new:
```json
{"code": 5, "message": "KEM key handle not found: 6e2...", "details": []}
```

this will be a breaking change for customers that rely on `response.error`

Check if we can use `runtime.WithErrorHandler(...)` to avoid this problem

keymanager/workload_service/server_test.go
```diff
@@ -687,8 +725,8 @@ func TestHandleEnumerateKeysMethodNotAllowed(t *testing.T) {
 	w := httptest.NewRecorder()
 	srv.Handler().ServeHTTP(w, req)
 
-	if w.Code != http.StatusMethodNotAllowed {
-		t.Fatalf("expected status 405, got %d", w.Code)
+	if w.Code != http.StatusNotImplemented {
+		t.Fatalf("expected status 501, got %d", w.Code)
```
atulpatildbz: grpc-gateway has documented this behavior. but also provided a way to retain 405. can you check if this is feasible? https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_your_gateway/#customizing-routing-errors

keymanager/workload_service/server.go
```diff
@@ -305,104 +356,88 @@ func decapsAADContext(kemUUID uuid.UUID, algorithm keymanager.KemAlgorithm) []by
 	return []byte(fmt.Sprintf("wsd:keys:decaps:v1:%d:%s", algorithm, kemUUID))
 }
 
-func (s *Server) handleDecaps(w http.ResponseWriter, r *http.Request) {
-	var req api.DecapsRequest
-	if err := readRequest(r, &req); err != nil {
```
atulpatildbz, line 310: since we've now removed readRequest call, we're no longer doing `protovalidate.Validate`

Can you add `protovalidate.Validate` to all the endpoints?

keymanager/workload_service/server.go
atulpatildbz: check if we can remove this method now. seems like the only thing we need from here is `protovalidate.Validate` which we can either add independently in every endpoint or have a separate helper for it.

keymanager/workload_service/server.go
```diff
@@ -230,21 +244,58 @@ func NewServer(keyProtectionService KeyProtectionService, workloadService Worklo
...
+
+	mux := runtime.NewServeMux(
+		runtime.WithForwardResponseOption(customResponseModifier),
+		runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{
+			MarshalOptions: protojson.MarshalOptions{
+				UseProtoNames:   true,
+				EmitUnpopulated: true,
+			},
+		}),
+	)
```
atulpatildbz, line 337: we lose the `X-Content-Type-Options: nosniff` header that was previously set in `writeJSON` and `writeError`.
While this is minor for non-browser clients, we should consider wrapping the `mux` here to preserve this defense-in-depth header.
 
 Example fix:
 ```go
 handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
     w.Header().Set("X-Content-Type-Options", "nosniff")
     mux.ServeHTTP(w, r)
 })
 s.httpServer = &http.Server{Handler: handler}
 ```

keymanager/workload_service/server.go
```diff
@@ -305,104 +356,88 @@ func decapsAADContext(kemUUID uuid.UUID, algorithm keymanager.KemAlgorithm) []by
...
-	var req api.GenerateKeyRequest
-	if err := readRequest(r, &req); err != nil {
-		writeError(w, err.Error(), http.StatusBadRequest)
-		return
+func (s *Server) GenerateKey(_ context.Context, req *api.GenerateKeyRequest) (*api.GenerateKeyResponse, error) {
+	if req == nil || req.Algorithm == nil || req.Algorithm.Params == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "algorithm details cannot be empty")
+	}
+	if err := protovalidate.Validate(req); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid request structure: %v", err)
```
atulpatildbz: nit: older string used in `readRequest` was `invalid request: ...`

keymanager/workload_service/integration_test.go
```diff
@@ -165,10 +162,10 @@ func TestIntegrationDestroyKey(t *testing.T) {
 	}
 
 	var respGen api.GenerateKeyResponse
-	if err := json.NewDecoder(wGen.Body).Decode(&respGen); err != nil {
+	if err := protojson.Unmarshal(wGen.Body.Bytes(), &respGen); err != nil {
```
NilanjanDaw: Any reason we are moving from `json.NewDecoder` to `protojson.Unmarshal`?

keymanager/workload_service/server.go
```diff
@@ -221,6 +227,14 @@ func New(_ context.Context, socketPath string, mode keymanager.KeyProtectionMech
 	return NewServer(kps, &workloadService{}, socketPath)
 }
 
+func customResponseModifier(ctx context.Context, w http.ResponseWriter, resp proto.Message) error {
```
NilanjanDaw: this is on the critical path for all requests, is there a more elegant way to handle this?

keymanager/workload_service/server.go
```diff
@@ -230,21 +244,58 @@ func NewServer(keyProtectionService KeyProtectionService, workloadService Worklo
 		mu:                   sync.RWMutex{},
 		claimsChan:           make(chan *ClaimsCall, 4),
 	}
+	if socketPath == "" || !filepath.IsAbs(socketPath) {
+		socketPath = filepath.Join("/tmp", "wsd_keys_"+uuid.NewString()+".sock")
```
NilanjanDaw: the socketpath name should not be randomly generated. It needs to be `kmaserver.sock` and `kmaserver-grpc.sock`

keymanager/workload_service/server.go
```diff
@@ -230,21 +244,58 @@ func NewServer(keyProtectionService KeyProtectionService, workloadService Worklo
...
-	mux.HandleFunc("POST /v1/keys:destroy", s.handleDestroy)
-	s.httpServer = &http.Server{Handler: mux}
+	// 1. Initialize Native gRPC Server
+	_ = os.Remove(grpcSocketPath)
+	grpcListener, err := net.Listen("unix", grpcSocketPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to listen on gRPC socket %s: %w", grpcSocketPath, err)
+	}
+
+	grpcServer := grpc.NewServer()
```
atulpatildbz, line 302: you're not using the `grpcServer` variable defined on line 201. instead this creates a local variable. was this intentional?

keymanager/workload_service/server.go
NilanjanDaw: Additionally, if the socketpath is empty we should error out rather than falling back silently.

keymanager/workload_service/server.go
atulpatildbz, line 373: you need to shutdown grpcServer as well

```suggestion
    if s.grpcServer != nil {
        s.grpcServer.GracefulStop()
    }
    return s.httpServer.Shutdown(ctx)
```

You might also want to close the `grpcListener` if `GracefulStop` doesn't do it automatically for Unix sockets, or ensure it's handled

keymanager/workload_service/server.go
```diff
@@ -230,21 +244,58 @@ func NewServer(keyProtectionService KeyProtectionService, workloadService Worklo
...
+	api.RegisterWorkloadServiceServer(grpcServer, s)
 
-	_ = os.Remove(socketPath)
-	ln, err := net.Listen("unix", socketPath)
+	go func() {
+		if err := grpcServer.Serve(grpcListener); err != nil {
```
NilanjanDaw, line 306: You might want to provide a `Stop()` or `Close()` method on the Server struct that calls `s.grpcServer.GracefulStop()` and `s.httpServer.Shutdown()`. Also I would suggest modularizing this into a separate function.

keymanager/workload_service/server_test.go
atulpatildbz: let's use `srv.Shutdown(context.Background())` so that it shuts down both REST and gRPC.

assuming that you fix the grpc shutdown in server.go

keymanager/workload_service/server.go
```diff
@@ -230,21 +244,58 @@ func NewServer(keyProtectionService KeyProtectionService, workloadService Worklo
...
+			log.Printf("native gRPC server stopped: %v", err)
+		}
+	}()
+
+	// 2. Initialize gRPC-Gateway REST proxy client
+	conn, err := grpc.Dial(
```
NilanjanDaw: `grpc.Dial` is deprecated in favor of `grpc.NewClient`, please consider moving to the newer version.

keymanager/go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	cel.dev/expr v0.25.1 // indirect
 	github.com/antlr4-go/antlr/v4 v4.13.1 // indirect
 	github.com/google/cel-go v0.27.0 // indirect
+	github.com/grpc-ecosystem/grpc-gateway/v2 v2.26.3 // indirect
```
atulpatildbz: grpc isn't liste here. can you run `go mod tidy` once?

keymanager/workload_service/server_test.go
```diff
@@ -391,12 +391,13 @@ func TestHandleGenerateKeyBadRequest(t *testing.T) {
 			}
 
 			if tc.name == "unsupported algorithm" {
-				var resp map[string]string
+				var resp map[string]interface{}
 				if err := json.Unmarshal(w.Body.Bytes(), &resp); err != nil {
 					t.Fatalf("failed to decode response: %v", err)
 				}
 				expectedSubstr := "Supported algorithms: KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256"
-				if errMsg, ok := resp["error"]; !ok || !strings.Contains(errMsg, expectedSubstr) {
```
NilanjanDaw, line 399: Why are we changing the "error" param to "message"?

keymanager/workload_service/server.go
atulpatildbz: `grpc.Dial` seems to be deprecated: https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#the-wrong-way-grpcdial

can we use `grpc.NewClient`?

keymanager/workload_service/server_test.go
```diff
@@ -624,57 +625,94 @@ func TestHandleEnumerateKeysWithKeys(t *testing.T) {
...
+		if info1.PubKey.Algorithm.GetParams().GetKemId() != keymanager.KemAlgorithm_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256 {
+			t.Fatalf("expected algorithm %v, got %v", keymanager.KemAlgorithm_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256, info1.PubKey.Algorithm.GetParams().GetKemId())
+		}
+		if !bytes.Equal(info1.PubKey.PublicKey, kemPubKey1) {
+			t.Fatalf("KEM pub key mismatch for kem1")
+		}
+		if info1.KeyProtectionMechanism != keymanager.KeyProtectionMechanism_KEY_PROTECTION_VM_EMULATED.String() {
+			t.Fatalf("expected key protection mechanism %s, got %s", keymanager.KeyProtectionMechanism_KEY_PROTECTION_VM_EMULATED.String(), info1.KeyProtectionMechanism)
+		}
+		// Approximate check for expiration time
```
NilanjanDaw: The logic here, and for the grpc test case is significantly identical. Either move the parsing logic to a separate util function or make the test parameterized.

keymanager/workload_service/server_test.go
```diff
@@ -753,34 +791,56 @@ func TestToHpkeAlgorithm(t *testing.T) {
...
-	}
+		if len(resp.SupportedAlgorithms) != 1 ||
+			resp.SupportedAlgorithms[0].Algorithm.GetParams().GetKemId() != keymanager.KemAlgorithm_KEM_ALGORITHM_DHKEM_X25519_HKDF_SHA256 ||
+			resp.SupportedAlgorithms[0].Algorithm.Type != "kem" {
+			t.Errorf("unexpected supported algorithms via REST: %v", resp.SupportedAlgorithms)
+		}
+	})
+
+	t.Run("GRPC", func(t *testing.T) {
+		baseAddr := srv.listener.Addr().String()
```
NilanjanDaw, line 798: same comment as above.

keymanager/workload_service/server_test.go
```diff
@@ -1100,7 +1168,25 @@ func TestHandleDestroy(t *testing.T) {
 				req.Header.Set("Content-Type", "application/json")
 			}
 			w := httptest.NewRecorder()
-			srv.Handler().ServeHTTP(w, req)
+			if tc.name == "success_via_grpc" {
```
atulpatildbz: Table-driven tests are great when the execution flow is identical for all cases. When one case requires a completely different setup (gRPC dial vs. HTTP request), it often makes more sense to pull it out into a separate test or subtest (like `TestHandleGetCapabilities` does)

atulpatildbz (changes requested):

keymanager/workload_service/server_test.go
NilanjanDaw: what's this magic string?

keymanager/workload_service/server.go
atharvasaraf123: Initialised the server's grpc server to this at the end of this function.

keymanager/workload_service/integration_test.go
atharvasaraf123: Moved back to json.NewDecoder.

keymanager/workload_service/server.go
atharvasaraf123: Could not find anything apart from this online.

PR #765, opened 2026-04-28
Update ci.yml to show line number for failure in go linter job

atharvasaraf123:
Updated go linter job to show line number where linter failed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,6 +152,7 @@ jobs:
             -E revive
             -E gofmt
             -E goimports
+            --out-format=colored-line-number
             --exclude-use-default=false
             --max-same-issues=0
             --max-issues-per-linter=0


.github/workflows/ci.yml
```diff
@@ -152,6 +152,7 @@ jobs:
             -E revive
             -E gofmt
             -E goimports
```
atulpatildbz, line 154: existing issue: goimports is listed twice, can you remove that in current PR as well?

.github/workflows/ci.yml
```diff
@@ -152,6 +152,7 @@ jobs:
...
+            --out-format=colored-line-number
```
atulpatildbz, line 155: we're using `golangci-lint-action@v3.2.0`. 

and in this version i see the following check:
https://github.com/golangci/golangci-lint-action/blob/v3.2.0/src/run.ts#L130-L133

```ts
  if (userArgNames.has(`out-format`)) {
    throw new Error(`please, don't change out-format for golangci-lint: it can be broken in a future`)
  }
  addedArgs.push(`--out-format=github-actions`)
```

can you check and confirm?

If this is really needed, we can upgrade golangci-lint-action in the new key-protection-module repository.

atharvasaraf123: Sure, will close this review.

NilanjanDaw: Seems like this will be broken in the future, maybe not a good idea to add in the new repo either?

PR #772, opened 2026-05-04
Support running unprivileged workload container.

susienme:
Implement support for unprivileged rootless workload execution inside the launcher using user namespaces (UID/GID shifting), isolated network namespaces with CNI, and clean stdout/stderr redirection.

Key changes:
- User Namespace Support: Map container root (UID 0) to host unprivileged UID (10000) via `containerd.WithRemappedSnapshot` and OCI spec user mappings.
- Namespace Isolation & CNI: Create an isolated network namespace owned by the new user namespace, bypassing sysfs privilege restrictions and enabling CNI config loading and port forwarding natively via go-cni.
- I/O redirection bypass: Bind-mount a host FIFO directly over /dev/stdout and /dev/stderr to allow non-root workloads to open them without crossing user namespace boundaries.
- Socket permissions: Update socket permissions to allow unprivileged processes to query the socket.
- Integration Tests: Updated 12 integration test configurations to assert proper environment behavior, capabilities, ODA verification, and network features under rootless container environments.
- Dependencies: Registered github.com/containerd/go-cni along with minimal direct dependencies in go.mod and go.sum.


launcher/image/10-workload.conf
```diff
@@ -0,0 +1,15 @@
...
+  "bridge": "br0",
+  "isGateway": true,
+  "ipMasq": true,
+  "ipam": {
+    "type": "host-local",
+    "subnet": "172.20.0.0/24",
+    "routes": [
+      { "dst": "0.0.0.0/0" }
+    ]
+  }
```
alexmwu: what about IPv6

launcher/image/10-workload.conf
```diff
@@ -0,0 +1,15 @@
+{
```
jkl73, line 1: looks like majority of this is change is to add cni, I'm assuming it's because the unprivileged container process cannot bind to the lower ports? If that's the only concern maybe we can accept that if it's a unprivileged workload you have to bind to a higher ports)

susienme: This CNI configuration is providing a bridge between the host and the container, which is separate from the lower port binding itself. Lower port binding is enabled by the introduction of the network namespace. Suggesting that we run unprivileged workloads on higher ports is essentially suggesting to disable the network namespace entirely and run on the host network stack.

While requiring unprivileged workloads to bind to higher ports ($>1024$) is a standard practice in some environments, we opted to implement isolated network namespaces and CNI for two primary reasons:

1. **Backward Compatibility:** Our primary objective is to enable non-root container execution while maintaining compatibility, thereby avoiding the need for customer modifications to existing workloads in most scenarios. For instance, many standard workloads (like Nginx) run as UID 0 within the container and attempt to bind to port 80 or 443. If we disabled the network namespace and forced workloads to bind to higher host ports, customers would have to modify their container configurations, entrypoints, and Dockerfiles, breaking our goal of a seamless out-of-the-box migration.

2. **Bypassing Sysfs Privilege Restrictions:** Inside a user-namespace setup, creating an isolated network namespace owned by the new user namespace is a strict kernel requirement to bypass sysfs privilege restrictions. This is necessary to ensure the container runtime and utilities can mount and access `/sys` cleanly.

For the same reasons, Docker only enables and disables the user and network namespace at the same time.

By leveraging network namespaces and CNI, the bridge connects the host interface to the container interface, enabling privileged port access securely and cleanly without any workload modifications.

launcher/container_runner.go
```diff
@@ -809,14 +843,40 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get image from container: %w", err)
+	}
+	imageConfig, err := getImageConfig(ctx, image)
+	if err != nil {
+		return fmt.Errorf("failed to get image config: %w", err)
+	}
+	var containerIP string
+	if r.launchPolicy.NonrootContainer {
+		containerIP, err = r.getContainerIP(ctx, fmt.Sprintf(netnsPathFmt, task.Pid()))
```
jkl73: looks like container ip is defined and hardcoded in the conf, can we just use the coded value?

launcher/spec/launch_policy.go
```diff
@@ -233,6 +235,11 @@ func GetLaunchPolicy(imageLabels map[string]string, logger logging.Logger) (Laun
 		}
 	}
 
+	if v, ok := imageLabels[nonrootContainer]; ok {
+		if launchPolicy.NonrootContainer, err = strconv.ParseBool(v); err != nil {
```
jkl73, line 243: should we make sure this option is mutually exclusive to "privilegedCaps"

launcher/container_runner.go
```diff
@@ -809,14 +843,40 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+	}
+	if err := openPorts(imageConfig.ExposedPorts, containerIP); err != nil {
```
jkl73, line 723: Ideally if we can know container IP from the config, we don't need to move openPorts to here (and we don't need to pass in launchPolicy to containerrunner)

launcher/container_runner.go
```diff
@@ -963,6 +1025,33 @@ func openPorts(ports map[string]struct{}) error {
...
+			if err != nil {
+				return fmt.Errorf("failed to forward port %s to container %s: %v %s", port, containerIP, err, out)
+			}
+
+			// Allow traffic in FORWARD chain to the container IP on this port
+			forwardInCmd := exec.Command("iptables", "-A", "FORWARD", "-d", containerIP, "-p", protocol, "--dport", port, "-j", "ACCEPT")
+			if out, err := forwardInCmd.CombinedOutput(); err != nil {
+				return fmt.Errorf("failed to add FORWARD rule for container %s: %v %s", containerIP, err, out)
+			}
+
```
jkl73: nit: extra line

launcher/image/10-workload.conf
jkl73: thanks ack

launcher/image/10-workload.conf
```diff
@@ -0,0 +1,15 @@
...
+  "cniVersion": "0.4.0",
+  "name": "workload",
+  "type": "bridge",
+  "bridge": "br0",
+  "isGateway": true,
+  "ipMasq": true,
+  "ipam": {
+    "type": "host-local",
```
jkl73, line 9: can we use static? so we can define the ip address

launcher/container_runner.go
susienme: Currently, the configuration defines a subnet, not a static IP address. While it is true that the current CNI reliably assigns the same starting IP on a clean boot, how future CNI plugins or updates will allocate IPs is not guaranteed.

Furthermore, obtaining the IP address is completely free because it is just a byproduct of executing CNI setup. To make this clearer in the code, I have renamed the function to setupCNI.

launcher/spec/launch_policy.go
susienme: I think they should **NOT** be mutually exclusive. For example, if a workload wants to create a VPN connection or virtual interface inside the container, it needs `CAP_NET_ADMIN`. Similarly, mounting `tmpfs` requires CAP_SYS_ADMIN.

Indeed, I think allowing `NonrootContainer=true` alongside capabilities is actually ideal for defense-in-depth. Because the container runs in unprivileged user and network namespaces, the Linux kernel scopes these capabilities strictly inside the container's sandbox without granting any power over the host VM's physical network or host disks.

launcher/container_runner.go
susienme: I think we should dynamically obtain the IP address, as it is a byproduct of setting up CNI and could change in the future if theCNI plugin is updated (as discussed in the thread above). 

Additionally, we still need `launchPolicy` in this function to decide whether to configure the user and network namespaces for `taskOpts` (line: 847), unless we pass those namespace options in as arguments to `Run()`.

launcher/container_runner.go
susienme: Deleted.

launcher/image/10-workload.conf
susienme: I think it is technically feasible as the `static` IPAM plugin is present.

However, I think the code should not rely on a hardcoded IP address in the configuration to enforce loose coupling and separation of concerns. The CNI configuration should remain the single source of truth for the network layout (subnets and routing), while the launcher dynamically obtains the assigned IP. This ensures that any future changes to the network layout (such as shifting subnets) do not break the launcher or require coordinated modifications across multiple files.

WDYT?

launcher/container_runner.go
```diff
@@ -48,13 +49,15 @@ import (
 type ContainerRunner struct {
 	container      containerd.Container
 	launchSpec     spec.LaunchSpec
+	launchPolicy   spec.LaunchPolicy
```
jkl73: launcherPolicy doesn't need to be added into the struct.

Instead, can we create a new field in launchSpec (nonroot), the value can be derived from the launchpolicy

so we don't need to update all those function signatures

susienme: I agree that doing so would reduce the number of fields on ContainerRunner. However, the comment on LaunchSpec notes, _"LaunchSpec contains specification set by the **operator**..."_.

I think the purpose is to maintain a clear boundary between the operator and the workload author. Introducing a field into LaunchSpec that is set by the workload author blurs this boundary.

If the concern is the number of fields on ContainerRunner, we could group related fields into a nested struct. If the concern is memory overhead from keeping LaunchPolicy, the struct is small.

launcher/image/10-workload.conf
jkl73: ack

launcher/container_runner.go
jkl73: Ya I agree LaunchSpec should be specified by the operator, my thought is the container image is chosen by the operator so it's technically still operator controlled...

But of course, we can always make unprivileged container a flag the operator can control.

susienme: I see your point about the operator selecting the image reference, but I'm still not sure about the suggestion. Are you recommending that we add `NonrootContainer`, which is set by the author, to `LaunchSpec`, whose docstring explicitly defines it as _**'... specification set by the operator ...'**_?

I can do that if you strongly prefer it, but it breaks the convention and blurs the separation between operator metadata and image author policy. That's why keeping it in `LaunchPolicy` feels cleaner.

susienme: Notes from a meeting:

We are going to introduce both a launch policy and an MDS flag to maintain a clear boundary between the workload author and the operator.

1. Launch Policy (Workload Author)
- Default to unrestricted (unspecified): The operator can decide between a root or non-root container.
- Restricted: At the moment, an author can restrict this to force a non-root container.

2. MDS Flag (Operator)
- Default to root (current): The operator can choose to run as root or non-root, only when the launch policy does not restrict it.
- Non-root: Explicitly requests a non-root container.
- Note: If the launch policy is restricted to non-root and the operator fails to set the MDS flag to non-root (leaving it as the root default), the launcher will return an error because the MDS flag does not comply with the author's policy.

Here is a 2x2 matrix of how these two configurations will interact:
| Launch Policy | MDS Flag | Final Outcome | Reason |
| --- | --- | --- | --- |
| Unrestricted (Default) | Root (Default) | Runs as Root | The author allows either, and the operator either explicitly chose root or fell back to the default. |
| Unrestricted (Default) | Non-root | Runs as Non-root | The author allows either, and the operator explicitly opted into running as non-root. |
| Restricted (Non-root only) | Root (Default) | Error | The operator failed to opt into the non-root container (or explicitly tried to run as root), which violates the author's strict policy. |
| Restricted (Non-root only) | Non-root | Runs as Non-root | Both the author's policy and the operator's flag align on requiring a non-root container. |

jkl73 (approved):
lgtm!

PR #774, opened 2026-05-06
[launcher] cache gpu driver installer files

jkl73:
cache the driver files so no need to download at run time.

From now on we need to manually update the base image and gpu driver version.

### DIFF
--- a/launcher/cloudbuild.yaml
+++ b/launcher/cloudbuild.yaml
@@ -1,5 +1,5 @@
 substitutions:
-  '_BASE_IMAGE': '' # If left empty, will use the latest image in _BASE_IMAGE_FAMILY of _BASE_IMAGE_PROJECT
+  '_BASE_IMAGE': 'cos-125-19216-220-180' # If left empty, will use the latest image in _BASE_IMAGE_FAMILY of _BASE_IMAGE_PROJECT
   '_BASE_IMAGE_FAMILY': 'cos-125-lts'
   '_BASE_IMAGE_PROJECT': 'cos-cloud'
   '_OUTPUT_IMAGE_PREFIX': 'confidential-space'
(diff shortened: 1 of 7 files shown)


launcher/cloudbuild.yaml
```diff
@@ -80,9 +80,18 @@ steps:
           'gs://confidential-space-images_third-party/confidential_space_experiments',
           './launcher/image/confidential_space_experiments']
 
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: DownloadGpuDrivers
+  entrypoint: 'gcloud'
+  args: ['storage', 'cp',
+          'gs://cos-nvidia-gpu-drivers/NVIDIA-Linux-x86_64-595.58.03.run',
+          'gs://cos-infra-prod-artifacts-official/lakitu-release/R125-19216.220.180/gpu_driver_versions.bin',
```
rkolchmeyer: can you use the released artifacts here?

gs://cos-tools/19216.220.180/lakitu/gpu_driver_versions.bin
gs://cos-tools/19216.220.180/lakitu/nvidia-drivers-595.58.03.tgz

rkolchmeyer (approved):

launcher/cloudbuild.yaml
jkl73: replaced

alexmwu (approved):

yawangwang (approved):

launcher/cloudbuild.yaml
```diff
@@ -84,9 +80,18 @@ steps:
           'gs://confidential-space-images_third-party/confidential_space_experiments',
           './launcher/image/confidential_space_experiments']
 
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: DownloadGpuDrivers
+  entrypoint: 'gcloud'
+  args: ['storage', 'cp',
+          'gs://cos-nvidia-gpu-drivers/NVIDIA-Linux-x86_64-595.58.03.run',
```
meetrajvala, line 87: can you add a comment here to ensure this GCS bucket path remains consistent with the base image version so that drivers are sourced from the appropriate build artifacts ?

meetrajvala (approved):

launcher/cloudbuild.yaml
jkl73: will add in a follow up pr

PR #775, opened 2026-05-06
Add SHA384 support for uki image EFI Authenticode hash

yanchengchen7:
The new `compute_efi_authenticode_hash` function in measure.sh, computes the PE/COFF Authenticode hash by parsing the EFI binary directly in Python, mirroring sbsigntools' image.c. It
 - walks the DOS / PE / optional headers,
 - skips the CheckSum field and the Certificate Table data-directory entry,
 - hashes sections in section-table order,
 - appends trailing endjunk and zero-extends to an 8-byte alignment,

The image type (`default` vs `uki`) is now auto-detected by inspecting the boot EFI binary's PE section table (parsed with the stdlib `struct` module). Detection requires **positive evidence on both sides**:
 - `.sbat` present **and** containing `https://github.com/rhboot/shim` → `default` (shim + grub + kernel layout)
 - `.setup` present (x86/x86_64 COS EFI-stub layout) → `uki` (single-blob `bootx64.efi` with EFI stub)
 - anything else (parse failure, missing markers, unknown layout) → hard error that lists the sections found, rather than silently falling back

The extraction and detection steps were also reordered for clarity. `extract_boot_components` now runs first and attempts to copy every component from partition 12, silently skipping any that aren't present, only the boot EFI fails hard. `detect_image_mode` then infers the image type from the extracted boot EFI as a pure inspection step. If a default type image is missing any required component, the subsequent hashing step fails loudly when it tries to read the file.

Example output:
```
{
  "channel": "hardened",
  "alg": "sha384",
  "image_type": "uki",
  "shim": "",
  "grub": "",
  "vmlinuz_a": "",
  "vmlinuz_b": "",
  "kernel_cmdline_a_hash": "",
  "kernel_cmdline_b_hash": "",
  "kernel_cmdline_a": "",
  "kernel_cmdline_b": "",
  "vmlinuz_a_pe": "",
  "vmlinuz_b_pe": "",
  "uki_efi": "fe01a800be3214ae1a27d9e8bda56284292923b54c3478d49309c273e04249d058a3ad96d10dc0e622e745755ccc2ad3"
}
```
In Cloudbuild, command to run this will be like
`    /usr/local/bin/measure.sh bchardeneddisk/disk.raw measure_output_bc_hardened_sha384.json hardened x86_64 sha384
`

Note: this change was generated with AI and carefully reviewed by the author.

## Tested on debug images
- Regression (default): ran the bumped measure container against the previous release image and verified the kernel digest in the RIM matches the event log on a booted VM.
  - Image: confidential-space-debug-260400
  - Compared output vmlinuz.B against the RIM.
- UKI mode: verified the EFI Authenticode hash for debug images.
  - Compared .uki_efi against EV_EFI_BOOT_SERVICES_APPLICATION (PCR 4) in the event log.
- Auto-detect: confirmed `default` is selected for the shim-based debug image and `uki` is selected for the bc/UKI debug image, and that the resulting JSON matches the explicit-flag output from the prior commit.

Prompt for AI review
To review `compute_efi_authenticode_hash`
```
Compare the compute_efi_authenticode_hash Python implementation in launcher/image/measure/measure.sh against the reference C implementation in sbsigntools image.c (https://git.kernel.org/pub/scm/linux/kernel/git/jejb/sbsigntools.git/tree/src/image.c). Flag any divergence that would change the resulting digest for a given EFI binary.
```

To review `detect_image_mode`
```
Review the `detect_image_mode` function in launcher/image/measure/measure.sh, which classifies the boot EFI binary as `default` (shim) or `uki` (EFI-stub) by inspecting PE sections. Flag any case where the detection logic would misclassify a real boot EFI binary, hard-error on a valid one, or crash on a malformed one.
```

### DIFF
--- a/launcher/image/measure/measure.sh
+++ b/launcher/image/measure/measure.sh
@@ -94,6 +94,7 @@ extract_partition_12() {
 #
 extract_boot_components() {
     local arch="$1"
+    local image_mode="$2"
     local shim_path=""
 
     if [ "$arch" == "x86_64" ]; then
@@ -108,15 +109,17 @@ extract_boot_components() {
     echo "Copying files from partition image '$P12_FILE'..."
 
     declare -A files_to_copy=(
-        ["::/syslinux/vmlinuz.A"]="$VMLINUZ_A_FILE"
-        ["::/efi/boot/grub.cfg"]="$GRUB_CFG_FILE"
         ["$shim_path"]="$BOOT_EFI_FILE"
-        ["::/efi/boot/grub-lakitu.efi"]="$GRUB_EFI_FILE"
     )
 
-    # vmlinuz.B file exists on amd64 but not on arm64
-    if [ "$arch" == "x86_64" ]; then
-        files_to_copy["::/syslinux/vmlinuz.B"]="$VMLINUZ_B_FILE"
+    if [ "$image_mode" == "default" ]; then
+        files_to_copy["::/syslinux/vmlinuz.A"]="$VMLINUZ_A_FILE"
+        files_to_copy["::/efi/boot/grub.cfg"]="$GRUB_CFG_FILE"
+        files_to_copy["::/efi/boot/grub-lakitu.efi"]="$GRUB_EFI_FILE"
+        # vmlinuz.B file exists on amd64 but not on arm64
+        if [ "$arch" == "x86_64" ]; then
+            files_to_copy["::/syslinux/vmlinuz.B"]="$VMLINUZ_B_FILE"
+        fi
     fi
 
     for src in "${!files_to_copy[@]}"; do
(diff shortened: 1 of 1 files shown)


launcher/image/measure/measure.sh
```diff
@@ -89,31 +89,66 @@ extract_partition_12() {
...
+# extract_boot_components does not need to copy it again.
+#
+# @param $1: build architecture ('x86_64' or 'aarch64').
+# @return The detected mode ('default' or 'uki') to stdout.
+#
+detect_image_mode() {
+    local arch="$1"
+    local shim_src
+    case "$arch" in
+        x86_64)   shim_src="::/efi/boot/bootx64.efi" ;;
```
alexmwu: shim_src should be bootx64_src or something. it's not definitely the shim

launcher/image/measure/measure.sh
```diff
@@ -89,31 +89,66 @@ extract_partition_12() {
...
+    if ! mcopy -i "$P12_FILE" "$shim_src" "$BOOT_EFI_FILE" &>/dev/null; then
+        echo "Error: Failed to extract '$shim_src' from '$P12_FILE'." >&2
+        return 1
+    fi
+
+    python3 - "$BOOT_EFI_FILE" <<'PY'
+import sys, lief
+b = lief.parse(sys.argv[1])
+if b is None:
+    print("uki"); sys.exit(0)
```
alexmwu: Is this a safe default? Consider using pefile https://github.com/erocarrera/pefile to get the sections (.sbat for shim, .linux for UKI) instead, and throw an error if it is neither UKI nor shim

launcher/image/measure/measure.sh
```diff
@@ -89,31 +89,66 @@ extract_partition_12() {
...
+    local image_mode="$2"
 
-    if [ "$arch" == "x86_64" ]; then
-        shim_path="::/efi/boot/bootx64.efi"
-    elif [ "$arch" == "aarch64" ]; then
-        shim_path="::/efi/boot/bootaa64.efi"
-    else
-        echo "Error: Unknown arch '$arch'."
-        return 1
+    # The boot EFI (bootx64.efi/bootaa64.efi) is already extracted by
```
alexmwu: This seems like the wrong approach. We should extract all boot components first then detect the image type, not rely on the detection to extract the bootx64 file

launcher/image/measure/measure.sh
```diff
@@ -286,40 +452,52 @@ main() {
...
+        echo "Kernel (vmlinuz.A) hash: $vmlinuz_a_hash"
+        echo "Kernel (vmlinuz.B) hash: $vmlinuz_b_hash"
+        echo "Kernel (vmlinuz.A) PE hash: $vmlinuz_a_pe_hash"
+        echo "Kernel (vmlinuz.B) PE hash: $vmlinuz_b_pe_hash"
+        echo "Kernel cmdline (image A): $kernel_cmdline_a"
+        echo "Kernel cmdline (image B): $kernel_cmdline_b"
+        echo "Kernel cmdline (image A) hash: $kernel_cmdline_a_hash"
+        echo "Kernel cmdline (image B) hash: $kernel_cmdline_b_hash"
+        echo "bootx64.efi/bootaa64.efi (shim) hash: $shim_hash"
+        echo "grub-lakitu.efi (grub) hash: $grub_hash"
```
alexmwu: You can remove these echos

launcher/image/measure/measure.sh
yanchengchen7: Updated to
```
#   .sbat present AND content contains 'https://github.com/rhboot/shim'
#       => 'default' (shim + grub + kernel layout)
#   .setup present (x86/x86_64 cos EFI-stub image)
#       => 'uki' (single-blob: bootx64.efi w/ EFI stub)
```
the shim detection logic is what is being used, and the current uki image only has three sections `.setup / .text / .data`

launcher/image/measure/measure.sh
yanchengchen7: Make sense, renamed.

launcher/image/measure/measure.sh
yanchengchen7: Got it, making the detect_image_mode to extract boot efi and infer image type. Is that what you have in mind?

launcher/image/measure/measure.sh
yanchengchen7: This is unchanged code, do we want to remove them?

launcher/image/measure/measure.sh
alexmwu: extract_boot_components first then detect_image_mode with a path to the bootx64 based on the architecture

launcher/image/measure/measure.sh
alexmwu: Not necessarily. You can keep it if you don't want to rerun the cloud build

alexmwu (approved):

launcher/image/measure/measure.sh
yanchengchen7: I see, reversed the order of the two functions, 
- extract_boot_components will try to copy all components, skip silently if component not present, and missing boot efi will fail hard. 
- `detect_image_mode` infer from the boot efi, if default image type missing components, the subsequent hash will fail.

PR #780, opened 2026-05-07
fix: prevent panic on all-0xFF CCEL data and limit issuing cert body reads

evilgensec:
## Summary

Two independent security fixes:

### 1. Index-out-of-range panic on all-0xFF CCEL data (`verifier/ita/client.go`)

The padding-trim loop in `convertRequestToTokenRequest` uses `trimIndex >= 0` as its guard:

```go
for ; trimIndex >= 0; trimIndex-- {
    c := data[trimIndex-1]  // panics when trimIndex == 0: data[-1]
```

When `trimIndex` reaches `0`, the loop body executes `data[-1]`, triggering an index-out-of-range panic. This happens unconditionally when `CcelData` consists entirely of `0xFF` bytes — for example, an uninitialized or all-padding ACPI CCEL table from a TDX platform. A crash here takes down the attestation service for the VM.

**Fix:** Change the loop guard to `trimIndex > 0`.

### 2. Unbounded `io.ReadAll` on issuing certificate URLs (`internal/cert.go`)

`fetchIssuingCertificate` fetches URLs taken from `cert.IssuingCertificateURL` — which originates from the AK certificate field in client-supplied `pb.Attestation` protobufs. There was no size limit on the HTTP response body:

```go
certBytes, err := io.ReadAll(resp.Body)  // no size limit
```

An attacker-controlled certificate pointing to a server returning an arbitrarily large body causes the process to exhaust memory (OOM-kill).

**Fix:** Wrap with `io.LimitReader` capped at 1 MiB — sufficient for any real DER certificate.

## Tests

- `TestConvertRequestToTokenRequestAllPaddedCCELData`: verifies all-`0xFF` CCEL data is handled without panic.
- `TestFetchIssuingCertificateLargeBodyTruncated`: verifies oversized cert bodies are truncated and return an error rather than OOM.

### DIFF
--- a/internal/cert.go
+++ b/internal/cert.go
@@ -10,6 +10,9 @@ import (
 const (
 	maxIssuingCertificateURLs = 3
 	maxCertChainLength        = 4
+	// maxCertBodyBytes limits HTTP response reads when fetching issuing
+	// certificates to prevent OOM on malicious or misbehaving CA servers.
+	maxCertBodyBytes = 1 * 1024 * 1024 // 1 MiB
 )
 
 // GetCertificateChain constructs the certificate chain for the key's certificate.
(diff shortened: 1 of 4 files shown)


evilgensec:
Friendly ping. Is this in the review queue, or is there something you need from me to move it along?

PR #782, opened 2026-05-08
[Do not submit] Draft launcher updates for BC guest

mayafleischer:


agent/proto/hostservicepb/attestation.proto
mayafleischer, line 1: Files in this folder will be removed once this proto definition is added to the confidential space repo

launcher/image/cloudbuild.yaml
```diff
@@ -27,8 +27,8 @@ steps:
            '-script=fixup_oem.sh']
   - name: 'gcr.io/cos-cloud/cos-customizer'
     args: ['finish-image-build',
-           '-oem-size=900M',
-           '-disk-size-gb=12',
+           '-oem-size=500M',
```
mayafleischer, line 30: These values were some what arbitrary but seem to work

PR #783, opened 2026-05-08
Update BC build 

mayafleischer:
Update the BC build process to skip GCE image creation if the image already exists with that name.  Currently this is causing re-runs of the presubmits tests for the same commit to fail.

### DIFF
--- a/launcher/image/bc_cloudbuild.yaml
+++ b/launcher/image/bc_cloudbuild.yaml
@@ -164,10 +164,11 @@ steps:
         if [ -z "$$FEATURES" ]; then
           FEATURES="UEFI_COMPATIBLE"
         fi
+        gcloud compute images delete ${_OUTPUT_IMAGE_NAME} --quiet || true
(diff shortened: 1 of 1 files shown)


launcher/image/bc_cloudbuild.yaml
```diff
@@ -164,10 +164,14 @@ steps:
         if [ -z "$$FEATURES" ]; then
           FEATURES="UEFI_COMPATIBLE"
         fi
+        if gcloud compute images describe ${_OUTPUT_IMAGE_NAME} > /dev/null 2>&1; then
+          echo "Result image ${_OUTPUT_IMAGE_NAME} already exists in project. Exiting."
```
yanchengchen7, line 168: Not deletion?

mayafleischer: It will skip the images create step, since an image already exists.
This is consistent with what happens for the vg and regular cs images.

yanchengchen7 (approved):

jkl73 (approved):

PR #786, opened 2026-05-12
Adding network setup script to prioritize idpf nics and provide tap dev IP address

gjacobwells:
Without this, tap devices try to take traffic that is meant to leave on the IDPF NIC. This prevents that by prioritizing IDPF with the metric number of 100. The default is 1024 and setting the number lower means higher priority Also providing static IP address to the tap device.

The script is placed into the OEM partition and then on boot, the entry point script will attempt to run the net setup script.

Wrapped into host image here: https://confidential-space-internal-review.git.corp.google.com/c/chewbacca/host-acos/+/2420?tab=checks

Tested on A4 machine and I see the IP route has the correct "metric 100" for the IDPF devices
```
jacobdw_google_com@jacobdw-a4-test ~ $ ip route show
default via 10.128.0.1 dev eth0 proto dhcp src 10.128.0.117 metric 100 
10.10.0.0/24 via 10.10.0.1 dev eth1 proto dhcp src 10.10.0.195 metric 100 
10.10.0.1 dev eth1 proto dhcp scope link src 10.10.0.195 metric 100 
10.128.0.0/20 via 10.128.0.1 dev eth0 proto dhcp src 10.128.0.117 metric 100 
10.128.0.1 dev eth0 proto dhcp scope link src 10.128.0.117 metric 100 
169.254.169.254 via 10.128.0.1 dev eth0 proto dhcp src 10.128.0.117 metric 100 
169.254.169.254 dev eth1 proto dhcp scope link src 10.10.0.195 metric 100 
```

### DIFF
--- a/launcher/image/bc-guest/bc_network_setup.sh
+++ b/launcher/image/bc-guest/bc_network_setup.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+mkdir -p /etc/systemd/network/
+
+cat << 'EOF' > /etc/systemd/network/10-idpf.network
+[Match]
+Driver=idpf
+
+[Network]
+DHCP=yes
+IPv6AcceptRA=yes
+
+[DHCPv4]
+RouteMetric=100
+
+[DHCPv6]
+RouteMetric=100
+EOF
+
+cat << 'EOF' > /etc/systemd/network/10-virtio.network
+[Match]
+Driver=virtio_net
+
+[Network]
+Address=192.168.100.2/24
+
+[Route]
+Gateway=192.168.100.1
+Metric=1024
+EOF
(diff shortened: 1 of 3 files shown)


launcher/cloudbuild.yaml
```diff
@@ -12,6 +12,50 @@ substitutions:
   '_BC_DEBUG_IMAGE': 'cos-bcdbg-guest-125-19216-220-185'
 
 steps:
+- name: 'ubuntu'
```
jkl73: Instead of creating it here, can we just create the setup-network.sh as a static file and put it in the bc-guest directory?

gjacobwells: Oh yeah, good call. Refactored that.

launcher/image/bc-guest/bc_network_setup.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
jkl73, line 1: Can you move this file to bc-guest directory?

jkl73 (approved):

launcher/image/bc-guest/bc_network_setup.sh
gjacobwells: done

PR #789, opened 2026-05-13
Dynamically fetch GPU drivers and vmlinux artifacts based on base ima…

mayafleischer:
Update the build steps in both `launcher/cloudbuild.yaml` and `launcher/image/bc_cloudbuild.yaml` to dynamically determine the correct versioned paths for downloading required files

PR #790, opened 2026-05-13
bc launcher updates

mayafleischer:
Update the launcher code for bare metal use case.

A new experiment `BcMode ` is introduced to gate the logic flow.

**Agent Changes**
The following changes were made, conditional on BcMode being enabled:
- Skip TPM initialization 
- Added HostAttestation() method to fetch host attestation from the host service via VSOCK 
- Skip attestation key fetching, PCR measurements, and certificate chain fetching 
- Updated agent to handle nil attestation key

**Launcher Changes**
- Added BcMode experiment flag to control launcher behavior
- Skip TPM-related initialization and DA (Dictionary Attack) parameter checking in BC mode
- Updated key manager initialization to skip when in BC mode
- Modified filesystem verification to bypass verifyFsAndMount in BC mode
- Updated host attestation handler to use real host attestation from the service when in BC mode instead of dummy implementation

**Testing**
- Added test for HostAttestation() in the agent test to verify it errors when BcMode is disabled
- Updated mock implementations in tests to support the new HostAttestation() method


agent/agent.go
```diff
@@ -133,10 +142,19 @@ type agent struct {
 // - signaturesFetcher is a func to fetch container image signatures associated with the running workload.
 // - logger will log any partial errors returned by VerifyAttestation.
 func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher, verifierClient verifier.Client, principalFetcher principalIDTokenFetcher, sigsFetcher SignatureFetcher, exps Experiments, logger Logger, deviceROTs []DeviceROT, signedImageRepos []string) (AttestationAgent, error) {
+	if exps.BcMode && tpm != nil {
+		return nil, fmt.Errorf("TPM should not be passed in BC mode")
+	}
```
yawangwang: W/ this new bc mode, should KPS attestation service follow the same pattern to skip TPM relevant initializations? And why would want to skip it for bc?

agent/agent.go
```diff
@@ -68,6 +74,7 @@ type AttestationAgent interface {
 	AttestationEvidence(ctx context.Context, challenge []byte, extraData []byte, opts AttestAgentOpts) (*attestationpb.VmAttestation, error)
 	Refresh(context.Context) error
 	Close() error
+	HostAttestation(ctx context.Context, challenge []byte) ([]byte, error)
```
jkl73: AttestHost

agent/agent.go
```diff
@@ -133,10 +142,19 @@ type agent struct {
...
 func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher, verifierClient verifier.Client, principalFetcher principalIDTokenFetcher, sigsFetcher SignatureFetcher, exps Experiments, logger Logger, deviceROTs []DeviceROT, signedImageRepos []string) (AttestationAgent, error) {
+	if exps.BcMode && tpm != nil {
+		return nil, fmt.Errorf("TPM should not be passed in BC mode")
+	}
 	// Fetched the AK and save it, so the agent doesn't need to create a new key everytime
-	ak, err := akFetcher(tpm)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create an Attestation Agent: %w", err)
+	var ak *client.Key
+	if !exps.BcMode {
```
jkl73: Instead of adding condition statements all over the function:
Consider creating a new agent struct (bcagent , implements AttestationAgent)

agent/agent.go
jkl73: Right, those VMs won't have TPMs, so will have to skip those

launcher/launcher/main.go
```diff
@@ -87,11 +88,20 @@ func main() {
...
-		logger.Error(fmt.Sprintf("failed to verify filesystem and mounts: %v\n", err))
-		exitCode = rebootRC
-		logger.Error(exitMessage, "exit_code", exitCode, "exit_msg", rcMessage[exitCode])
-		return
+	// Check for BC Mode via environment variable because experiments are not loaded yet.
+	// In BC mode, we bypass verifyFsAndMount because the filesystem structure (Virtio devices,
+	// host-emulated state) differs from standard CS/VG and does not use a vTPM for stateful encryption.
+	bcMode := os.Getenv("BC_MODE") == "true"
+
+	if !bcMode {
```
jkl73: We should still check the mounts make sure they are as expected.

The mounts might be different, maybe we should create a different set.

Also we won't need to create env var just for this check.

launcher/container_runner.go
```diff
@@ -721,8 +722,12 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 	}
 
 	var workloadService *workloadservice.Server
-	// create and start the key manager server
-	if r.launchSpec.Experiments.EnableKeyManager {
+	if r.launchSpec.Experiments.BcMode {
```
jkl73: KeyManager can still be enabled under BC, we can leave this logic as is, and change later

launcher/image/bc_cloudbuild.yaml
```diff
@@ -113,7 +127,7 @@ steps:
           "--vmlinux=./vmlinux"
           "--src-image=./disk.raw"
           "--out-image=output.bin"
-          "--oem-fs-size=500M"
+          "--oem-fs-size=1500M"
```
jkl73: what's the reason for the size increase?

launcher/container_runner.go
mayafleischer: Reverted.

agent/agent.go
mayafleischer: Renamed.

agent/agent.go
mayafleischer: Updated to create bcAgent using composition

launcher/image/bc_cloudbuild.yaml
mayafleischer: I needed to increase this to have space for the GPU driver installation. 
If we update to not copy the gpu files, we can revert.

launcher/launcher/main.go
mayafleischer: Added this check back - seems to be the same.

agent/agent.go
```diff
@@ -193,17 +211,57 @@ func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher
...
-		attestAgent.avRot = tpmAR
 	}
 
 	// Add deviceRoTs to the CPU attestation root.
 	attestAgent.avRot.AddDeviceROTs(deviceROTs)
 	return attestAgent, nil
 }
 
+func createBCAgent(tpm io.ReadWriteCloser, verifierClient verifier.Client, principalFetcher principalIDTokenFetcher, sigsFetcher SignatureFetcher, exps Experiments, logger Logger, deviceROTs []DeviceROT, signedImageRepos []string) (AttestationAgent, error) {
+	if tpm != nil {
```
yawangwang: If BC env doesn't support TPM at all, we don't have to pass in tpm as a param.

launcher/launcher/main.go
```diff
@@ -109,6 +103,13 @@ func main() {
 		return
 	}
 
+	if err := verifyFsAndMount(); err != nil {
+		logger.Error(fmt.Sprintf("failed to verify filesystem and mounts: %v\n", err))
+		exitCode = rebootRC
+		logger.Error(exitMessage, "exit_code", exitCode, "exit_msg", rcMessage[exitCode])
+		return
+	}
```
yawangwang: Let's keep it to the original place.

launcher/launcher/main.go
```diff
@@ -190,42 +191,47 @@ func startLauncher(launchSpec spec.LaunchSpec, serialConsole *os.File) error {
...
-	} else {
-		if !daInfo.StartupClearOrderly {
-			logger.Warn(fmt.Sprintf("Failed orderly startup. Avoid using instance reset. Instead, use instance stop/start. DA lockout counter incremented: LockoutCounter: %d / MaxAuthFail: %d", daInfo.LockoutCounter, daInfo.MaxTries))
-		}
-
-		if err := launcher.SetTPMDAParams(tpm, expectedTPMDAParams); err != nil {
-			logger.Error(fmt.Sprintf("Failed to set DA params: %v", err))
+	var tpm io.ReadWriteCloser
+	if !launchSpec.Experiments.BcMode {
+		tpm, err = tpm2.OpenTPM("/dev/tpmrm0")
```
yawangwang: Can we abstract tpm relevant initializations/operations into a separate func?

yawangwang (approved):

launcher/cloudbuild.yaml
jkl73: we should update this to the latest one

launcher/image/bc_cloudbuild.yaml
jkl73: I think we don't need to increase the size for this image, we can keep it 500 and the disk-size can be 15

agent/agent.go
jkl73: yup, we can make also make the check before calling createBCAgent

agent/agent.go
```diff
@@ -193,17 +211,57 @@ func CreateAttestationAgent(tpm io.ReadWriteCloser, akFetcher util.TpmKeyFetcher
...
+			qp:     qp,
+			cosCel: gecel.NewConfComputeMR(),
+		}
+		baseAgent.measuredRots = append(baseAgent.measuredRots, tdxAR)
+
+		logger.Info("Using TDX RTMR as attestation root.")
+		baseAgent.avRot = tdxAR
+	} else {
+		return nil, fmt.Errorf("running in BC mode but TDX not supported")
+	}
```
jkl73: looks like some repeating code, maybe worth to create new function

launcher/image/bc_cloudbuild.yaml
```diff
@@ -56,8 +62,16 @@ steps:
...
+    entrypoint: 'bash'
+    args:
+      - '-c'
+      - |
+        if [[ "${_IMAGE_ENV}" == "debug" ]]; then
+          KERNEL_DIR="cchostdebug-amd64-gcp"
+        else
+          KERNEL_DIR="cchost-amd64-gcp"
+        fi
+        gcloud storage cp "gs://cos-tools/19216.220.185/$${KERNEL_DIR}/vmlinux" ./vmlinux
```
jkl73: might want to make "19216.220.185" a variable for easier update later

agent/agent.go
jkl73: Also we probably don't need verifierClient, we don't need AttestWithClient or Attest to work in bc

launcher/image/bc_cloudbuild.yaml
mayafleischer: This is addressed in https://github.com/google/go-tpm-tools/pull/789

agent/agent.go
mayafleischer: Removed the tpm and verifierClient params

launcher/launcher/main.go
mayafleischer: Done

agent/agent.go
mayafleischer: Done

launcher/image/bc_cloudbuild.yaml
mayafleischer: Ok - I have decreased it

launcher/image/bc_cloudbuild.yaml
```diff
@@ -48,6 +48,12 @@ steps:
            /workspace/oem-install/confidential_space/
         cp ./launcher/image/nodeproblemdetector/* /workspace/oem-install/confidential_space/nodeproblemdetector/
 
+        mkdir -p /workspace/oem-install/gpu_driver
```
jkl73: We don't need to copy them here

mayafleischer: Removed and reduced the size.

launcher/cloudbuild.yaml
```diff
@@ -8,8 +8,8 @@ substitutions:
   '_BUCKET_NAME': '${PROJECT_ID}_cloudbuild'
   '_VG_BASE_IMAGE': 'cos-125-19216-220-185'
   '_BC_BASE_IMAGE_PROJECT': 'confidential-space-images-dev'
-  '_BC_BASE_IMAGE': 'cos-bc-guest-125-19216-220-185'
-  '_BC_DEBUG_IMAGE': 'cos-bcdbg-guest-125-19216-220-185'
+  '_BC_BASE_IMAGE': 'cos-bc-guest-125-19216-395-7'
```
jkl73: the version should match between the two images (maybe just use 395-4)

agent/agent.go
```diff
@@ -284,14 +336,16 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
...
-		certChain, err := internal.GetCertificateChain(a.fetchedAK.Cert(), http.DefaultClient)
-		if err != nil {
-			return nil, fmt.Errorf("failed when fetching certificate chain: %w", err)
-		}
-
 		v.CanonicalEventLog = cosCel.Bytes()
-		v.IntermediateCerts = certChain
-		v.AkCert = a.fetchedAK.CertDERBytes()
+
+		if !a.experiments.BcMode {
```
jkl73: This is probably not needed, iiuc bcagent shouldn't be able call this function anyway

jkl73 (approved):

PR #799, opened 2026-05-20
Add experiment retry

mayafleischer:
Adds retry to experiment sync and skips experiment sync entirely for bc & vg.


launcher/spec/launch_spec.go
```diff
@@ -407,18 +409,30 @@ func fetchExperiments(logger logging.Logger) experiments.Experiments {
...
+	var e experiments.Experiments
+	_ = backoff.Retry(func() error {
+		if err := exec.Command(binaryPath, args).Run(); err != nil {
+			logger.Error(fmt.Sprintf("failure during experiment sync: %v\n", err))
+		}
+		var err error
+		e, err = experiments.New(experimentsFile)
+		if err != nil {
+			logger.Error(fmt.Sprintf("failed to read experiment file: %v\n", err))
+			// do not fail if experiment retrieval fails
```
yawangwang: this error suggest something goes wrong when unshalling JSON bytes, the `exec.Command` above retrieves experiment value.

launcher/spec/launch_spec.go
```diff
@@ -407,18 +409,30 @@ func fetchExperiments(logger logging.Logger) experiments.Experiments {
...
-	if err != nil {
-		logger.Error(fmt.Sprintf("failure during experiment sync: %v\n", err))
-	}
-	e, err := experiments.New(experimentsFile)
-	if err != nil {
-		logger.Error(fmt.Sprintf("failed to read experiment file: %v\n", err))
-		// do not fail if experiment retrieval fails
-	}
+	var e experiments.Experiments
+	_ = backoff.Retry(func() error {
```
yawangwang: Can we log the error returned by retry policy? This will give us a way to calculate the error ratio after applying this strategy.

launcher/spec/launch_spec.go
yawangwang, line 408: Add unit tests as the behavior of this func grows more complicated.

mayafleischer: Done

launcher/spec/launch_spec.go
mayafleischer: Done

launcher/spec/launch_spec.go
```diff
@@ -406,19 +408,46 @@ func isHardened(kernelCmd string) bool {
...
+		if err != nil {
+			logger.Error(fmt.Sprintf("failed to read experiment file: %v\n", err))
+		}
+		// This is expected to be true if experiment sync is successful.
+		if !e.EnableTestFeatureForImage {
+			return fmt.Errorf("experiments synced but EnableTestFeatureForImage is false")
+		}
+		return nil
+	}, experimentSyncBackoffPolicy()); err != nil {
+		logger.Error(fmt.Sprintf("experiment retrieval failed after retries: %v\n", err))
```
mayafleischer, line 440: Perhaps we do want to fail if experiment retrieval fails after these changes?

launcher/spec/launch_spec.go
yawangwang: A better way to test it is to pass in a file path so that you don't have to do extra restore or cleanup in your unit test code.

launcher/spec/launch_spec.go
yawangwang: Good call, that was my original thought too. But if we fail here then we may lose the ability to monitor if experiment sync has actually been improved or not.

launcher/spec/launch_spec.go
```diff
@@ -406,19 +408,46 @@ func isHardened(kernelCmd string) bool {
...
+		// Do not fail if experiment retrieval fails.
 	}
 	return e
 }
 
+func experimentSyncBackoffPolicy() backoff.BackOff {
+	b := backoff.NewConstantBackOff(time.Second)
+	return backoff.WithMaxRetries(b, 3)
```
yawangwang: nit: so this assumes that the average latency for the sync call is ~1s, could you confirm if that's sufficient based on historical data? I'm a bit concerned if 1s would being a short amount of time.

yawangwang (approved):

PR #802, opened 2026-05-21
Add presubmit image cleanup build step

yanchengchen7:
Add an image cleanup step in the launcher Cloud Build workflow to only delete **presubmit** generated images. 

## Test
[Cloud Build](https://pantheon.corp.google.com/cloud-build/builds;region=us-west1/d922740e-f80e-44c7-ba4b-df3a8aa1807c?project=confidential-space-images-dev&e=13802955&mods=logs_tg_prod)


jkl73:
we might want to keep the presubmit images for testing, so I don't think we want to delete them right away

yanchengchen7:
> we might want to keep the presubmit images for testing, so I don't think we want to delete them right away

I see, ok, reverting it.

PR #806, opened 2026-05-22
Address conflicting routes and apply network optimizations

gjacobwells:
This PR accomplishes a few things:
- Addresses "martian source" issues by configuring primary GCE interface (eth0) with RouteMetric=100 and secondary GCE interfaces with RouteMetric=200 so eth0 is seen as the primary when reaching MDS.
- Disable DHCP and LinkLocalAddressing on virtio network so the tap device won't conflict with idpf devices.
- Add's a one-time network optimization script that changes the network ring lengths
- Add a runtime network optimization script that runs after the device is reset to avoid the settings being overwritten.
- Set the SWIOTLB to 32G since this is the largest it could feasibly need at max ring lengths
- Disable google_set_multiqueue, which can override network settings.


launcher/image/bc-guest/bc_network_optimization.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
jkl73, line 1: Indentation in this file looks different comparing to other scripts

jkl73 (approved):

launcher/image/bc-guest/bc_network_optimization.sh
gjacobwells: good eye. Fixed

PR #809, opened 2026-05-27
Update BC base COS image

meetrajvala:
The current base COS image is missing the `TDX_CAPABLE` and `SNP_SVSM_CAPABLE` Guest OS features. Because bc aCOS build process inherits these features from the base image, their absence in base COS prevents them from being included in the bc aCOS image. Consequently, we cannot spin up TDX or SEV-enabled VMs using the new image.

**Solution:** Revert the base to cos-bc-guest-125-19216-395-4, which contains the required Guest OS attributes.


jkl73 (dismissed):

meetrajvala:
As we have recreated images, this PR is no longer required.

PR #811, opened 2026-05-27
Add GetKeyEndorsement gRPC call to teeserver

yawangwang:
This PR introduces BC mode to teeserver to make gRPC network to get keyEndorsement from KPS VM.

Tested on a BC baremetal machine that launches two VMs, and we can successfully verify the KeyEndorsement attestation from KPS can be collected.


launcher/teeserver/kem_attestation.go
```diff
@@ -0,0 +1,91 @@
...
+			Handle: req.KeyHandle.Handle,
+		},
+	}
+	client := pb.NewAttestationServiceClient(a.conn)
+	resp, err := client.GetKeyEndorsement(ctx, kemReq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get remote key endorsement: %v", err)
+	}
+
+	return resp.GetKeyAttestation().GetAttestation(), nil
```
jkl73, line 103: should double check the resp to avoid panic (if the resp is malformed for some reasons)

launcher/teeserver/kem_attestation.go
```diff
@@ -0,0 +1,91 @@
...
+	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+
+	kemReq := &pb.GetKeyEndorsementRequest{
+		Challenge: req.Challenge,
+		KeyHandle: &kpmkeymanager.KeyHandle{
+			Handle: req.KeyHandle.Handle,
+		},
+	}
+	client := pb.NewAttestationServiceClient(a.conn)
```
jkl73: Maybe the client can be saved in the remoteKEMAttester instead of creating it here. Up to you, I'm not too worried

launcher/teeserver/kem_attestation.go
```diff
@@ -0,0 +1,91 @@
...
+	"time"
+
+	// Bootstrap environment flags to prevent duplicate proto registration panics.
+	_ "github.com/google/go-tpm-tools/launcher/teeserver/envinit"
+
+	attestationpb "github.com/GoogleCloudPlatform/confidential-space/server/proto/gen/attestation"
+
+	pb "github.com/GoogleCloudPlatform/key-protection-module/keymanager/attestation_service/proto/gen"
+
+	kpmkeymanager "github.com/GoogleCloudPlatform/key-protection-module/km_common/proto"
```
jkl73, line 15: can we just use this proto, to avoid the duplicate proto situation?

launcher/teeserver/tee_server.go
```diff
@@ -81,8 +87,14 @@ func New(ctx context.Context, unixSock string, a agent.AttestationAgent, logger
...
 	}
 
+	kemAttester, conn, err := initKEMAttester(launchSpec.Experiments.BcMode, keyClaimsProvider, a)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize KEM attester: %v", err)
+	}
+
 	teeServer := TeeServer{
 		netListener: nl,
+		kpsConn:     conn,
```
jkl73: do we need to save the conn here, since we already have the kemAttester

jkl73 (approved):

launcher/teeserver/kem_attestation.go
yawangwang: Unfortunately no, the `kem_attestation.go` implements GetKeyEndorsement for both VG and BC. Since VG relies on library call which imports `github.com/google/go-tpm-tools/keymanager/km_common/proto`, whereas BC relies on gRPC call which imports `github.com/GoogleCloudPlatform/key-protection-module/km_common/proto` . Those tow libraries have to be imported within the same package, causing proto registration panics. 

To fully resolve that, we have to fully migrate using protos defined in https://github.com/GoogleCloudPlatform/key-protection-module, which can be treated as a lower priority task.

launcher/teeserver/kem_attestation.go
yawangwang: protobuf is nil safe, but i still managed to add nil checks as you suggested.

launcher/teeserver/tee_server.go
yawangwang: ah that also works, done.

launcher/teeserver/kem_attestation.go
yawangwang: done, saving client in remoteKEMAttester.

PR #813, opened 2026-05-28
Add acpi to endpoint

mayafleischer:
Add option to get ACPI data as part of getKeyEndorsement request.  Only available for BcMode.


launcher/teeserver/tee_server.go
mayafleischer, line 214: Would we like to add support for "acpi_data" in the query parameters?
It is not necessary for this purpose.
But if so, I will still configure it such that asking for acpi_data in the query if not in BcMode has no effect.

alexmwu: It appears we implemented this incorrectly: https://google.aip.dev/157. That's on me 😅. We were supposed to use google.protobuf.FieldMask. See https://netflixtechblog.com/practical-api-design-at-netflix-part-1-using-protobuf-fieldmask-35cfdc606518 or https://docs.cloud.google.com/apis/docs/system-parameters as examples

You're right this is unnecessary at this moment. I think we can probably remove this implementation and just rely on the bool Request* pattern (moving the DeviceReport over to the API request object as well). If we wanted to follow the AIP and cloud API standards, then we would implement using the X-Goog-FieldMask system parameter

alexmwu (approved):

launcher/teeserver/tee_server.go
mayafleischer: Ack - I am going to make a note of this and leave it as is for right now.

jkl73 (approved):

PR #821, opened 2026-06-04
Hash the challenge before concatenating it with `hash(extraData)`

susienme:


jkl73:
need to be careful on changing the computation of nonce, may need to double check the verifier side as well.

PR #822, opened 2026-06-04
[Launcher] Install GPU Daemons and set to READY state

Sibcgh:
# Install GPU Daemons and set to READY state

## Overview
This PR adds support for running GPU passthrough in Confidential Computing (`GpuBcMode`). It introduces a privileged background sidecar container (`guest-gpu-tools`) to initialize NVLink switches, load driver modules, and promote enclaves to `READY` state before launching the workload container.

---

## Changes

*   **`launcher/launcher/main.go`**
    Parses `tee-gpu-bc-mode` metadata, creates shared host paths, launches the GPU sidecar, and blocks workload startup until GPU services report ready.

*   **`launcher/spec/launch_spec.go`**
    Adds the `GpuBcMode` configuration flag (mapped from `tee-gpu-bc-mode`).

*   **`launcher/container_runner.go`**
    Exposes host GPU libraries and character devices to the workload container, and conditionalizes the `/var/run/nvidia-fabricmanager` bind-mount to only load when `GpuBcMode` is active.

*   **`launcher/internal/gpu/daemons/gpudaemoninstaller.go`**
    Manages building and starting the GPU sidecar container, bind-mounts `/var/run/nvidia-fabricmanager` to share the Unix socket, and polls `/run/nvidia/gpu-ready` to verify readiness.

*   **`launcher/internal/gpu/daemons/entrypoint.sh`**
    Sidecar container startup script: configures NVSwitch rail policy, launches `nvlsm`/`nv-fabricmanager`, sets GPU enclaves to ready state, and writes the `gpu-ready` completion marker.

*   **`launcher/internal/gpu/daemons/Dockerfile`**
    Defines the GPU sidecar image packaging `nv-fabricmanager`, `nvlsm`, and `opensm` daemons.

*   **`launcher/internal/gpu/daemons/distroless_placeholder/`** (includes `Dockerfile.distroless.template` and `manager.go.template`)
    Adds template scaffolding to support migrating to lightweight (~30MB) Distroless sidecar images in the future.

*   **`launcher/internal/gpu/driverinstaller.go`**
    Configures host-level GPU driver modules and forces GSP (GPU System Processor) firmware offloading (required for Blackwell/Hopper GPUs in TDX).

*   **`launcher/internal/launchermount/mount.go`** and **`tmpfs.go`**
    Relaxes strict mount constraints to allow mounting `/var/lib/containerd` and `/var/lib/docker` to a `tmpfs` RAM-disk (min 10GB) for local manual compiling.

*   **`launcher/image/bc_cloudbuild.yaml`**
    Adds the `/workspace/oem-install/gpu_daemons` directory setup and copies the GPU sidecar Dockerfile/entrypoint scripts to package them into the preloader VM image build.


launcher/internal/gpu/daemons/gpudaemoninstaller.go
Sibcgh, line 1: need to remove this and just keep the usr share path, perhaps more paths later for different build paths

launcher/internal/gpu/daemons/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Install dependencies needed for the script and installation
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    infiniband-diags \
+    sed \
+    grep \
+    wget \
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+# Add NVIDIA CUDA repository
```
yawangwang, line 15: normally we don't install cuda libraries for customers; they would need to install cuda libraries via their workload, what we provide is just a minimal set of dependences to run a GPU workload.

launcher/internal/gpu/daemons/entrypoint.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# Dynamically set LD_LIBRARY_PATH to use host driver libraries
+if [ -f /proc/driver/nvidia/version ]; then
+    driver_version=$(cat /proc/driver/nvidia/version | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -n 1)
+    if [ -d "/opt/nvidia-host/${driver_version}/lib64" ]; then
+        export LD_LIBRARY_PATH=/opt/nvidia-host/${driver_version}/lib64:$LD_LIBRARY_PATH
+    elif [ -d "/opt/nvidia-host/lib64" ]; then
+        export LD_LIBRARY_PATH=/opt/nvidia-host/lib64:$LD_LIBRARY_PATH
+    fi
+fi
```
yawangwang, line 14: this could have been avoided if customers choose to install CUDA libraries on their own, which keeps consistent with VG.

launcher/launcher/main.go
```diff
@@ -205,14 +207,43 @@ func startLauncher(launchSpec spec.LaunchSpec, serialConsole *os.File) error {
...
+		logger.Info("Waiting for GPU services to report ready...")
+		waitCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+		defer cancel()
+		if err := daemons.WaitForGPUServices(waitCtx); err != nil {
+			return fmt.Errorf("failed to initialize GPU daemons: %w", err)
+		}
+		logger.Info("GPU services are ready. Proceeding to launch workload container.")
 	} else {
-		// TODO: Remove this when BC GPU installation is supported
+		// TODO: Remove this check when BC GPU installation is supported
```
yawangwang, line 232: I think you can remove this check because this PR installs GPU for BC

launcher/internal/gpu/driverinstaller.go
```diff
@@ -109,51 +166,31 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
...
+			// we fallback to mounting a temporary, volatile RAM-disk (tmpfs) to complete the driver
+			// compilation in memory.
+			di.logger.Info(fmt.Sprintf("Failed to install GPU driver: %v. Retrying with tmpfs mount...", installErr))
+			var tmpfsErr error
+			di.tmpfs, tmpfsErr = launchermount.CreateTmpfsMount(map[string]string{
+				launchermount.DestinationKey: InstallationHostDir,
+				// The driver installer container requires ~4G of unpack space. We allocate a bare
+				// minimum 2GB RAM-disk to complete compilation and clear out immediately after.
+				launchermount.SizeKey: "2G",
+			})
```
yawangwang: Can we increase the size of guest boot disk, that way we can avoid mounting to a tmpFS.

launcher/launcher/main.go
```diff
@@ -205,14 +207,43 @@ func startLauncher(launchSpec spec.LaunchSpec, serialConsole *os.File) error {
 	}
 	ctx := namespaces.WithNamespace(context.Background(), namespaces.Default)
 
+	var gpuTmpfs launchermount.Mount
 	if launchSpec.InstallGpuDriver {
 		installer := gpu.NewDriverInstaller(containerdClient, launchSpec, logger)
```
yawangwang, line 210: BC workload VM will fall into this path and install GPU driver via cos-gpu-installer, which is something we want to avoid at the very begining.

launcher/spec/launch_spec.go
```diff
@@ -96,6 +96,7 @@ const (
 	cgroupNS                   = "tee-cgroup-ns"
 	gcaServiceEnv              = "gca-service-env"
 	installGpuDriver           = "tee-install-gpu-driver"
+	gpuBcMode                  = "tee-gpu-bc-mode"
```
yawangwang, line 99: BC mode is controlled via experiment flag, not launchSpec. (see https://github.com/google/go-tpm-tools/blob/cd1e39201697405d9e0f24334c6d1f3c11cb28ef/launcher/image/bc-guest/bcexperiment.json#L6)

launcher/internal/gpu/daemons/gpudaemoninstaller.go
```diff
@@ -0,0 +1,377 @@
...
+	"github.com/containerd/containerd/cio"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/oci"
+	"github.com/google/go-tpm-tools/launcher/internal/logging"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+const (
+	// GuestGPUToolsImageRef is the image reference for guest GPU tools sidecar
+	GuestGPUToolsImageRef = "docker.io/library/guest-gpu-tools:latest"
```
yawangwang, line 24: what's the plan to measure this container?

launcher/image/bc_cloudbuild.yaml
```diff
@@ -64,6 +65,9 @@ steps:
            ./launcher/image/bc-guest/bc_network_rebind.sh \
            /workspace/oem-install/confidential_space/
         cp ./launcher/image/nodeproblemdetector/* /workspace/oem-install/confidential_space/nodeproblemdetector/
+        cp ./launcher/internal/gpu/daemons/Dockerfile \
```
yawangwang: are they needed? especially since you'll be running a remote workload to launch GPU daemons.

yawangwang:
Fix lint errors.

launcher/spec/launch_spec.go
```diff
@@ -142,6 +143,7 @@ type LaunchSpec struct {
 	AddedCapabilities          []string
 	CgroupNamespace            bool
 	InstallGpuDriver           bool
+	GpuBcMode                  bool
```
jkl73, line 146: Why do we need an additional flag? we have bc and gpu already, which should be sufficient?

launcher/internal/gpu/driverinstaller.go
jkl73: we can maybe bypass this by increase the size of image when doing oem-preloading

launcher/internal/gpu/daemons/distroless_placeholder/Dockerfile.distroless.template
```diff
@@ -0,0 +1,37 @@
+# TODO(future): Build this Dockerfile to generate the distroless GPU sidecar container image.
```
jkl73, line 1: Do we need all the distroless_placeholder ?

launcher/internal/gpu/daemons/Dockerfile
jkl73: looks like this is just downloading the keyring for the repo for the FM nvlsm, etc..

so the cuda library is not installed.

launcher/internal/gpu/daemons/gpudaemoninstaller.go
jkl73: ya, this needs to be part of the OEM partition

PR #825, opened 2026-06-04
 launcher: Pin Google CA in GCP TLS connections

brongan:
This change creates a HTTP client configured with `pki.goog/roots.pem` as
 its trust bundle ensuring rogue CAs cannot be used to MiTM connections
 to Google owned services from the launcher.

 During image compilation, `roots.pem` is dropped into the OEM partition
 and used by all Gcloud SDK calls.

### DIFF
--- a/launcher/cloudbuild.yaml
+++ b/launcher/cloudbuild.yaml
@@ -14,6 +14,10 @@ substitutions:
 
 
 steps:
+- name: 'gcr.io/cloud-builders/curl'
+  id: DownloadGoogleRoots
+  args: ['-sSL', 'https://pki.goog/roots.pem', '-o', 'launcher/image/google_roots.pem']
+
 # determine the base image
 - name: 'gcr.io/cloud-builders/gcloud'
   id: BaseImageIdent
@@ -110,7 +114,7 @@ steps:
 
 - name: 'gcr.io/cloud-builders/gcloud'
   id: DebugVgImageBuild
-  waitFor: ['BaseImageIdent', 'LauncherCompile', 'DownloadExpBinary', 'DownloadGpuDrivers']
+  waitFor: ['BaseImageIdent', 'LauncherCompile', 'DownloadExpBinary', 'DownloadGpuDrivers', 'DownloadGoogleRoots']
   env:
   - 'OUTPUT_IMAGE_NAME=${_OUTPUT_IMAGE_PREFIX}-vg-debug-${_OUTPUT_IMAGE_SUFFIX}'
   - 'OUTPUT_IMAGE_FAMILY=${_OUTPUT_IMAGE_FAMILY}-vg'
@@ -139,7 +143,7 @@ steps:
 
 - name: 'gcr.io/cloud-builders/gcloud'
   id: HardenedVgImageBuild
-  waitFor: ['BaseImageIdent', 'LauncherCompile', 'DownloadExpBinary', 'DownloadGpuDrivers']
+  waitFor: ['BaseImageIdent', 'LauncherCompile', 'DownloadExpBinary', 'DownloadGpuDrivers', 'DownloadGoogleRoots']
   env:
   - 'OUTPUT_IMAGE_NAME=${_OUTPUT_IMAGE_PREFIX}-vg-hardened-${_OUTPUT_IMAGE_SUFFIX}'
   - 'OUTPUT_IMAGE_FAMILY=${_OUTPUT_IMAGE_FAMILY}-vg'
@@ -169,7 +173,7 @@ steps:
 
 - name: 'gcr.io/cloud-builders/gcloud'
   id: DebugImageBuild
-  waitFor: ['BaseImageIdent', 'LauncherCompile', 'DownloadExpBinary', 'DownloadGpuDrivers']
+  waitFor: ['BaseImageIdent', 'LauncherCompile', 'DownloadExpBinary', 'DownloadGpuDrivers', 'DownloadGoogleRoots']
   env:
   - 'OUTPUT_IMAGE_NAME=${_OUTPUT_IMAGE_PREFIX}-debug-${_OUTPUT_IMAGE_SUFFIX}'
   - 'OUTPUT_IMAGE_FAMILY=${_OUTPUT_IMAGE_FAMILY}'
@@ -198,7 +202,7 @@ steps:
 
 - name: 'gcr.io/cloud-builders/gcloud'
   id: HardenedImageBuild
-  waitFor: ['BaseImageIdent', 'LauncherCompile', 'DownloadExpBinary', 'DownloadGpuDrivers']
+  waitFor: ['BaseImageIdent', 'LauncherCompile', 'DownloadExpBinary', 'DownloadGpuDrivers', 'DownloadGoogleRoots']
   env:
   - 'OUTPUT_IMAGE_NAME=${_OUTPUT_IMAGE_PREFIX}-hardened-${_OUTPUT_IMAGE_SUFFIX}'
   - 'OUTPUT_IMAGE_FAMILY=${_OUTPUT_IMAGE_FAMILY}'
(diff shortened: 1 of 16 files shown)


launcher/image/bc_cloudbuild.yaml
```diff
@@ -65,6 +65,7 @@ steps:
            /workspace/oem-install/confidential_space/
         cp ./launcher/image/nodeproblemdetector/* /workspace/oem-install/confidential_space/nodeproblemdetector/
         cp ./launcher/image/bc-guest/wsd/* /workspace/oem-install/wsd/
+        cp ./launcher/image/google_roots.pem /workspace/oem-install/google_roots.pem
```
yawangwang, line 80: no need to copy google_roots.pem for bc as GCA is not supposed to support it.

launcher/image/vgpreload.sh
```diff
@@ -120,6 +124,8 @@ main() {
   copy_experiment_file
   # Install container launcher.
   copy_launcher
+  # Copy Google Root bundle.
+  copy_google_roots
```
yawangwang, line 128: GCA is not supposed to support VG too.

verifier/util/util.go
```diff
@@ -41,13 +40,7 @@ func PrincipalFetcher(audience string, mdsClient *metadata.Client) ([][]byte, er
...
 // It defaults to the Attestation Verifier instance at
 // https://confidentialcomputing.googleapis.com.
-func NewRESTClient(ctx context.Context, asAddr string, ProjectID string, Region string) (verifier.Client, error) {
-	httpClient, err := google.DefaultClient(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create HTTP client: %v", err)
-	}
-
-	opts := []option.ClientOption{option.WithHTTPClient(httpClient)}
+func NewRESTClient(ctx context.Context, asAddr string, ProjectID string, Region string, opts ...option.ClientOption) (verifier.Client, error) {
```
yawangwang, line 43: Can we make `tls.GoogleRootHTTPClient` as the default option for RESTClient, same for `FetchImpersonatedToken`

launcher/image/bc_cloudbuild.yaml
brongan: bc still makes TLS connections to google owned services.

verifier/util/util.go
brongan: I really like this suggestion. Will do.

launcher/image/vgpreload.sh
brongan: VG makes other calls to Google owned services.

launcher/image/bc_cloudbuild.yaml
yawangwang: cool good catch, but could you add the usage of root pem for BC if it makes TLS connection to those google owned services?

brongan: What do you mean by add?

launcher/cloudbuild.yaml
```diff
@@ -14,6 +14,10 @@ substitutions:
 
 
 steps:
+- name: 'gcr.io/cloud-builders/curl'
```
yawangwang, line 17: If Google's root CA isn't likely to change and has a very long expiration time (over 10 years), I'm thinking of baking roots.pem directly into go-tpm-tools. This way, we don't have to download it during every image build or manually copy it over to the OEM partition. WDYT?

launcher/google_roots.go
```diff
@@ -0,0 +1,61 @@
+// Package launcher provides the entry point and core orchestration logic for the TEE container launcher.
+package launcher
```
yawangwang, line 2: Can we move google_roots.go to verifier/util? I view root CA pinning as the default option when connecting to Google-owned services across all go-tpm-tools modules. This change avoids having to pass GoogleHTTPClient all the way down from the launcher; instead, the GCA and Cloud Logging clients can import it directly.

yawangwang:
Before merging the PR, can we do some e2e tests (or manually verify) to ensure clients only trusting Google CAs can connect to google owned services?

launcher/cloudbuild.yaml
brongan: Pinning roots is not a good idea. These roots regularly change. We do not want all of our customers to be broken when our image breaks one day.

launcher/google_roots.go
brongan: This constructor based dependency injection scheme was chosen intentionally. If GCA and Cloud logging clients import it directly, then their test needs to handle more complex mocking and error handling.
By injecting the HTTP client from the launcher, we can customize the certificate pinning behavior of those modules.

Do you have another dependency injection scheme that you would suggest here?

yawangwang: I would view GoogleHTTPClient as a package-level variable so could you consider using gostub controller for mock testing https://pkg.go.dev/github.com/prashantv/gostub?

launcher/cloudbuild.yaml
yawangwang: well if PKI roots regularly change then downloading during the image build time makes sense.

launcher/google_roots.go
brongan: Thanks for the suggestion! 

I'd prefer to keep `GoogleHTTPClient` as a function and continue passing the client down explicitly as a parameter, rather than converting it to a package-level variable to use `gostub`.

In Go, avoiding mutable global/package-level state is generally preferred because global variables can make concurrent testing (`t.Parallel()`) fragile and prone to race conditions. 

Since the `*http.Client` is already being passed explicitly as a parameter down to `startLauncher` and the runner constructors (dependency injection), we don't actually need to stub any package-level variables for testing. Tests can simply construct their own test/mock HTTP client and pass it directly into those entry points.

yawangwang: Thanks for sharing your thoughts.

I agree that standard Go recommendations favor DI over stubbing. However, my concern is that with so many dependencies already in place, passing the Google HTTP client and its options as separate parameters will ultimately lead to a 'fat constructor' problem and introduce extra complexity to the launcher. Making the Google HTTP client a default option would offer better encapsulation, serving as a trusted client that strictly accepts the Google root CA.

Regarding concurrent testing: since it is enabled for the launcher, we could potentially move `google_roots.go` into a separate 'verifier' module. IMO concurrent testing isn't a dealbreaker for adopting this pattern.

launcher/cloudbuild.yaml
yawangwang: Sorry for the churn, but looking back at this approach, my concern arises from a scenario where the PKI roots might change after the image build. What would our plan be for those 'broken' images? Would we just deprecate them? 

An alternative could be a solution that requires fetching PKI root certs at runtime, but I'd like to defer to @jkl73  for the final decision.

brongan: This PR does not affect the image build PKI root changing situation. Before this change, the roots used for TLS connections made to Google services are pinned by the image's trust bundle (aCOS). After this change, a small subset of these certificates are used when making TLS connections to Google services.

The set of roots used when making connections is pinned in the image in either case. There is no regression.

launcher/google_roots.go
brongan: I completely understand and agree with your desire to avoid the 'fat constructor' problem. I believe the method in question already has 8 parameters without this change. I do not believe this PR is introducing the 'fat constructor' problem.

My suggestion is to fix the fat constructor problem in a separate PR focused on that refactor. What are your thoughts?

yawangwang: sgtm

launcher/cloudbuild.yaml
yawangwang: Just to clarify—your changes aren't causing a regression. Right now, updating the root trust store requires bumping the base COS image version by creating a new image. My main concern is that even with your changes, we’ll still need to build a new image (and deprecate old images if necessary) whenever the PKI root cert changes. Fetching the cert in real time would make the system more resilient here.

That's why I think having Jiankun weigh in would be helpful to help us justify this approach.

brongan: The intention of this change is to scope the trust used in connections to Google services to the minimum required. Concerns surrounding lack of patching or updates to the OS in our images should be considered separately to this PR.

jkl73: we always need to build a new image to "refresh" the cert, whether we relying on COS trust store, or baking roots directly into this repo, or downloading it during building time.

my only concern to this is we're implicitly trusting "gcr.io/cloud-builders/curl" to have the correct cert when talking to pki.goog (but honestly I'm not too worried)

if we baked the certs into the repo, then we need to make sure we remember to update cert frequently (at least once a month?)



Refreshing certs at runtime is another very tricky problem we might want avoid now...

yawangwang (approved):

PR #847, opened 2026-06-19
[launcher] update gpu driver in bc to 595.58.03

jkl73:


yawangwang (approved):

launcher/internal/gpu/config.go
```diff
@@ -5,6 +5,8 @@ const (
 	InstallationHostDir = "/var/lib/nvidia"
 	// BuiltInInstallation590_48_01HostDir is the pre-installed directory.
 	BuiltInInstallation590_48_01HostDir = "/opt/nvidia/590.48.01"
```
stevenrutherford: is this still needed?

stevenrutherford (approved):
Nothing super spicy here iiuc.

launcher/internal/gpu/config.go
jkl73: deleted

PR #848, opened 2026-06-22
update attestationEvidence partial responses

ballesterosa:
Change partial responses to use a protobuf FieldMask to better align with cloud API standards.

- added read_mask to the teeserver.proto
- updated filterVMAttestationFields to use the FieldMask rather than a string
- added some new cases for potential snake case field names along with other names such as readMask, read_mask, $fields, and X-Goog-FieldMask in the HTTP Header instead of Query param to align with [System Parameters Documentation]( https://docs.cloud.google.com/apis/docs/system-parameters)


launcher/teeserver/tee_server.go
```diff
@@ -240,10 +241,35 @@ func (a *attestHandler) getAttestationEvidence(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	fields := r.URL.Query().Get("fields")
+	var mask *fieldmaskpb.FieldMask
```
jessieqliu: nit: Can we put the mask parsing in its own function so it can be united tested independently?

jessieqliu (changes requested):

launcher/teeserver/tee_server.go
```diff
@@ -240,10 +253,22 @@ func (a *attestHandler) getAttestationEvidence(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	fields := r.URL.Query().Get("fields")
+	mask := parseFieldMask(r)
+
+	enableGPU := false
+	if mask != nil {
+		for _, path := range mask.GetPaths() {
+			if path == "*" || path == "deviceReports" || path == "device_reports" ||
```
mayafleischer: This adds support for a requesting a subfield of device reports (i.e. device_reports.<something>) but when we filter the attestation, we only look for exactly "deviceReports".  I think either way is fine but we should be consistent between the two.

launcher/teeserver/tee_server.go
```diff
@@ -268,15 +293,23 @@ func (a *attestHandler) getAttestationEvidence(w http.ResponseWriter, r *http.Re
...
+		if path == "*" {
+			return att, nil
+		}
+	}
+
 	fieldMap := make(map[string]bool)
-	for _, f := range fieldSlice {
-		fieldMap[strings.TrimSpace(f)] = true
+	for _, path := range mask.GetPaths() {
+		fieldMap[strings.TrimSpace(path)] = true
```
mayafleischer, line 312: Similar to my comment above - whitespace is trimmed here, but not in the earlier filtering.

mayafleischer (changes requested):

mayafleischer (approved):

jessieqliu (approved):

PR #849, opened 2026-06-22
Best practices - prevent deadlock on runSystemCmd

lindsayzad:
Add 30 second timeout to launcher


jkl73 (approved):

launcher/internal/systemctl/systemctl.go
yawangwang, line 71: Shall we pass the `ctx` to the cmdFunc rather than creating a new context here? This would guarantee that a hang in either D-Bus or systemd unit activation will be caught and aborted within the 30-second deadline.

PR #852, opened 2026-06-26
Fix wrong test dependency in the log redirection test

susienme:
Fixes a typo in the log redirection test configuration where a cleanup step waited for the same verification step twice instead of waiting for both parallel verification steps. This led to premature VM deletion before the serial log check could complete.


Sibcgh (approved):
lgtm

PR #855, opened 2026-07-07
launcher: refactor NewRunner to use a RunnerConfig struct

brongan:
Simplify the `NewRunner` constructor by grouping its 10 parameters into a `RunnerConfig` struct. This resolves the "fat constructor" code smell and makes it easier to pass optional configuration parameters in the future.

The fields of `RunnerConfig` are mapped to local variables at the beginning of `NewRunner` to keep the diff minimal and avoid churning the rest of the function body.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -95,8 +95,31 @@ const (
 // Default OOM score for a CS container.
 const defaultOOMScore = 1000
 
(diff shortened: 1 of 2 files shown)


brongan:
This is a requested follow to https://github.com/google/go-tpm-tools/pull/825

yawangwang (approved):

PR #863, opened 2026-07-09
launcher: dependency inject TPM and mount verifier

brongan:
### Rationale
The launcher's boot pipeline (`launcher/launcher/main.go`) was previously untestable because it relied on hardcoded system calls—specifically, opening the physical TPM device (`/dev/tpmrm0`) and executing host CLI binaries (`dmsetup`, `findmnt`, `cryptsetup`) to verify partitions. 

By dependency injecting these resources as interfaces, we can now pass lightweight mock implementations in unit tests, allowing us to test the startup and verification logic in isolation.

### Changes
*   **TPM Handle Injection**: Moved the `tpm2.OpenTPM` call directly to `main()`. The opened `io.ReadWriteCloser` is now passed as a parameter to `initTPM` and `startLauncher`, allowing tests to easily inject a fake TPM.
*   **Mount Verifier Injection**: Defined a `mountVerifier` interface to abstract the device mapper and mount queries. Refactored `verifyFsAndMount` to accept this interface, using `osMountVerifier` in production and enabling mock verifiers in tests.

### DIFF
--- a/launcher/launcher/main.go
+++ b/launcher/launcher/main.go
@@ -131,11 +131,37 @@ func main() {
 		return
 	}
 
-	if err := verifyFsAndMount(launchSpec); err != nil {
-		logger.Error(fmt.Sprintf("failed to verify filesystem and mounts: %v\n", err))
+	verifier := osMountVerifier{}
+	if err := verifyDiskIntegrity(verifier); err != nil {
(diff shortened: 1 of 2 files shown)


Sibcgh (approved):
LGTM!

PR #865, opened 2026-07-09
launcher: enable unit testing for ContainerRunner creation

brongan:
NewRunner was previously untestable because it directly depended on a live `containerd` daemon socket and hardware-specific GCE TPM NVRAM indices. Introduce a mockable `containerdClient` interface and make the Attestation Key fetcher configurable. This allows testing runner configuration logic (such as environment variable log redaction and launch policies) locally using a TPM  simulator.


yawangwang:
LGTM contingent on successful cloud build runs

yawangwang (approved):

PR #866, opened 2026-07-09
launcher: refactor monolithic NewRunner initialization into StartLauncher

brongan:
To make the runner testable in isolation, this change moves resource acquisition (TPM handles, registry authentication, containerd client connections, and verifier clients) out of the library code (NewRunner) and into the orchestration layer (StartLauncher). This allows us to inject mock or fake resources in unit tests instead of requiring real hardware and network resources during construction.


Sibcgh (approved):
LGTM, thanks for doing this!

PR #868, opened 2026-07-10
ITA nvgpu support

ballesterosa:
Add changes related to supporting GPU attestation through ITA:

- creates `gpuDeviceEvidence` and `nvgpuEvidence` structs for the `tokenRequest` struct
- adds logic in `client.go` to convert the `NvidiaAttestationReport` struct to `nvgpuEvidence` for the ITA specific `tokenRequest`


verifier/ita/client.go
```diff
@@ -258,9 +259,52 @@ func convertRequestToTokenRequest(request verifier.VerifyAttestationRequest) tok
...
 	}
 
+	if request.NvidiaAttestation != nil {
+		tokenReq.Nvgpu = convertNvidiaAttestationToITANvgpu(request.NvidiaAttestation)
+	}
+
 	return tokenReq
 }
 
+func convertNvidiaAttestationToITANvgpu(nvAtt *attestationpb.NvidiaAttestationReport) *nvgpuEvidence {
```
jessieqliu: nit: `convertNvidiaAttestationToNvgpu` since this is already in the ITA package/namespace, having it in the function is unnecessary repetition.

jessieqliu (commented):
LGTM for readability/basic correctness, deferring to Yawang for domain-specific review

verifier/ita/client.go
ballesterosa: fixed.

verifier/ita/evidence.go
```diff
@@ -49,11 +49,23 @@ type gcpData struct {
 	CSInfo            confidentialSpaceInfo `json:"confidential_space_info"`
 }
 
+type gpuDeviceEvidence struct {
+	Evidence    []byte `json:"evidence"`
+	Certificate []byte `json:"certificate"`
```
yawangwang, line 54: just for my understanding: from the API schema doc shared by Intel, looks like these two fields are expected to be base64 encoded string, is the raw byte format acceptable for ITA as well?

agent/agent.go
```diff
@@ -359,7 +361,7 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 		return nil, fmt.Errorf("received an unsupported attestation type! %v", v)
 	}
 
-	if a.experiments.EnableGpuGcaSupport {
+	if a.experiments.EnableGpuGcaSupport || a.experiments.EnableGpuItaSupport {
```
yawangwang: We need to be careful about the conditional check for GpuItaSupport - as GpuGcaSupport already goes GA, `a.experiments.EnableGpuGcaSupport` will always return true. So this condition check will always return true even if ITA support is OFF. Some strict checks should be added here to distinguish GpuGcaSupport vs GpuItaSupport.

verifier/ita/evidence.go
ballesterosa: Yep, when the []byte type gets marshaled, it becomes a base64 encoded string and ITA accepted the evidence when sent with this type in the struct in Go.

yawangwang:
Don't forget to add an ITA network connection test to the image tests! Since we already have one for GCA, we should make sure ITA (as a separate verifier service) is covered too. See example in https://github.com/google/go-tpm-tools/pull/895.

agent/agent.go
ballesterosa: Added a check for whether the Challenge starts with "ita//" which indicates which experiment flag is relevant: https://github.com/google/go-tpm-tools/pull/868/changes/0a92e00354323c139590b9e437fddb75a2b32612#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8bR366

launcher/scratch_test/test_ita_agent.go
```diff
@@ -0,0 +1,160 @@
+// scratch_test/test_ita_agent.go:go test -run TestITA ./launcher/scratch_test
```
yawangwang: I see the value of this tests as it's useful for manual integration tests, but it's not ideal for production code because it's not integrated into CI/CD pipeline and relies on ITA's tricky changes in their dev env.

ballesterosa: removed the test

yawangwang (approved):

PR #870, opened 2026-07-10
Allow fluentbit forwarding

mayafleischer:
Create a dedicated fluent-bit config file to accept forwarded logs and enable additional collection and filtration.

Key changes:

- [new] bc-fluent-bit.config: 
Collect and filter kernel, network logs and WSD
Configure fluent-bit forward receiver

- bc-entrypoint.sh
Update `iptables` firewall rules to allow incoming TCP port 24224 traffic exclusively from the KPS vm.
Update to use newly created config file

Additional filtering logic is expected to be added.
See https://github.com/GoogleCloudPlatform/key-protection-module/pull/64 for corresponding KPS VM change.

### DIFF
--- a/launcher/image/bc-guest/bc-fluent-bit-cs.conf
+++ b/launcher/image/bc-guest/bc-fluent-bit-cs.conf
@@ -69,10 +69,21 @@
     Match         kernel-logs
     Regex         msg NVRM:\s*Xid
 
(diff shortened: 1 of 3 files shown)


atulpatildbz (approved):
Thanks for taking up this implementation

jkl73 (approved):

PR #872, opened 2026-07-13
Unpack guest-gpu-tools:latest onto OEM partition

rkolchmeyer:
Extends read-only security features to guest-gpu-tools container.

Test: after build and boot, saw
`[ WorkloadVM ] Fabric readiness completed for all GPUs.` in logs.


rkolchmeyer:
@jkl73 PTAL, thanks!

jkl73 (approved):

PR #879, opened 2026-07-17
launcher: extract OCI spec options construction for unit testing

brongan:
Extracts the OCI container specification options builder from NewRunner into a standalone, testable function `createOCISpecOpts` in container_spec.go. This allows us to verify container options (envs, mounts, limits, cgroups, GPUs) in unit tests by applying them in-memory to a blank spec.

### DIFF
--- a/launcher/container_spec.go
+++ b/launcher/container_spec.go
@@ -1,0 +1,171 @@
+package launcher
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/containers"
+	"github.com/containerd/containerd/oci"
+	"github.com/google/go-tpm-tools/cel"
+	"github.com/google/go-tpm-tools/launcher/internal/gpu"
+	"github.com/google/go-tpm-tools/launcher/internal/logging"
+	"github.com/google/go-tpm-tools/launcher/launcherfile"
+	"github.com/google/go-tpm-tools/launcher/spec"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func createOCISpecOpts(image containerd.Image, launchSpec spec.LaunchSpec, envs []string, listFiles func(string, string) ([]string, error), logger logging.Logger) ([]oci.SpecOpts, error) {
+	var mounts []specs.Mount
+	for _, lsMnt := range launchSpec.Mounts {
+		mounts = append(mounts, lsMnt.SpecsMount())
+	}
+	mounts = appendTokenMounts(mounts)
+	if launchSpec.CgroupNamespace {
+		mounts = appendCgroupRw(mounts)
+	}
+	hostname, err := os.Hostname()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


launcher/container_spec.go
```diff
@@ -0,0 +1,162 @@
...
+					Source:      fmt.Sprintf("%s/bin", gpu.BuiltInInstallation595_58_03HostDir),
+					Destination: fmt.Sprintf("%s/bin", gpu.InstallationContainerDir),
+					Options:     []string{"rbind", "rw"},
+				},
+			}
+		}
+
+		specOpts = append(specOpts, oci.WithMounts(gpuMounts))
+
+		gpuDeviceFiles, err := listFiles("/dev", "nvidia")
```
Sibcgh, line 107: Can we still keep the following comment above this gpuDeviceFiles line?

```
		// /dev/nvidia-caps/* will not be listed here and will not be passed to
		// the container workload
		//
		// following devices should be listed:
		// /dev/nvidiactl
		// /dev/nvidia-uvm
		// /dev/nvidia-uvm-tools
		// /dev/nvidia{0,1,2,...}
		// /dev/nvidia-modeset
```

brongan: That shouldn't have been dropped. Brought back. Thanks for catching this!

launcher/container_spec_test.go
```diff
@@ -0,0 +1,328 @@
...
+				}
+			}
+			if foundCgroupNS != tc.wantCgroupNS {
+				t.Errorf("Cgroup namespace presence mismatch: got %v, want %v", foundCgroupNS, tc.wantCgroupNS)
+			}
+		})
+	}
+}
+
+func TestCreateOCISpecOpts_GPU(t *testing.T) {
```
Sibcgh, line 303: Could we also add unit test coverage for GPU BC mode (`Experiments.BcMode = true`)?

Confidential Space supports two GPU driver modes:
1. Dynamic Mode: Downloads and mounts drivers at boot time into `/usr/local/nvidia`.
2. BC Mode: Uses pre-baked driver libraries (`595.58.03`) directly on the OEM partition for the BC project.

I think we could also add GPU Device List Error Handling (`listFiles` error propagation). If the host system fails to list GPU devices (e.g., `/dev/nvidia0` device nodes are corrupt or missing kernel drivers), `createOCISpecOpts` must fail fast.

launcher/container_spec.go
Sibcgh, line 1: @yawangwang are there any other potential test cases we could be missing?

Sibcgh (changes requested):
Small changes, I think there's just a few more tests we could add in.

launcher/container_spec.go
brongan: These are change detector tests that aren't adding much value.

launcher/container_spec_test.go
brongan: Done I think

Sibcgh (approved):

susienme:
`deviceROTs` is not used. Did we forget anything?

brongan:
@susienme Some duplicate code was kept from an improperly handled rebase. Removed it.

PR #883, opened 2026-07-20
Update TPM simulator to track new version

ballesterosa:
The Google fork of the TPM reference implementation is tracked here now: https://github.com/google/TPM/tree/main
and the actual implementation is no longer hosted by Microsoft, instead by the TCG (Trusted Computing Group) here: https://github.com/TrustedComputingGroup/TPM

This change refactors `simulator.go` to reference the Go bindings at https://github.com/google/TPM/tree/bindings.


jkl73:
Ideally can we should just remove the simulator package from this repo entirely and depends on https://github.com/google/TPM for the simualtor @josephlr wdyt? is this doable now?

ballesterosa:
> Ideally can we should just remove the simulator package from this repo entirely and depends on https://github.com/google/TPM for the simualtor @josephlr wdyt? is this doable now?

I think this is also doable, we just need to change the bindings and add some code to simulator.go

PR #884, opened 2026-07-20
Add 128GB of hugetlbfs, in respone to customer request.

gjacobwells:


mayafleischer (approved):

mayafleischer (changes requested):
KeymanagerTests (vg) failing with this change.  I think this test runs on n2d-standard-2 so it makes sense and I don't think indicates any real issue but we will need to ensure the presubmit pass consistently.

mayafleischer (approved):
LGTM with successful cloudbuild

PR #894, opened 2026-07-23
Restore logging the launch spec used by image tests.

susienme:
Re-adds the launch spec logging. This log output is required by image integration tests (such as test_experiment_value.sh) that verify experiment values from the VM serial console.

### DIFF
--- a/launcher/launcher/main.go
+++ b/launcher/launcher/main.go
@@ -114,6 +114,8 @@ func main() {
 		logger.Error(exitMessage, "exit_code", exitCode, "exit_msg", rcMessage[exitCode])
 		return
 	}
+	// Do not delete the folliwing line, this line is used for tests.
+	logger.Info(fmt.Sprintf("Launch Spec: %+v", launchSpec.LogFriendly()))
 
 	verifier := osMountVerifier{}
 	if err := verifyDiskIntegrity(verifier); err != nil {


brongan:
Is the test that asserts the line was logged in `/gcbrun`?

launcher/launcher/main.go
```diff
@@ -114,6 +114,8 @@ func main() {
 		logger.Error(exitMessage, "exit_code", exitCode, "exit_msg", rcMessage[exitCode])
 		return
 	}
+	// Do not delete the folliwing line, this line is used for tests.
```
brongan, line 117: I'm used to comments like this not being necessary because their removal causes tests to fail. Any PR that would remove these should have its tests fail. Is that not the case here?

susienme:
> Is the test that asserts the line was logged in `/gcbrun`?

Yes, `12: ExperimentsTests`

launcher/launcher/main.go
susienme: You are correct that the tests fail. However, this exact line was deleted in a previous PR where the test failure was overlooked and the PR was merged anyway. I added this explicit comment as an extra safeguard to ensure it doesn't happen again.

brongan: Automerge ignoring a failing gcbrun doesn't seem ideal

brongan (approved):
Sorry about breaking this!

PR #895, opened 2026-07-23
launcher: pass googleClient explicitly to StartLauncher and createAttestClients

brongan:
Confidential Space launcher failed to start in production environments with
HTTP 401 (`CREDENTIALS_MISSING`) errors when initializing the GCA REST verifier client:

  "failed to create REST verifier client: listing regions in project ...:
   googleapi: Error 401: Request is missing required authentication credential."

This regression occurred when verifier client creation was refactored out of
`ContainerRunner` into `StartLauncher`. `StartLauncher` initialized `googleClient`
(which contains root CA certificates and HTTP transport settings) but invoked
`createAttestClients` with raw `clientOpts`, omitting `option.WithHTTPClient(googleClient)`.
Consequently, `util.NewRESTClient` fell back to unauthenticated transport when calling
`ListLocations`. Presubmit integration tests (`/gcbrun`) failed but the PR was merged anyway.

To resolve this and prevent future regressions:
1. `StartLauncher` prepends `option.WithHTTPClient(googleClient)` to `clientOpts`
   before calling `createAttestClients`.
2. `createAttestClients` verifies that `clientOpts` includes an `option.WithHTTPClient`
   transport before attempting REST client creation when `DisableGcaRefresh` is false.
3. `launcher/start_test.go` adds a hermetic table-driven test suite covering all
   verifier client creation paths.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -303,6 +303,7 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 	if err != nil {
 		return nil, err
 	}
+	a.logger.Info("Successfully created challenge", "challenge_name", challenge.Name)
 
 	tokenOpts := opts.TokenOptions
 	if tokenOpts == nil {
--- a/launcher/google_roots.go
+++ b/launcher/google_roots.go
@@ -12,16 +12,10 @@ import (
 // GoogleRootsPath is the path to the Google roots PEM file on the OEM partition.
 const GoogleRootsPath = "/usr/share/oem/google_roots.pem"
 
-// GoogleHTTPClient creates an HTTP client that only trusts the roots required for connecting to Google.
-func GoogleHTTPClient() (*http.Client, error) {
-	return googleHTTPClientWithRoots(GoogleRootsPath)
-}
-
-// googleHTTPClientWithRoots allows internal tests to inject mock certificate paths.
-func googleHTTPClientWithRoots(rootsPath string) (*http.Client, error) {
-	pool, err := loadCertPool(rootsPath)
-	if err != nil {
-		return nil, err
+// PinnedHTTPTransport creates an HTTP transport configured with the provided root certificate pool.
+func PinnedHTTPTransport(pool *x509.CertPool) (http.RoundTripper, error) {
+	if pool == nil {
+		return nil, fmt.Errorf("cert pool must be non-nil")
 	}
 
 	// We copy the default transport so we get all the proxy, keep-alive,
@@ -37,9 +31,7 @@ func googleHTTPClientWithRoots(rootsPath string) (*http.Client, error) {
 	}
 	customTransport.TLSClientConfig.RootCAs = pool
 
-	return &http.Client{
-		Transport: customTransport,
-	}, nil
+	return customTransport, nil
 }
 
 // GoogleCertPool loads the Google root certificates into an x509.CertPool.
--- a/launcher/image/test/scripts/test_real_gca_workload.sh
+++ b/launcher/image/test/scripts/test_real_gca_workload.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+source util/read_serial.sh
+
+SERIAL_OUTPUT=$(read_serial $1 $2)
+print_serial=false
+
+if echo "$SERIAL_OUTPUT" | grep -q 'Successfully created challenge'
+then
+    echo "- challenge creation verified"
+else
+    echo "FAILED: challenge creation failed (GCA client creation or VM startup failed)"
+    echo 'TEST FAILED.' > /workspace/status.txt
+    print_serial=true
+fi
+
+if $print_serial; then
+    echo "$SERIAL_OUTPUT"
+fi
--- a/launcher/image/test/test_real_gca_cloudbuild.yaml
+++ b/launcher/image/test/test_real_gca_cloudbuild.yaml
@@ -1,0 +1,36 @@
+substitutions:
+  '_IMAGE_NAME': ''
+  '_IMAGE_PROJECT': ''
+  '_CLEANUP': 'true'
+  '_VM_NAME_PREFIX': 'cs-real-gca-test'
+  '_ZONE': 'us-west1-a'
+  '_CC': ''
+  '_WORKLOAD_IMAGE': 'us-west1-docker.pkg.dev/confidential-space-images-dev/cs-integ-test-images/basic-test:latest'
+steps:
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: CreateVM
+  entrypoint: 'bash'
+  env:
+  - 'BUILD_ID=$BUILD_ID'
+  args: ['create_vm.sh','-i', '${_IMAGE_NAME}',
+          '-p', '${_IMAGE_PROJECT}',
+          '-m', 'tee-image-reference=${_WORKLOAD_IMAGE},tee-container-log-redirect=true',
+          '-n', '${_VM_NAME_PREFIX}-${BUILD_ID}',
+          '-z', '${_ZONE}',
+          '-c', '${_CC}',
+          '-v', 'false',
+        ]
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: RealGCAWorkloadTest
+  entrypoint: 'bash'
+  args: ['scripts/test_real_gca_workload.sh', '${_VM_NAME_PREFIX}-${BUILD_ID}', '${_ZONE}']
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: CleanUp
+  entrypoint: 'bash'
+  env:
+  - 'CLEANUP=${_CLEANUP}'
+  args: ['cleanup.sh', '${_VM_NAME_PREFIX}-${BUILD_ID}', '${_ZONE}']
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: CheckFailure
+  entrypoint: 'bash'
+  args: ['check_failure.sh']
--- a/launcher/launcher/main.go
+++ b/launcher/launcher/main.go
@@ -7,18 +7,19 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net/http"
 	"os"
 	"os/exec"
 	"regexp"
 	"strings"
 	"time"
 
+	"cloud.google.com/go/auth/httptransport"
 	"cloud.google.com/go/compute/metadata"
 	"github.com/google/go-tpm-tools/launcher"
 	"github.com/google/go-tpm-tools/launcher/internal/logging"
 	"github.com/google/go-tpm-tools/launcher/launcherfile"
 	"github.com/google/go-tpm-tools/launcher/spec"
-	"google.golang.org/api/option"
 )
 
 const (
@@ -60,23 +61,6 @@ func main() {
 		os.Exit(exitCode)
 	}()
 
-	googleClient, err := launcher.GoogleHTTPClient()
-	if err != nil {
-		log.Default().Printf("Failed to initialize Google root HTTP client: %v", err)
-		exitCode = failRC
-		log.Default().Printf("%s, exit code: %d (%s)\n", exitMessage, exitCode, rcMessage[exitCode])
-		return
-	}
-	clientOpts := []option.ClientOption{option.WithHTTPClient(googleClient)}
-
-	pool, err := launcher.GoogleCertPool()
-	if err != nil {
-		log.Default().Printf("Failed to load Google root certificates: %v", err)
-		exitCode = failRC
-		log.Default().Printf("%s, exit code: %d (%s)\n", exitMessage, exitCode, rcMessage[exitCode])
-		return
-	}
-
 	serialConsole, err := os.OpenFile(serialConsoleFile, os.O_WRONLY, 0)
 	if err != nil {
 		log.Default().Printf("Failed to open serial console: %v", err)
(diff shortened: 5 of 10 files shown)


launcher/start.go
```diff
@@ -75,17 +78,18 @@ func StartLauncher(ctx context.Context, launchSpec spec.LaunchSpec, logger loggi
 		}
 	}
 
-	googleClient, err := GoogleHTTPClient()
+	pinnedClient, err := GooglePinnedHTTPClient()
```
yawangwang: Could you add brief comments here explaining usages of pinnedClient vs googleClient?

yawangwang:
As what we discussed in the sync, could you add image e2e tests to launcher/image/test to verify the actual network connection can be successfully setup?

brongan:
> As what we discussed in the sync, could you add image e2e tests to launcher/image/test to verify the actual network connection can be successfully setup?

The tests in launcher/image/test already verify the actual network connection can be successfully setup. No change is necessary beyond the branch protection that you added.

yawangwang:
> > As what we discussed in the sync, could you add image e2e tests to launcher/image/test to verify the actual network connection can be successfully setup?
> 
> The tests in launcher/image/test already verify the actual network connection can be successfully setup. No change is necessary beyond the branch protection that you added.

Could you point me to which tests verify the network connection? Looks like we disabled the gca connection in here https://github.com/google/go-tpm-tools/blob/349338c4a6d845ee9c240322931435d960d8c7e5/launcher/image/test/create_vm.sh#L39.

brongan:
> > > As what we discussed in the sync, could you add image e2e tests to launcher/image/test to verify the actual network connection can be successfully setup?
> > 
> > 
> > The tests in launcher/image/test already verify the actual network connection can be successfully setup. No change is necessary beyond the branch protection that you added.
> 
> Could you point me to which tests verify the network connection? Looks like we disabled the gca connection in here
> 
> https://github.com/google/go-tpm-tools/blob/349338c4a6d845ee9c240322931435d960d8c7e5/launcher/image/test/create_vm.sh#L39
> 
> .

I am mistaken

launcher/start.go
brongan: Done!

brongan:
@yawangwang I don't think we should reintroduce real REST client(s) in the e2e tests until we mitigate or resolve the underlying issues that caused us to implement #607 in the first place. What are your thoughts?

yawangwang:
> @yawangwang I don't think we should reintroduce real REST client(s) in the e2e tests until we mitigate or resolve the underlying issues that caused us to implement #607 in the first place. What are your thoughts?

We intentionally implemented #607 because most of our tests evaluate CS image behavior—such as logging and memory monitoring—which doesn't require GCA. However, since the pinnedGoogleClient changes how the launcher interacts with GCA, adding a network connection test would be ideal to close this testing gap.

I recommend introducing an option to override the fake verifier specifically for network connection tests, while leaving the remaining tests untouched.

brongan:
Looking through historical bugs, I believe that the fake verifier was added to avoid quota issues and other things involving remote service dependencies in the client end-to-end tests. Is that not accurate?

brongan:
@yawangwang 
I've updated `createAttestClients` so that `validateAttestHTTPClient(googleClient)` executes unconditionally at the beginning of client initialization (before checking `FakeVerifierEnabled`).

Because every launcher startup in our E2E integration test suite runs through `StartLauncher()` -> `createAttestClients()`, **all 17 existing E2E tests now automatically exercise the assertion that `googleClient` has valid OAuth2 token credentials and pinned Google Root CAs.**

This ensures our HTTP client authentication and certificate-pinning validation are fully covered across all CI runs without requiring live network calls to `confidentialcomputing.googleapis.com`.

yawangwang:
> @yawangwang I've updated `createAttestClients` so that `validateAttestHTTPClient(googleClient)` executes unconditionally at the beginning of client initialization (before checking `FakeVerifierEnabled`).
> 
> Because every launcher startup in our E2E integration test suite runs through `StartLauncher()` -> `createAttestClients()`, **all 17 existing E2E tests now automatically exercise the assertion that `googleClient` has valid OAuth2 token credentials and pinned Google Root CAs.**
> 
> This ensures our HTTP client authentication and certificate-pinning validation are fully covered across all CI runs without requiring live network calls to `confidentialcomputing.googleapis.com`.

My understanding is that you are testing the exact code path executed to bypass the live network test, which seems like a viable approach for this specific fix. However, the takeaway from our recent image release blocker is that we need a more robust/reliable presubmit system. We need to capture image bugs that might be overlooked by engineers or missed by unit tests. Having an e2e network connection test closes this gap; it not only validates your fix but also acts as a guardrail in our presubmit pipeline to detect unintended client configuration changes or future GCA issues.

Looking back at the original PR (see [PR #825, comment](https://www.google.com/url?q=https%3A%2F%2Fgithub.com%2Fgoogle%2Fgo-tpm-tools%2Fpull%2F825%23issuecomment-4747893548)), a live network test was mentioned but that point was somehow lost in the long discussions around root pinning. I believe we should re-emphasize the importance of implementing an image e2e test.

Please let me know if you have any concerns about adding an image e2e test.

launcher/google_roots.go
susienme, line 42: I noticed this function is called from multiple paths, causing the PEM file to be read and parsed each time. To avoid redundant I/O and parsing, please consider using init() or sync.Once() to cache and reuse the parsed pool. If you decide to go with init(), I think it is totally fine to panic on an error, since the launcher can't operate without these root certs anyway.

launcher/start.go
```diff
@@ -214,7 +266,7 @@ func createAttestClients(ctx context.Context, launchSpec spec.LaunchSpec, logger
 		}
 		attestClients.ITA = itaClient
```
susienme, line 228: Consider returning from here to keep the happy path less indented (go/gotip/episodes/1).

launcher/start_test.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+func TestCreateAttestClients_Behaviors(t *testing.T) {
+	mockOauth2Server, err := util.NewMockOauth2Server()
+	if err != nil {
+		t.Fatalf("failed to create mock OAuth2 server: %v", err)
+	}
+	defer mockOauth2Server.Stop()
+
+	google.Endpoint = oauth2.Endpoint{
```
susienme: This modifies the global `google.Endpoint`. To prevent unintended side effects on tests added in the future, consider saving the original state and restoring it.

brongan:
I suggest that an image e2e test is more appropriate for a separate PR. The purpose of this PR is to fix the regression. Adding presubmit test coverage so that it does not happen again is valuable. There are many choices on how much image behavior and GCA compatibility is covered.

yawangwang:
> I suggest that an image e2e test is more appropriate for a separate PR. The purpose of this PR is to fix the regression. Adding presubmit test coverage so that it does not happen again is valuable. There are many choices on how much image behavior and GCA compatibility is covered.

If adding image e2e tests would take much time and would delay the image release, I am fine with deferring them to a follow-up PR. However, please ensure we manually verify that the live network connection works (e.g., by running a CVM using the presubmit image) before merging this PR. I want to avoid another situation where a buggy image might block our release.

mayafleischer:
+1 to Yawang's comment.  Ideally we merge this before OED so that the rollout that starts on Monday can include the new image.

launcher/google_roots.go
brongan: Shuffled construction around to avoid redundant I/O and parsing. Please let me know what you think!

launcher/start.go
brongan: Great suggestion! I've improved the situation a bit here. Please let me know what you think.

launcher/start_test.go
brongan: Great catch! I've refactored this to avoid mutating package variables while maintaining test coverage.

brongan:
Added an e2e test.

launcher/start_test.go
```diff
@@ -0,0 +1,176 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			logger := &fakeLogger{}
+			clients, err := createAttestClients(t.Context(), tc.launchSpec, logger, tc.googleClient)
+
+			if (err != nil) != tc.wantErr {
+				t.Fatalf("createAttestClients() error = %v, wantErr %v", err, tc.wantErr)
+			}
+			if tc.wantErr && tc.errContains != "" {
+				if err == nil || !strings.Contains(err.Error(), tc.errContains) {
+					t.Fatalf("expected error containing %q, got %v", tc.errContains, err)
```
susienme: "Got before want" go/go-style/decisions#got-before-want

launcher/google_roots.go
susienme: Thanks for the update! While hoisting the call to `main.go` solves the redundant I/O for the current implementation, it relies on the caller to manage the state. If `GoogleCertPool()` is called from a new path in the future, we will hit the same issue. To make this more robust, consider caching the parsed pool on the callee side so any future callers automatically get the cached version.

susienme (approved):

launcher/image/test/test_real_gca_cloudbuild.yaml
```diff
@@ -0,0 +1,36 @@
...
+          '-m', 'tee-image-reference=${_WORKLOAD_IMAGE},tee-container-log-redirect=true',
+          '-n', '${_VM_NAME_PREFIX}-${BUILD_ID}',
+          '-z', '${_ZONE}',
+          '-c', '${_CC}',
+          '-v', 'false',
+        ]
+- name: 'gcr.io/cloud-builders/gcloud'
+  id: RealGCAWorkloadTest
+  entrypoint: 'bash'
+  args: ['scripts/test_real_gca_workload.sh', '${_VM_NAME_PREFIX}-${BUILD_ID}', '${_ZONE}']
```
yawangwang, line 26: I appreciate your efforts of writing image e2e tests, but looks like this test is not integrated into the presubmit cloudbuild run `launcher/cloudbuild.yaml`

launcher/image/test/scripts/test_real_gca_workload.sh
```diff
@@ -0,0 +1,46 @@
...
+if echo $SERIAL_OUTPUT | grep -q 'iss: fake-issuer-for-testing'
+then
+    echo "FAILED: token issued by fake verifier instead of real GCA"
+    echo 'TEST FAILED.' > /workspace/status.txt
+    print_serial=true
+else
+    echo "- verified token was NOT issued by fake verifier"
+fi
+
+if echo $SERIAL_OUTPUT | grep -q 'iss: https://confidentialcomputing.googleapis.com'
```
yawangwang: I doubt this check will work since GCA will NOT return a successful token as this image is not added to image db.

I'd suggest simply doing a logging after googleClient is successfully connected to GCA, then searching for that log in this test.

yawangwang (changes requested):

launcher/image/test/scripts/test_real_gca_workload.sh
brongan: Simplified the test.

launcher/image/test/test_real_gca_cloudbuild.yaml
brongan: Wow I was really hoping tests were automatically enumerated. Thanks for catching this. Added

launcher/start_test.go
brongan: Thanks! Fixed

agent/agent.go
```diff
@@ -303,6 +303,7 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 	if err != nil {
 		return nil, err
 	}
+	a.logger.Info("Successfully created challenge", "challenge_name", challenge.Name)
```
susienme, line 306: Just to confirm my understanding: since the new test sets `test-fake-verifier=false`, if the launcher fails to connect to the GCA, `client.CreateChallenge()` will return an error and this success message won't be printed. Is that correct?

launcher/image/test/scripts/test_real_gca_workload.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+source util/read_serial.sh
+
+SERIAL_OUTPUT=$(read_serial $1 $2)
+print_serial=false
+
+if echo "$SERIAL_OUTPUT" | grep -q 'Successfully created challenge'
```
susienme, line 8: Maybe appending `| grep -qv 'projects/fakeProject'` to ensure the fake verifier is not used?

launcher/google_roots.go
brongan: I think that if we are calling this twice then we are doing something wrong. I am opposed to hidden caching behavior. Are there any alternatives that you would accept?

agent/agent.go
yawangwang: "Successfully created challenge" should just be fine, we don't want to log the exact nonce issued by GCA unless necessary.

launcher/launcher/main.go
```diff
@@ -60,42 +61,55 @@ func main() {
...
 		return
 	}
-	defer workloadLogger.Close()
 
-	logger := logging.DualLogger(workloadLogger, logging.NewSerialLogger(serialConsole))
+	pinnedClient := &http.Client{Transport: pinnedTransport}
+
+	googleClient, err := httptransport.NewClient(&httptransport.Options{
+		BaseRoundTripper: pinnedTransport,
+	})
```
yawangwang, line 106: I noticed some major changes have been made compared to the previous commit, so just for my understanding - httptransport.NewClient will automatically attach ADC to the client?

launcher/image/google_roots.pem
yawangwang, line 1: what is this file used for? I thought google root certs will be fetched at image build time?

yawangwang (approved):

launcher/start.go
brongan: The `pinnedClient` configures the bare TLS transport to strictly verify against the embedded Google Root Certificate Authorities. The `googleClient` is a standard GCP HTTP client that automatically discovers and attaches Application Default Credentials (ADC) for our regular resource access. The `googleClient` is built on top of the pinned transport.

launcher/google_roots.go
brongan: I'm going to decline this change, because hiding the I/O behind a cache makes the behavior less predictable. If we ever find ourselves initializing this pool from multiple parallel paths, there's a deeper architectural problem we should fix rather than just paving over it with `sync.Once`. I want to keep this code explicit and avoid adding unnecessary complexity.

launcher/launcher/main.go
brongan: Yes, `httptransport.NewClient` automatically discovers and attaches Application Default Credentials (ADC) to the client connection.

launcher/image/google_roots.pem
brongan: That file was added completely by mistake and was not meant to be tracked in the repository. It has been removed in the latest commit.

agent/agent.go
brongan: I believe `challenge.Name` is purely the GCP resource identifier (e.g., `projects/.../challenges/...`), and doesn't actually log the sensitive nonce bytes itself; those should be stored in `challenge.Nonce`. So it should be safe to leave the name in the log!

launcher/image/test/scripts/test_real_gca_workload.sh
brongan: I'd prefer not to use log grepping to ensure the fake verifier isn't used, as it's a bit brittle. We're already explicitly preventing it via the `-test-fake-verifier=false` configuration. The better long-term solution to this is to structure our dependencies so we don't link the fakes into the production binaries at all.

PR #899, opened 2026-07-23
Remove nvlsm/fabricmanager

mayafleischer:
Moving nvlsm and fabric manager to the host for BC.


mayafleischer:
If we proceed with this change/strategy there is additional gpu installation/set up that should be removed.

Sibcgh:
LGTM, and we can easily revert back if something fails.

Sibcgh (approved):

PR #902, opened 2026-07-26
client: add `FindHandle` to look up a persistent key by its public key

### LINKED ISSUE #901: client: add a way to find a persistent key by its public key
`client.LoadCachedKey` requires the caller to know the handle a key sits at. That handle is set by whoever provisioned the key and can differ between machines, so applications end up carrying it in config and keeping it in sync with provisioning.

If you have an X.509 certificate for a TPM key, the certificate's public key is the key's public half, so the key can be found by matching that public key against the TPM's persistent keys, with no configured handle needed.

Today that takes manual enumeration:

```go
certPub := cert.PublicKey

handles, err := client.Handles(rw, tpm2.HandleTypePersistent)
if err != nil { ... }
for _, h := range handles {
    pub, _, _, err := tpm2.ReadPublic(rw, h)
    if err != nil { ... }
    key, err := pub.Key()
    if err != nil { ... }
    if key.(interface{ Equal(crypto.PublicKey) bool }).Equal(certPub) {
        // found it
    }
}
```

Every consumer that pairs a TPM key with an X.509 certificate (TLS client auth, CSR generation) has to write this same loop.

## Proposal

```go
// FindHandle returns the handle of the persistent key whose public key is pub,
// or ErrNoKeyFound if no persistent key matches.
func FindHandle(rw io.ReadWriter, pub crypto.PublicKey) (tpmutil.Handle, error)
```

This composes with the existing API: the result feeds into `client.LoadCachedKey`.

```go
handle, err := client.FindHandle(rw, cert.PublicKey)
// ...
key, err := client.LoadCachedKey(rw, handle, mySession)
```

I'd scope this to persistent handles only, since the point is finding a key another process provisioned, and transient handles don't cross process boundaries anyway.

Happy to send a PR!

bschaatsbergen:
Closes #901

`client.LoadCachedKey` requires the caller to know a key's handle, but that handle is set at provisioning and can differ between machines, so applications end up carrying it in config and keeping it in sync with provisioning.

If you have an X.509 certificate for a TPM key, the certificate's public key is the key's public half, so the key can be found by matching that public key against the TPM's persistent keys, with no configured handle needed.

This PR adds `FindHandle` to the client package. It returns the handle of the persistent key whose public key matches, or `ErrNoKeyFound` when none does. The result feeds into `LoadCachedKey`.

The search on purpose covers only persistent handles: transient handles never cross a process boundary.

Matching uses `Equal` that all current standard library public key types implement; the one type without it (deprecated [crypto/dsa](https://pkg.go.dev/crypto/dsa@go1.26.5)) gets an error rather than a false not-found. Persistent objects that have no [crypto.PublicKey](https://pkg.go.dev/crypto#PublicKey) form, or that cannot be read, are skipped rather than failing the search.

### DIFF
--- a/client/handles.go
+++ b/client/handles.go
@@ -1,6 +1,8 @@
 package client
 
 import (
+	"crypto"
+	"errors"
 	"fmt"
 	"io"
 	"math"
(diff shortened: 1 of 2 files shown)


bschaatsbergen:
Hi @jkl73, would it be possible to run the tests?

PR #903, opened 2026-07-27
launcher/internal/systemctl: avoid 35s sleep in unit test

brongan:
launcher/internal/systemctl: avoid 35s sleep in unit test via context cancellation

### Summary
Refactor `runSystemdCmd` to accept a caller-supplied `context.Context` instead of creating a hardcoded 30-second timeout context internally. In `systemctl_test.go`, pass a pre-canceled context for the `"timeout"` test case to trigger context cancellation instantly.

### Problem
Previously, `runSystemdCmd` internally created a hardcoded `30 * time.Second` timeout context. In `systemctl_test.go`, the `timeout` test case relied on `time.Sleep(35 * time.Second)` inside a mock unit function to trigger that internal timeout. As a result, running unit tests for `launcher/internal/systemctl` blocked real-time execution for 30+ seconds on every test run.

### Solution
1. **Pass Context to `runSystemdCmd`**: Update `runSystemdCmd` and its callers (`Start`, `Stop`, etc.) to take a `ctx context.Context` parameter.
2. **Pre-cancel Context in Unit Test**: Pass a pre-canceled context for the `"timeout"` test case in `TestRunSystmedCmd`, allowing `runSystemdCmd` to hit `<-ctx.Done()` instantly without blocking or requiring experimental packages.

### Impact / Benchmarks
Running `go test -v ./launcher/internal/systemctl/...`:
- **Before**: `30.02s`
- **After**: `0.015s` (~30s reduction per test execution)

### DIFF
--- a/launcher/internal/systemctl/systemctl.go
+++ b/launcher/internal/systemctl/systemctl.go
@@ -36,12 +36,16 @@ func New() (*Systemctl, error) {
 
 // Start is the equivalent of `systemctl start $unit`.
 func (s *Systemctl) Start(unit string) error {
-	return runSystemdCmd(s.dbus.StartUnitContext, "start", unit)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	return runSystemdCmd(ctx, s.dbus.StartUnitContext, "start", unit)
 }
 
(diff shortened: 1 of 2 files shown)


lindsayzad (approved):
Looks good to me

PR #906, opened 2026-07-28
Launcher: journald JSON logging and add FluentBit

Sibcgh:
Upgrades the Confidential Space launcher to emit structured JSON workload logs to `journald`, allowing Fluent Bit to map `INFO`/`ERROR` severities to Google Cloud Logging while preserving local disk audit trails and preventing container thread stalls.

* **Structured JSON Logging (`JSONSeverityWriter`):** Formats container stdout/stderr as JSON lines (`{"severity":"...", "message":"..."}`) written to `os.Stdout`/`os.Stderr`.
* **Fluent Bit JSON Parser:** Added `parsers-cs.conf` and a parser filter to `fluent-bit-cs.conf` & `bc-fluent-bit-cs.conf` to extract and promote `severity` to Cloud Logging.
* **Boot Service Lifecycle:** Updated `entrypoint.sh`, `vgentrypoint.sh`, and `bcentrypoint.sh` to install `parsers-cs.conf` and restart `fluent-bit.service` before workload launch.
* **Cleanup:** Removed obsolete direct Cloud Logging writers (`SeverityWriter`, `NewInfoWriter`, `NewErrorWriter`) and `workloadLogger` from `container_runner.go`, `logging.go`, `start.go`, and `main.go`.
* **Testing & CI:** Added unit tests (`TestJSONSeverityWriter`) with 50k-line throughput verification, updated Cloud Build integration tests (`read_cloud_logging.sh` & `test_log_redirect.sh`) to verify `INFO` severity.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -50,7 +50,6 @@ type ContainerRunner struct {
 	launchSpec       spec.LaunchSpec
 	attestAgent      agent.AttestationAgent
 	logger           logging.Logger
-	workloadLogger   logging.Logger
 	deviceROTManager *device.ROTManager
 	serialConsole    *os.File
 	powerButton      *powerButtonListener // Populated only for a hardened image
@@ -105,7 +104,6 @@ type RunnerConfig struct {
 	WorkloadService  *workloadservice.Server
 	LaunchSpec       spec.LaunchSpec
 	Logger           logging.Logger
-	WorkloadLogger   logging.Logger
 	SerialConsole    *os.File
 }
 
@@ -114,7 +112,6 @@ func NewRunner(ctx context.Context, cfg *RunnerConfig) (*ContainerRunner, error)
 	cdClient := cfg.ContainerdClient
 	launchSpec := cfg.LaunchSpec
 	logger := cfg.Logger
-	workloadLogger := cfg.WorkloadLogger
 	serialConsole := cfg.SerialConsole
 	image := cfg.Image
 	attestAgent := cfg.AttestAgent
@@ -237,7 +234,6 @@ func NewRunner(ctx context.Context, cfg *RunnerConfig) (*ContainerRunner, error)
 		launchSpec:       launchSpec,
 		attestAgent:      attestAgent,
 		logger:           logger,
-		workloadLogger:   workloadLogger,
 		deviceROTManager: cfg.DeviceROTManager,
 		serialConsole:    serialConsole,
 		powerButton:      powerButton,
@@ -632,14 +628,11 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 		streamOpt = cio.WithStreams(nil, nil, nil)
 		r.logger.Info("Container stdout/stderr will not be redirected.")
 	case spec.Everywhere:
-		w := io.MultiWriter(os.Stdout, r.serialConsole)
-		streamOpt = cio.WithStreams(nil, w, w)
+		streamOpt = cio.WithStreams(nil, io.MultiWriter(logging.NewJSONInfoWriter(), r.serialConsole), io.MultiWriter(logging.NewJSONErrorWriter(), r.serialConsole))
 		r.logger.Info("Container stdout/stderr will be redirected to serial and Cloud Logging. This may result in performance issues due to slow serial console writes.")
 	case spec.CloudLogging:
-		stdoutWriter := logging.NewInfoWriter(r.workloadLogger)
-		stderrWriter := logging.NewErrorWriter(r.workloadLogger)
-		streamOpt = cio.WithStreams(nil, stdoutWriter, stderrWriter)
-		r.logger.Info("Container stdout/stderr will be redirected to Cloud Logging with INFO and ERROR severities respectively.")
+		streamOpt = cio.WithStreams(nil, logging.NewJSONInfoWriter(), logging.NewJSONErrorWriter())
+		r.logger.Info("Container stdout/stderr will be redirected to journald (JSON) for FluentBit scraping.")
 	case spec.Serial:
 		streamOpt = cio.WithStreams(nil, r.serialConsole, r.serialConsole)
 		r.logger.Info("Container stdout/stderr will be redirected to serial logging. This may result in performance issues due to slow serial console writes.")
--- a/launcher/image/entrypoint.sh
+++ b/launcher/image/entrypoint.sh
@@ -5,6 +5,7 @@ main() {
   cp /usr/share/oem/confidential_space/container-runner.service /etc/systemd/system/container-runner.service
   # Override default fluent-bit config.
   cp /usr/share/oem/confidential_space/fluent-bit-cs.conf /etc/fluent-bit/fluent-bit.conf
+  cp /usr/share/oem/confidential_space/parsers-cs.conf /etc/fluent-bit/parsers-cs.conf
 
   # Override default system-stats-monitor.json for node-problem-detector.
   cp /usr/share/oem/confidential_space/system-stats-monitor-cs.json /etc/node_problem_detector/system-stats-monitor.json
@@ -16,8 +17,8 @@ main() {
   cp /usr/share/oem/confidential_space/kernel-monitor-cs.json /etc/node_problem_detector/kernel-monitor.json
   systemctl daemon-reload
   systemctl enable container-runner.service
+  systemctl restart fluent-bit.service
   systemctl start container-runner.service
-  systemctl start fluent-bit.service
 }
 
 main
--- a/launcher/image/fluent-bit-cs.conf
+++ b/launcher/image/fluent-bit-cs.conf
@@ -37,6 +37,7 @@
     #
     # by default 'info' is set, that means it includes 'error' and 'warning'.
     log_level    info
+    parsers_file /etc/fluent-bit/parsers-cs.conf
     # Storage
     # =======
     # Fluent Bit can use memory and filesystem buffering based mechanisms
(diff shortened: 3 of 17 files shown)


launcher/container_runner.go
```diff
@@ -630,14 +630,15 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
...
 	case spec.Everywhere:
-		w := io.MultiWriter(os.Stdout, r.serialConsole)
-		streamOpt = cio.WithStreams(nil, w, w)
-		r.logger.Info("Container stdout/stderr will be redirected to serial and Cloud Logging. This may result in performance issues due to slow serial console writes.")
+		stdoutWriter := &logging.JSONSeverityWriter{Target: os.Stdout, Severity: "INFO"}
+		stderrWriter := &logging.JSONSeverityWriter{Target: os.Stderr, Severity: "ERROR"}
+		streamOpt = cio.WithStreams(nil, io.MultiWriter(stdoutWriter, r.serialConsole), io.MultiWriter(stderrWriter, r.serialConsole))
+		r.logger.Info("Container stdout/stderr will be redirected to serial and journald (JSON) for FluentBit scraping.")
 	case spec.CloudLogging:
-		stdoutWriter := logging.NewInfoWriter(r.workloadLogger)
```
yawangwang, line 639: do we still need `NewInfoWriter` or `NewErrorWriter`?

yawangwang (commented):
Should we add fluent bit configs for CS to route logs to stackdriver based on severity?

launcher/internal/logging/logging.go
```diff
@@ -336,3 +338,46 @@ func (w *SeverityWriter) Write(p []byte) (n int, err error) {
...
+		line = strings.TrimRight(line, "\r")
+		if line == "" {
+			continue
+		}
+		record, marshalErr := json.Marshal(map[string]string{
+			"severity": w.Severity,
+			"message":  line,
+		})
+		if marshalErr != nil {
+			return len(p), marshalErr
```
yawangwang: per standard from https://pkg.go.dev/io#Writer, Write must return n < len(p) when returning a non-nil error.

launcher/internal/logging/logging.go
```diff
@@ -336,3 +338,46 @@ func (w *SeverityWriter) Write(p []byte) (n int, err error) {
...
+		}
+		record = append(record, '\n')
+		if _, err = w.Target.Write(record); err != nil {
+			return len(p), err
```
yawangwang: ditto

launcher/internal/logging/logging.go
yawangwang: Your PR description mentioned these legacy writers are removed but they're still present. Could you keep your changes synced with the description?

launcher/container_runner.go
yawangwang: Do we still need workloadLogger?

launcher/image/fluent-bit-cs.conf
```diff
@@ -58,6 +59,13 @@
     DB /var/log/google-fluentbit/container-runner.log.db
     Read_From_Tail False
 
+[FILTER]
+    Name          parser
+    Match         *
```
yawangwang: shall we use `confidential-space-launcher` rather than `*`?

launcher/image/entrypoint.sh
```diff
@@ -16,8 +17,8 @@ main() {
   cp /usr/share/oem/confidential_space/kernel-monitor-cs.json /etc/node_problem_detector/kernel-monitor.json
   systemctl daemon-reload
   systemctl enable container-runner.service
+  systemctl restart fluent-bit.service
```
yawangwang, line 20: why would we want to change the startup order?

launcher/image/test/scripts/test_log_redirect.sh
yawangwang, line 20: Can we check the severity level ("INFO") specifically in cloud logging tests?

launcher/image/test/util/read_cloud_logging.sh
```diff
@@ -6,5 +6,5 @@
 read_cloud_logging() {
   gcloud logging read "resource.type=\"gce_instance\" jsonPayload._HOSTNAME=\"$1\"
 log_name=\"projects/$PROJECT_ID/logs/confidential-space-launcher\"" \
---format="value(jsonPayload.MESSAGE)" --order asc
+--format="value(jsonPayload.message, jsonPayload.MESSAGE)" --order asc
```
yawangwang: Could you verify if `jsonPayload.MESSAGE` is still needed with the JSON parser in place?

launcher/internal/logging/logging.go
Sibcgh: done!

launcher/internal/logging/logging.go
Sibcgh: done!

launcher/container_runner.go
Sibcgh: `workloadLogger `was only needed when ContainerRunner called the Cloud Logging SDK directly, will remove. Thanks for the catch.

launcher/internal/logging/logging.go
Sibcgh: Done!

launcher/image/fluent-bit-cs.conf
Sibcgh: Agreed, would make the filter more targetted

launcher/image/test/scripts/test_log_redirect.sh
Sibcgh: Updated `read_cloud_logging.sh` to extract the severity field alongside message, and updated `test_log_redirect.sh` to explicitly assert that the received log record has `INFO` severity in Cloud Logging.

launcher/image/test/util/read_cloud_logging.sh
Sibcgh: Yes, `jsonPayload.MESSAGE` is still needed. When testing earlier, I found that:

* **`jsonPayload.MESSAGE` (uppercase):** Launcher operational logs (`logger.Info(...)`) write directly to Cloud Logging via `CloudLogger` (`payloadMessageKey = "MESSAGE"` in `logging.go`).
* **`jsonPayload.message` (lowercase):** Container workload logs are formatted as JSON and unpacked by Fluent Bit into lowercase `message`.
 
Because `read_cloud_logging.sh` is a shared helper, querying both fields in `--format="value(...)"` ensures `test_launcher_workload_cloudlogging.sh` (launcher logs) and `test_log_redirect.sh` (workload logs) both pass without breaking each other.

launcher/image/entrypoint.sh
Sibcgh: In base COS, `fluent-bit.service` is already running on boot before `entrypoint.sh` executes. Without `restart`, Fluent Bit keeps its old config in memory and never loads `parsers-cs.conf` or `fluent-bit-cs.conf`.

I tested removing `restart` earlier today, and presubmit failed on `test_launcher_workload_cloudlogging.sh` because no workload logs were forwarded ([Cloud Build `b5b39c1b`](https://pantheon.corp.google.com/cloud-build/builds;region=us-west1/b5b39c1b-d91e-4a71-a197-3a89e18c96c2;step=21?project=confidential-space-images-dev)).

mayafleischer: Restart makes sense - is there a reason it needs to be before ' systemctl start container-runner.service'?

launcher/container_runner.go
```diff
@@ -649,14 +645,11 @@ func (r *ContainerRunner) Run(ctx context.Context) error {
 		streamOpt = cio.WithStreams(nil, nil, nil)
 		r.logger.Info("Container stdout/stderr will not be redirected.")
 	case spec.Everywhere:
-		w := io.MultiWriter(os.Stdout, r.serialConsole)
-		streamOpt = cio.WithStreams(nil, w, w)
-		r.logger.Info("Container stdout/stderr will be redirected to serial and Cloud Logging. This may result in performance issues due to slow serial console writes.")
+		streamOpt = cio.WithStreams(nil, io.MultiWriter(logging.NewJSONInfoWriter(), r.serialConsole), io.MultiWriter(logging.NewJSONErrorWriter(), r.serialConsole))
+		r.logger.Info("Container stdout/stderr will be redirected to serial and journald (JSON) for FluentBit scraping.")
```
mayafleischer: Should we keep the `"Container stdout/stderr will be redirected to serial and Cloud Logging. This may result in performance issues due to slow serial console writes."` log in this case, too?

launcher/image/entrypoint.sh
Sibcgh: We restart `fluent-bit.service` first so it reloads with our CS config (`parsers-cs.conf` / `fluent-bit-cs.conf`) before `container-runner.service` begins emitting logs, rather than having initial startup logs processed by the default COS Fluent Bit configuration that was already running

launcher/container_runner.go
Sibcgh: good catch, will re-add it back in

launcher/image/entrypoint.sh
yawangwang: Unsure if I followed - if `fluent-bis.service` is already running in base COS, does that mean the previous implementations "systemctl start fluent-bit.service" is actually a no-op and never loads `fluent-bit-cs.conf`?

Not a blocker for your PR but just trying to understand why.

yawangwang (approved):

launcher/image/entrypoint.sh
mayafleischer: My understanding is that for the debug image, "systemctl start fluent-bit.service" is a no-op because fluent-bit.service is started by the default COS boot behaviour before the custom configurations are copied over.
But for the hardened image, the boot target is `cloud-final.service` - which skips a lot of the boot operations (including starting fluent-bit).  So in the hardened case "systemctl start fluent-bit.service" was sufficient for the changes to take effect.

mayafleischer (approved):

PR #913, opened 2026-07-31
Introducing keep-sorted

susienme:
Introducing https://github.com/google/keep-sorted into our CI pipeline to enhance readability. Also, applying it to launch policy, launch spec, and policy constants.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,3 +203,31 @@ jobs:
         run: CGO_CFLAGS=-Werror CC=gcc go build ./...
       - name: Check for CGO Warnings (clang)
         run: CGO_CFLAGS=-Werror CC=clang go build ./...
+
+  keep-sorted:
+    name: Keep Sorted Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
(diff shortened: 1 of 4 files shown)


launcher/spec/launch_policy.go
```diff
@@ -13,14 +13,16 @@ import (
 // LaunchPolicy contains policies on starting the container.
 // The policy comes from the labels of the image.
 type LaunchPolicy struct {
```
yanchengchen7, line 15: I think resorting these unkeyed struct(like LaunchPolicy, and the launchSpec) could break downstreams if any?

.github/workflows/ci.yml
```diff
@@ -203,3 +203,31 @@ jobs:
         run: CGO_CFLAGS=-Werror CC=gcc go build ./...
       - name: Check for CGO Warnings (clang)
         run: CGO_CFLAGS=-Werror CC=clang go build ./...
+
+  keep-sorted:
+    name: Keep Sorted Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
```
yanchengchen7, line 212: Add fetch-depth:2?

susienme: Nice catch. Applied.

launcher/spec/launch_policy.go
susienme: I would agree if the two structs could be used outside launcher. But, those two are only useful inside the launcher. So, I do not expect any downstream breakage.

What do you think?

yanchengchen7: Yes, I don't see any usage outside of the package.

yanchengchen7 (approved):

PR #922, opened 2026-08-05
launcher: compile KeyManager using prebuilt toolchain image

brongan:
Confidential Space presubmit integration tests (`/gcbrun`) currently spend approximately 6 minutes in `KeyManagerCompile` executing runtime HTTP downloads and compiling `bindgen-cli` from crates.io source code on every invocation. This dynamic setup introduces significant serial latency and creates dependency drift risks during presubmit verification.

This change optimizes the build by:
- Adding `launcher/image/keymanager-builder/Dockerfile` to define a toolchain image with pre-installed Clang, CMake 3.28.3, and bindgen-cli 0.70.0.
- Creating `launcher/image/keymanager-builder/cloudbuild.yaml` to ensure the toolchain builder image is compiled 100% reproducibly.
- Updating `KeyManagerCompile` in `launcher/cloudbuild.yaml` to reference the toolchain image by its immutable SHA256 digest, eliminating all runtime toolchain installation commands.

This reduces the `KeyManagerCompile` execution time from ~10m 08s to ~4m 10s, saving approximately 6 minutes on every presubmit run while maintaining full build reproducibility.

### Context & Decisions
- **Deterministic Timestamps:** The toolchain config explicitly enables BuildKit, passes SOURCE_DATE_EPOCH=0, and uses the rewrite-timestamp=true exporter option. This strips all timing non-determinism from layer archives, ensuring that anyone compiling the image from this source gets a byte-for-byte identical digest.
- **No Extra Files Checked In:** The build config is kept minimal and checked in alongside the Dockerfile so that developers and reviewers can rebuild the compiler environment cleanly without ad-hoc scripts.

---

### Reviewer Verification

To verify that the prebuilt toolchain image is fully reproducible and matches the PR code:

1. Submit the build to GCB using your own project:
   ```bash
   gcloud builds submit --config launcher/image/keymanager-builder/cloudbuild.yaml launcher/image/keymanager-builder/ --project=<YOUR_PROJECT_ID>
   ```
2. Verify that the output image digest printed at the end of the build matches the pinned digest value specified in `launcher/cloudbuild.yaml`.

### DIFF
--- a/launcher/cloudbuild.yaml
+++ b/launcher/cloudbuild.yaml
@@ -44,27 +44,14 @@ steps:
     echo "base image:" ${base_image} "project:" ${BASE_IMAGE_PROJECT}
     echo ${base_image} > /workspace/base_image.txt
 
-- name: rust:1.85-bullseye
+- name: 'us-docker.pkg.dev/confidential-space-images-dev/cs-tools/keymanager-builder@sha256:0283448a48bcc515f15d708d5dcf9e368ea2553c8f5a3bb0908019372a47dbfe'
   id: KeyManagerCompile
   entrypoint: /bin/bash
   args:
     - -c
     - |
       set -exuo pipefail
-      # Install build dependencies for Rust
-      apt-get update && apt-get install -y clang build-essential curl tar
-
-      # Install CMake 3.28.3 from pre-compiled binaries for flexibility across environments
-      curl -sSL https://github.com/Kitware/CMake/releases/download/v3.28.3/cmake-3.28.3-linux-x86_64.tar.gz -o cmake.tar.gz
-      tar -zxvf cmake.tar.gz -C /opt/
-      export PATH="/opt/cmake-3.28.3-linux-x86_64/bin:$$PATH"
-      rm cmake.tar.gz
-
-      # Install bindgen-cli (needed by BoringSSL's CMake build)
-      cargo install bindgen-cli
-      export PATH="$$HOME/.cargo/bin:$$PATH"
-
-      # Build the Rust keymanager libraries
+      # Build the Rust keymanager libraries using prebuilt toolchain
       cd keymanager
       cargo build --release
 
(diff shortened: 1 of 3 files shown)


launcher/cloudbuild.yaml
```diff
@@ -41,27 +41,14 @@ steps:
     echo "base image:" ${base_image} "project:" ${BASE_IMAGE_PROJECT}
     echo ${base_image} > /workspace/base_image.txt
 
-- name: rust:1.85-bullseye
+- name: 'us-docker.pkg.dev/confidential-space-images-dev/cs-tools/keymanager-builder@sha256:40f0f7bd481f91ed703fd63d902304d411942afeabc64c680a9d91ce550a2dda'
```
NilanjanDaw: How do we make sure that this is an uncompromised build? and is there an automated way we can trigger rebuilds if the initial build spec changes?

brongan: Updated the PR to use a reproducible build. PR reviewers should be able to reproduce the exact toolchain container image. By matching the produced hash against the configured hash in this PR, you can ensure this is an uncompromised build.

I think triggering rebuilds on spec changes automatically would be in scope for future work.

launcher/image/keymanager-builder/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+FROM rust@sha256:0fdb1727a6c81e0811df9d67bd6defa9a4a07d2a44373ab0a017aedad9fcba3f
+
+ARG SOURCE_DATE_EPOCH=0
+
+ENV CMAKE_VERSION="3.28.3"
+ENV BINDGEN_VERSION="0.70.0"
+
+# Clang is required by bindgen-cli and BoringSSL's C++ FFI build.
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
NilanjanDaw, line 14: Do we need to additionally install `build-essential`? we were originally installing it.

NilanjanDaw (approved):
One minor query otherwise lgtm.

@atulpatildbz this PR is trying to speed up the keymanager compilation using prebuilt packages. Can you review once?

launcher/image/keymanager-builder/Dockerfile
brongan: `/gcbrun` passes without installing `build-essential`. It appears it was superfluous.

PR #924, opened 2026-08-06
launcher: route customizer image builds to test pool

brongan:
In `launcher/image/cloudbuild.yaml`, child image build steps were routed to `cs-image-build-vpc` (which provisions `e2-standard-16` / 16 vCPU worker instances).

Because `cos-customizer` sub-builds only execute lightweight CLI commands that orchestrate GCE customization VMs via API calls and do not perform local compilation, allocating 16-vCPU workers across all 4 concurrent image builds consumed 64 vCPUs of build pool capacity.

This change routes `launcher/image/cloudbuild.yaml` to `cs-image-test-vpc` (which uses `e2-medium` / 1 vCPU workers), freeing 60 vCPUs in the build pool for compilation jobs without affecting image customization speed.

### DIFF
--- a/launcher/image/cloudbuild.yaml
+++ b/launcher/image/cloudbuild.yaml
@@ -45,4 +45,4 @@ timeout: '3000s'
 options:
   dynamic_substitutions: true
   pool:
-    name: 'projects/confidential-space-images-dev/locations/us-west1/workerPools/cs-image-build-vpc'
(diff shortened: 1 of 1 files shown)


mayafleischer (approved):
With this change, is only the bc build process using the -build- pool?

brongan:
> With this change, is only the bc build process using the -build- pool?

After this change, the only uses of `cs-image-build-vpc` are for the parent build and BC image build.

PR #926, opened 2026-08-06
bc: blacklist mlx5_core module

rkolchmeyer:
Also remove konlet mask from kernel command line to avoid going over the command line size limit. Konlet is not installed on these images, so that is unnecessary.


rkolchmeyer:
@jkl73 PTAL

jkl73 (approved):

PR #933, opened 2026-08-12
bc: start fluent-bit explicitly

rkolchmeyer:
The fluent-bit systemd unit depends explicitly on
/etc/cloud-api-domains. This deployment only runs in GDU, so hard code the GDU values in /etc/cloud-api-domains and start fluent-bit explicitly.

### DIFF
--- a/launcher/image/bc-guest/bcentrypoint.sh
+++ b/launcher/image/bc-guest/bcentrypoint.sh
@@ -6,6 +6,13 @@ main() {
     cp /usr/share/oem/ima-policy /sys/kernel/security/ima/policy
   fi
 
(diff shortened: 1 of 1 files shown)


launcher/image/bc-guest/bcentrypoint.sh
```diff
@@ -6,6 +6,14 @@ main() {
     cp /usr/share/oem/ima-policy /sys/kernel/security/ima/policy
   fi
 
+  # Start fluent-bit
```
jkl73: should we do this after line 25 below?

rkolchmeyer: Fixed in new patchset

mayafleischer (approved):

PR #944, opened 2026-08-24
Change GetCertificateChain to GetAKIntermediateCerts in cert.go

ballesterosa:
- Adds partial certificate chain fetching to avoid needing to fetch the root CA in CS client by hard coding the known rootKeyIDs and checking if the issuer of the current cert has that rootKeyID before fetching from the CA.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -344,7 +344,7 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 
 		v.CanonicalEventLog = cosCel.Bytes()
 
-		certChain, err := internal.GetCertificateChain(a.fetchedAK.Cert(), http.DefaultClient)
+		certChain, err := internal.GetAKIntermediateCerts(a.fetchedAK.Cert(), http.DefaultClient)
 		if err != nil {
 			return nil, fmt.Errorf("failed when fetching certificate chain: %w", err)
 		}
--- a/internal/cert.go
+++ b/internal/cert.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"bytes"
 	"crypto/x509"
 	"fmt"
 	"io"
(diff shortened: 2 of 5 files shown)


agent/agent.go
```diff
@@ -344,7 +344,7 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 
 		v.CanonicalEventLog = cosCel.Bytes()
 
-		certChain, err := internal.GetCertificateChain(a.fetchedAK.Cert(), http.DefaultClient)
+		certChain, err := internal.GetCertificateChain(a.fetchedAK.Cert(), http.DefaultClient, 1)
```
jessieqliu: Do we want to rely on the assumption there will only be one intermediate cert? @jkl73 @yawangwang for thoughts.

If we're fine with that, we should define some kind of `numAkIntermediateCerts = 1` const.

internal/cert.go
```diff
@@ -13,11 +13,12 @@ const (
...
 // If an error is encountered in the process, return what has been constructed so far.
-func GetCertificateChain(cert *x509.Certificate, client *http.Client) ([][]byte, error) {
+func GetCertificateChain(cert *x509.Certificate, client *http.Client, maxLength int) ([][]byte, error) {
 	var certs [][]byte
 	currentCert := cert
-	for len(certs) <= maxCertChainLength {
+	for len(certs) <= maxCertChainLength && len(certs) < maxLength {
 		issuingCert, err := fetchIssuingCertificate(client, currentCert)
 		if err != nil {
 			return nil, err
```
jessieqliu, line 51: We can return `certs, err` here - the function is stated to return a partial cert list if an error is encountered but doesn't currently do so

jessieqliu (changes requested):

agent/agent.go
yawangwang: The AK PKI infra is under our control so for now there will be only one intermediate cert but unsure if any future plans to add additional intermediate certs. Adding @jkl73 to confirm.

Ideally, I'd expect a more robust way to not rely on the fixed length check.

jkl73: One option is to compare the Issuer or keyid of the root cert (hardcode the root cert info here), so we know the intermediate cert is signed by the root and we can stop.

we will stop if the singer keyid is a known root cert or the cert is self signed.

ballesterosa: If we use the Issuer.CommonName we can search for the substring "root" in it, if we hardcode the hashes of known root certs we would just have to update them when the certificates change (10 year lifespan?).
In order to figure out if a cert is self signed we would have to fetch it over the network (unless the data for the keyid is cached locally as well).
I think using the keyid known root idea is probably better than the name containing root but either would work.

jkl73: yes, I think adding a known root keyid is better.

We can always fallback to pull the root cert (if in any case, the root cert ever updated in the future), and throw a warning.

jessieqliu: > We can always fallback to pull the root cert

Granted we're talking about a long time period here, but this would result in all old CS images (that check for key ID vN-1) having to adopt the cross-region Root CA dependency whenever the Root CA is updated and we seem to want to avoid that. A more generic way of detecting the Root would avoid that dependency and save us a maintenance step. That being said, returning cert-chain-so-far would mitigate this issue as well so it's not the end of the world.

The general approach SGTM though. If we're hardcoding AK-specific behavior (ie. root keyID) the method name should be more specific as well (IE. `GetAKIntermediateCerts`) since it's currently generically named. I don't think we use the method for anything else so making it more specific should be fine here.

internal/cert.go
ballesterosa: Done.

agent/agent.go
```diff
@@ -344,7 +344,7 @@ func (a *agent) AttestWithClient(ctx context.Context, opts AttestAgentOpts, clie
 
 		v.CanonicalEventLog = cosCel.Bytes()
 
-		certChain, err := internal.GetCertificateChain(a.fetchedAK.Cert(), http.DefaultClient)
+		certChain, err := internal.GetAKIntermidiateCerts(a.fetchedAK.Cert(), http.DefaultClient)
```
jessieqliu: "Intermediate"

internal/cert.go
```diff
@@ -12,23 +13,47 @@ const (
...
+	}
+	for _, rootKeyID := range knownRootKeyIDs {
+		if bytes.Equal(cert.AuthorityKeyId, rootKeyID) {
+			return true
+		}
+	}
+	return false
+}
+
+// GetAKIntermidiateCerts constructs the AK's intermediate certificate chain.
```
jessieqliu: For posterity can you add a comment noting why we stop at the Root CA?

agent/agent.go
ballesterosa: Good catch, fixed.

jessieqliu (approved):

internal/cert.go
```diff
@@ -12,23 +13,47 @@ const (
...
 	var certs [][]byte
 	currentCert := cert
 	for len(certs) <= maxCertChainLength {
+		if isIssuedByKnownRoot(currentCert) {
+			return certs, nil
+		}
 		issuingCert, err := fetchIssuingCertificate(client, currentCert)
 		if err != nil {
-			return nil, err
+			return certs, err
```
yawangwang: if fetching issuer cert failed, then the returned intermediate certs cannot be used for constructing a complete cert chain. So I'd recommend returning nil certs on error.

internal/cert.go
```diff
@@ -12,23 +13,47 @@ const (
...
 		}
 		if issuingCert == nil {
 			return certs, nil
 		}
 		certs = append(certs, issuingCert.Raw)
 		currentCert = issuingCert
 	}
-	return nil, fmt.Errorf("max certificate chain length (%v) exceeded", maxCertChainLength)
+	return certs, fmt.Errorf("max certificate chain length (%v) exceeded", maxCertChainLength)
```
yawangwang: ditto

yawangwang (approved):

internal/cert.go
ballesterosa: Done.

internal/cert.go
ballesterosa: The function comment prior to this change stated that a partial chain should be returned upon fetching error, should I remove that line (line 38 above this function)?

ballesterosa: Actually, since the certs aren't used in testing anywhere if the err is non-nil, i'll revert this and change the comment to reflect the behavior.

internal/cert.go
ballesterosa: Fixed.

PR #947, opened 2026-08-25
add GCENumericInstanceName.

yanchengchen7:


server/instance_info.go
```diff
@@ -17,3 +17,14 @@ func GCEInstanceURL(i *pb.GCEInstanceInfo) string {
 		url.PathEscape(i.GetInstanceName()), // Can use either the name or id here
 	)
 }
+
+// GCENumericInstanceName returns the Compute Engine resource name for the
+// specified instance using its project number and instance id.
+func GCENumericInstanceName(i *pb.GCEInstanceInfo) string {
```
mayafleischer: Could we change this name to `GCEResourceName` or similar?
Having "InstanceName" in the this function when we are using the instance ID is a bit confusing.

server/instance_info.go
```diff
@@ -17,3 +17,14 @@ func GCEInstanceURL(i *pb.GCEInstanceInfo) string {
 		url.PathEscape(i.GetInstanceName()), // Can use either the name or id here
 	)
 }
+
+// GCENumericInstanceName returns the Compute Engine resource name for the
+// specified instance using its project number and instance id.
```
mayafleischer, line 22: Would we like to add to this comment something like "Used for TDX CVM"?   And the same for the function above now that there are essentially two versions.
Alternatively, could we combine the logic into `GCEInstanceURL` - giving the correct URL based on the use case? (Not sure if this is feasible)

mayafleischer (changes requested):

server/instance_info.go
yanchengchen7: Yeah sounds good, updated to GCEInstanceResourceName

server/instance_info.go
yanchengchen7: Seems like over complicating it for a one-liner helper, I think keeping it separate would be more convenient.

mayafleischer (approved):

yawangwang (approved):

PR #949, opened 2026-08-26
Update pinned boringssl commit to match go version 1.26

ballesterosa:
- Updates boringssl version to `440b43a` (current latest main commit) which is on go 1.26.5


atulpatildbz (approved):
Thanks for the PR.
with this change, we'll have some dead code for auto-submodule init in `keymanager/third_party/bssl-sys/build.rs`. created issue for it https://github.com/google/go-tpm-tools/issues/960 to be handled later

also the socket fix must be standardized with for all 5 tests. https://github.com/google/go-tpm-tools/issues/961

Both are lower priority and can be taken in future. Approving this PR

atulpatildbz:
cc: @NilanjanDaw

PR #950, opened 2026-08-26
Upgrade go-tpm dependency to most recent version 0.9.8

ballesterosa:


Sibcgh (approved):
Just waiting for it to pass all presubmits and then we can pull in!

PR #951, opened 2026-08-26
launcher: extract TEE server from runner

brongan:
## Overview

Extracts background server initialization (`TeeServer` and `KeyManager` server) and Unix socket permission validation out of `ContainerRunner.Run` and hoists them directly into `StartLauncher`. 

Additionally:
- Decouples `TeeServer` configuration from `spec.LaunchSpec` in favor of scoped `experiments.Experiments`.
- Simplifies `/v1/keyendorsement` access control by checking `keyClaimsProvider == nil && !experiments.BcMode` rather than relying on redundant boolean flags.
- Prevents spurious `net.ErrClosed` errors during graceful server shutdown.
- Adds comprehensive unit tests for host attestation, shutdown error aggregation, KeyManager integration, and BC-mode key endorsement.

```mermaid
sequenceDiagram
    autonumber
    participant Main as StartLauncher
    participant KM as KeyManager Server
    participant TEE as TeeServer
    participant CR as ContainerRunner

    opt EnableKeyManager is true
        Main->>KM: startKeyManagerServer()
        KM-->>Main: kmServer instance
    end

    Main->>TEE: startTeeServer(experiments, kmServer)
    TEE-->>Main: teeServer instance

    Main->>CR: NewRunner(...)
    Main->>CR: Run(ctx)
    CR-->>Main: workload exits

    Note over Main,KM: Strict LIFO Teardown
    Main->>CR: Close(ctx)
    Main->>TEE: Shutdown(ctx)
    opt EnableKeyManager was true
        Main->>KM: Shutdown(ctx)
    end
```

## Why

1. **Separation of Concerns**: Previously, `ContainerRunner` was burdened with managing daemon HTTP server lifecycles, requiring `AttestClients` and `WorkloadService` to be threaded through `RunnerConfig`. `ContainerRunner` is now solely responsible for container orchestration, execution, and event logging.
2. **Improved Shutdown Handling**: Elevating server lifecycle management to `StartLauncher` enables explicit, clean LIFO resource cleanup via standard `defer` semantics (`r.Close` $\rightarrow$ `teeServer.Shutdown` $\rightarrow$ `kmServer.Shutdown`), ensuring running container processes terminate before their backing Unix sockets are closed.
3. **Decoupled Architecture**: Removing `spec.LaunchSpec` dependencies from `teeserver` simplifies the constructor interface and clarifies dependencies across packages.

### DIFF
--- a/launcher/container_runner.go
+++ b/launcher/container_runner.go
@@ -30,15 +30,13 @@ import (
 	"github.com/google/go-tpm-tools/agent"
 	"github.com/google/go-tpm-tools/agent/device"
 	"github.com/google/go-tpm-tools/cel"
-	workloadservice "github.com/google/go-tpm-tools/keymanager/workload_service"
 	"github.com/google/go-tpm-tools/launcher/internal/healthmonitoring/nodeproblemdetector"
 	"github.com/google/go-tpm-tools/launcher/internal/logging"
 	"github.com/google/go-tpm-tools/launcher/internal/signaturediscovery"
 	"github.com/google/go-tpm-tools/launcher/launcherfile"
 	"github.com/google/go-tpm-tools/launcher/registryauth"
 	"github.com/google/go-tpm-tools/launcher/spec"
-	"github.com/google/go-tpm-tools/launcher/teeserver"
-	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 	"google.golang.org/protobuf/proto"
 
 	attestationpb "github.com/GoogleCloudPlatform/confidential-space/server/proto/gen/attestation"
@@ -53,13 +51,10 @@ type ContainerRunner struct {
 	deviceROTManager *device.ROTManager
 	serialConsole    *os.File
 	powerButton      *powerButtonListener // Populated only for a hardened image
-	attestClients    teeserver.AttestClients
-	workloadService  *workloadservice.Server
 }
 
 const tokenFileTmp = ".token.tmp"
 
-const teeServerSocket = "teeserver.sock"
 const keyManagerGrpcSocket = "kmaserver-grpc.sock"
 
 // Since we only allow one container on a VM, using a deterministic id is probably fine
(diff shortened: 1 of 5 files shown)


launcher/teeserver/tee_server.go
```diff
@@ -79,33 +77,28 @@ type TeeServer struct {
...
-		return nil, fmt.Errorf("failed to chmod unix socket %s: %v", unixSock, err)
-	}
-
-	if launchSpec.Experiments.EnableKeyManager && keyClaimsProvider == nil {
-		return nil, fmt.Errorf("key claims provider cannot be nil when key manager is enabled")
-	}
-
-	kemAttester, err := initKEMAttester(launchSpec.Experiments.BcMode, keyClaimsProvider, a)
+// New creates a TeeServer with the given listener and dependencies.
+func New(
```
susienme, line 81: Even though start.go currently guarantees that `keyClaimsProvider` is not nil when `EnableKeyManager` is set, it would be safer to keep the defensive check here since it is an exported function. Could we restore the previous check `exps.EnableKeyManager && keyClaimsProvider == nil`?

launcher/container_runner.go
```diff
@@ -30,14 +30,12 @@ import (
...
 	"github.com/google/go-tpm-tools/cel"
-	workloadservice "github.com/google/go-tpm-tools/keymanager/workload_service"
 	"github.com/google/go-tpm-tools/launcher/internal/healthmonitoring/nodeproblemdetector"
 	"github.com/google/go-tpm-tools/launcher/internal/logging"
 	"github.com/google/go-tpm-tools/launcher/internal/signaturediscovery"
 	"github.com/google/go-tpm-tools/launcher/launcherfile"
 	"github.com/google/go-tpm-tools/launcher/registryauth"
 	"github.com/google/go-tpm-tools/launcher/spec"
-	"github.com/google/go-tpm-tools/launcher/teeserver"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
```
susienme: Nit: I know this PR doesn't modify this line, but since you're already touching this file, could we remove the redundant `v1` alias here? :)

susienme (approved):

launcher/container_runner.go
brongan: Done!

launcher/teeserver/tee_server.go
brongan: My goal here is to avoid parallel arguements. Passing both `exps` as well as `keyClaimsProvider` is bad because an invalid state is representable. From this function's perspective, it should have `keyClaimsProvider` iff `EnableKeyManager` was set. I'll make this better.

PR #952, opened 2026-08-26
Download Go modules before checking protobuf generation

susienme:
Populate the Go module cache before running `go generate` so `go list` can resolve imported `.proto` paths in `gen_attest.sh`.


brongan (approved):
Thanks for cleaning this up quickly. This looks like the right place to populate the cache.

PR #953, opened 2026-08-27
[GB300] Add experiment and agent shell

mayafleischer:
Add the GB300CCMode experiment and create an initial GB300CCMode attestation agent. 

Most of the logic remains unimplemented, but this sets up the framework for easier follow-ups and enables us to create a test image that is usable. 

Follow up PRs will:
- Add the  GB300CCMode attestation agent implementation
- Add the necessary disk integrity and mount checks to the launcher for GB300CCMode

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -803,7 +803,7 @@ func convertToTPMQuote(v *pb.Attestation) *attestationpb.TpmQuote {
 }
 func (a *gb300ccaAgent) Attest(_ context.Context, _ AttestAgentOpts) ([]byte, error) {
 	a.logger.Info("GB300 CC mode: Skipping Attest (Hardware attestation not implemented)")
-	return []byte("dummy-token"), nil
+	return []byte("eyJhbGciOiJub25lIn0.eyJleHAiOjMyNTAzNjgwMDAwfQ."), nil
 }
 
 func (a *gb300ccaAgent) Refresh(_ context.Context) error {
@@ -818,5 +818,5 @@ func (a *gb300ccaAgent) MeasureEvent(_ gecel.Content) error {
 
 func (a *gb300ccaAgent) AttestWithClient(_ context.Context, _ AttestAgentOpts, _ verifier.Client) ([]byte, error) {
 	a.logger.Info("GB300 CC mode: Skipping AttestWithClient (Hardware attestation not implemented)")
-	return []byte("dummy-token"), nil
+	return []byte("eyJhbGciOiJub25lIn0.eyJleHAiOjMyNTAzNjgwMDAwfQ."), nil
 }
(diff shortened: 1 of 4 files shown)


launcher/start.go
atulpatildbz: i think we'll have to include `launchSpec.Experiments.GB300CCMode` here as well. My understanding is that the GPU drivers are pre-installed for Gb300 as well. @meetrajvala correct me if I'm wrong.

agent/agent.go
```diff
@@ -838,3 +865,22 @@ func convertToTPMQuote(v *pb.Attestation) *attestationpb.TpmQuote {
 		},
 	}
 }
+func (a *gb300ccAgent) Attest(_ context.Context, _ AttestAgentOpts) ([]byte, error) {
+	a.logger.Info("GB300 CC mode: Skipping Attest (Hardware attestation not implemented)")
```
atulpatildbz: iiuc this is for getting OIDC tokens. similar to bcmode we can return error here "attestation token is not supported in BC mode"

agent/agent.go
```diff
@@ -838,3 +865,22 @@ func convertToTPMQuote(v *pb.Attestation) *attestationpb.TpmQuote {
...
+
+func (a *gb300ccAgent) MeasureEvent(_ gecel.Content) error {
+	a.logger.Info("GB300 CC mode: Skipping MeasureEvent")
+	return nil
+}
+
+func (a *gb300ccAgent) AttestWithClient(_ context.Context, _ AttestAgentOpts, _ verifier.Client) ([]byte, error) {
+	a.logger.Info("GB300 CC mode: Skipping AttestWithClient (Hardware attestation not implemented)")
+	return []byte("eyJhbGciOiJub25lIn0.eyJleHAiOjMyNTAzNjgwMDAwfQ."), nil
+}
```
atulpatildbz, line 892: FYI: For `/v1/evidence` on GB300: once `bmsai-attestation-agent` is ready, `GetEvidence` over `bmsai_agent.sock` will return `Gb300Evidence` (`BmsaiEvidence` + `TpmQuote`) rather than `VmAttestation`. We plan to add a `GetGb300Evidence(ctx, challenge) ([]byte, error)` method on `AttestationAgent` (similar to `AttestHost` in `bcAgent`) and route `/v1/evidence` in `teeserver` to it when `GB300CCMode` is enabled.

We'll have to also override `AttestationEvidence` on `gb300ccAgent` here

We can either do it in this PR so calling `/v1/evidence` returns a clear "not yet implemented in GB300 CC mode"

Or we can handle this in a separate PR as well once AA is ready.

atulpatildbz (approved):
Minor comments but we can address them in future PRs as well.

Approving the PR. 

Please add a description too

atulpatildbz:
cc: @vbalain

agent/agent.go
mayafleischer: Thanks - I've added this so it returns a clear error, and can be updated later once we add more of the AA logic.

PR #915, opened 2026-07-31
Fix `zizmor` security findings

susienme:
`zizmor` (go/github-zizmor-help) is a security static analyzer for GitHub Actions workflows. It runs automatically whenever a pull request modifies GitHub Actions workflows. It flagged several unpinned actions and missing permissions in ci.yml (https://github.com/google/go-tpm-tools/actions/runs/30656971512?pr=913).

This change addresses those findings by:
- Adding top-level `permissions: contents: read` to ci.yml and releaser.yaml
- Pinning all GitHub Actions across ci.yml and releaser.yaml to commit SHAs

PR #954, opened 2026-08-27
Increase `golangci-lint` timeout for cold starts

susienme:
PR #915 disabled Go module caching (`cache: false`) to address zizmor security findings. Without the cache, downloading all dependencies over the network and processing them from scratch takes significantly longer. This caused the linter to hit its previous 2-minute deadline and fail with a "Timeout exceeded: try increasing it by passing --timeout option" error.

Fixed by increasing the `--timeout` argument to 5m in `ci.yml` to safely accommodate the initial module downloads.


.github/workflows/ci.yml
```diff
@@ -166,7 +166,7 @@ jobs:
             -E revive
             --max-same-issues=0
             --max-issues-per-linter=0
-            --timeout 2m
+            --timeout 5m
```
brongan, line 169: Does this have an effect on the overall `/gcbrun` time?

brongan (approved):

.github/workflows/ci.yml
susienme: no. this is only for Go linter.

PR #956, opened 2026-08-27
ci: decompose lint matrix and centralize constants

brongan:
## Overview

Decompose the multi-module linter into separate jobs for pure Go code and CGO/Rust dependencies, and centralize language and toolchain versions into workflow constants.

The previous CI configuration executed a 5-way matrix across all repository submodules. Every runner unconditionally installed the Rust toolchain, fetched apt build dependencies, and compiled `bindgen-cli` from source—even when linting standard Go packages with no CGO or Rust code. This added ~8–10 runner-minutes of redundant compilation to every PR.

Isolating KeyManager's CGO and Rust prerequisites to a dedicated job allows pure Go modules to lint cleanly in a single step without external toolchain overhead. Centralizing language and tool versions prevents toolchain drift across jobs and ensures future version bumps are single-line updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,23 +27,27 @@ on:
 permissions:
   contents: read
 
+env:
+  GO_VERSION: '1.26.x'
+  GOLANGCI_LINT_VERSION: 'v2.12.2'
+  PROTOC_VERSION: '3.20.1'
+  PROTOC_GEN_GO_VERSION: 'v1.28.0'
+  KEEP_SORTED_VERSION: 'v0.9.1'
+
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


ballesterosa (approved):
Is it possible to use `dtolnay/install@cargo:bindgen-cli` in CI instead of cargo install?

brongan:
> Is it possible to use `dtolnay/install@cargo:bindgen-cli` in CI instead of cargo install?

This is technically possible. However, we prefer compiling via `cargo install`.
This allows us to
* to build directly against the runner's system Clang/LLVM
* preserve supply chain provenance
* avoid unpinned action / pre-built binary dependencies that conflict with our `zizmor` security policies.

PR #964, opened 2026-09-01
launcher: move BC socket setup to startup

brongan:
## Overview

Moves BC mode KeyManager Unix domain socket permissions adjustment and verification out of the container runner and into host daemon initialization during startup.

## What Changed

* Extracted KeyManager Unix domain socket permission configuration out of container execution.
* Relocated socket permission setup to host server startup so socket permissions are established alongside daemon initialization.

## Why

In BC mode, the KeyManager Unix domain sockets require world-accessible permissions so container workloads can communicate with the host daemon.

Previously, socket permission mutations occurred inside the container runner immediately before starting the container task. Performing host socket manipulations inside the runner tightly coupled container orchestration to host filesystem state and blocked isolated unit testing of the runner.

Configuring all socket permissions during host daemon initialization ensures the host environment is fully established before the runner is invoked, keeping the container runner decoupled from host sockets and ready for isolated unit tests.

### DIFF
--- a/launcher/start.go
+++ b/launcher/start.go
@@ -35,8 +35,9 @@ import (
 )
 
 const (
-	teeServerSocket  = "teeserver.sock"
-	keyManagerSocket = "kmaserver.sock"
+	teeServerSocket      = "teeserver.sock"
+	keyManagerSocket     = "kmaserver.sock"
+	keyManagerGrpcSocket = "kmaserver-grpc.sock"
 )
 
 var expectedTPMDAParams = TPMDAParams{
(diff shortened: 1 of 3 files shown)


launcher/start.go
```diff
@@ -183,6 +184,25 @@ func StartLauncher(ctx context.Context, launchSpec spec.LaunchSpec, logger loggi
 		teeSocket.Close()
 		return fmt.Errorf("failed to create TEE server: %w", err)
 	}
+	if launchSpec.Experiments.BcMode {
```
susienme, line 187: Since this block's errors are only logged (and not returned), consider extracting this socket setup logic into an isolated helper function to improve readability. This function started getting long.

susienme (approved):

launcher/start.go
brongan: Done. Extracted the socket setup logic into setupBCSocketPermissions and added unit test coverage.

PR #969, opened 2026-09-02
test/debug: make step dependencies explicit

brongan:
test/debug: make step dependencies explicit

In `launcher/image/test/test_debug_cloudbuild.yaml`, steps previously omitted the `waitFor` field. When `waitFor` is omitted, Cloud Build defaults to implicit serial execution, causing each step to wait for all preceding steps in the file rather than declaring its actual requirements.

This change adds explicit `waitFor` dependencies to each step in the workflow:
- `CreateVM` sets `waitFor: ['-']` to begin immediately at build start.
- `BasicWorkloadTest` and `BasicWorkloadTestCloudLogging` explicitly wait for `CreateVM`.
- `ChangeMDSVariables` explicitly waits for `BasicWorkloadTest`, and `ChangeMDSVariablesTest` waits for `ChangeMDSVariables`.
- `CheckDebugVMAliveAfterLauncherExits` explicitly waits for `BasicWorkloadTest`.
- `CleanUp` explicitly waits for all test tracks (`BasicWorkloadTestCloudLogging`, `ChangeMDSVariablesTest`, and `CheckDebugVMAliveAfterLauncherExits`) to ensure the instance is preserved until all tests finish.
- `CheckFailure` explicitly waits for `CleanUp` before verifying build results.

Declaring these dependencies directly removes reliance on Cloud Build's implicit serial execution fallback and ensures step prerequisites and teardown barriers are explicitly defined.

#### Before (Implicit Serial Execution)
```mermaid
graph TD
  S0[CreateVM] --> S1[BasicWorkloadTest]
  S1 --> S2[BasicWorkloadTestCloudLogging]
  S2 --> S3[ChangeMDSVariables]
  S3 --> S4[ChangeMDSVariablesTest]
  S4 --> S5[CheckDebugVMAliveAfterLauncherExits]
  S5 --> S6[CleanUp]
  S6 --> S7[CheckFailure]
```

#### After (Explicit DAG)
```mermaid
graph TD
  Start((Build Start)) --> VM[CreateVM]
  VM --> BW[BasicWorkloadTest]
  VM --> CL[BasicWorkloadTestCloudLogging]

  BW --> CM[ChangeMDSVariables] --> CMT[ChangeMDSVariablesTest]
  BW --> CD[CheckDebugVMAliveAfterLauncherExits]

  CL --> Clean[CleanUp]
  CMT --> Clean
  CD --> Clean

  Clean --> Barrier[CheckFailure]
```

### DIFF
--- a/launcher/image/test/test_debug_cloudbuild.yaml
+++ b/launcher/image/test/test_debug_cloudbuild.yaml
@@ -9,6 +9,7 @@ substitutions:
 steps:
 - name: 'gcr.io/cloud-builders/gcloud'
   id: CreateVM
+  waitFor: ['-']
   entrypoint: 'bash'
   env:
   - 'BUILD_ID=$BUILD_ID'
@@ -21,16 +22,19 @@ steps:
         ]
 - name: 'gcr.io/cloud-builders/gcloud'
   id: BasicWorkloadTest
+  waitFor: ['CreateVM']
   entrypoint: 'bash'
   args: ['scripts/test_launcher_workload.sh', '${_VM_NAME_PREFIX}-${BUILD_ID}', '${_ZONE}']
 - name: 'gcr.io/cloud-builders/gcloud'
   id: BasicWorkloadTestCloudLogging
+  waitFor: ['CreateVM']
   entrypoint: 'bash'
   env:
   - 'PROJECT_ID=$PROJECT_ID'
   args: ['scripts/test_launcher_workload_cloudlogging.sh', '${_VM_NAME_PREFIX}-${BUILD_ID}']
 - name: 'gcr.io/cloud-builders/gcloud'
   id: ChangeMDSVariables
+  waitFor: ['BasicWorkloadTest']
   entrypoint: 'bash'
   args: ['util/change_metadata_vars.sh',
           '-n', '${_VM_NAME_PREFIX}-${BUILD_ID}',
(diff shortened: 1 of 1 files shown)


launcher/image/test/test_debug_cloudbuild.yaml
atulpatildbz, line 60: FYI. even though the intent has been to wait for 150 s here, in the current sequencial flow we're waiting for much longer. 

with your changes `CheckDebugVMAliveAfterLauncherExits` would start immediately after workload finishes. shutdown timer in `exit_script.sh` is 2 minutes. which means the GCE status must transition out of RUNNING in ~30s.

This only matters if there's a regression where debug image accidentally executes shutdown script.
 
Check if that's enough or do we need to increase this sleep time.

atulpatildbz, line 68: this step, as well as BasicWorkloadTestCloudLogging and ChangeMDSVariablesTest write to same status.txt file.
it was fine before since the steps were sequential. but now do you think we should create step-specific files `/workspace/status_${STEP_NAME}.txt`?

atulpatildbz (commented):
Didn't realize i was accidentally added as a reviewer. still adding some minor comments. Feel cree to close them

launcher/image/test/test_debug_cloudbuild.yaml
brongan: Great catch!

I think we should make the functionality match the intent. I think that `/gcbrun` on this PR should be sufficient evidence that this is fine. If this flakes later, we can both bump the number and the comment explaining why.

brongan: I do not think that we need step specific files. Checking multiple result files makes the join barrier much more complex. If we have issues determining which step failed, then this could be a potential improvement.

yawangwang (approved):

PR #975, opened 2026-09-03
launcher: add container runner unit tests

brongan:
## Overview

Adds mock task primitives and hermetic unit test coverage for the container runner's execution and cleanup lifecycle.

## What Changed

- Implements runtime doubles to simulate task execution, exit statuses, stream waits, and signals without live daemons.
- Tests workload execution lifecycle, including normal exit success and workload exit error mapping.
- Tests runtime failure scenarios during task creation, startup failure, and exit status streaming.
- Tests container log stream redirection across all supported destinations and console handles.
- Tests error propagation for attestation event logging and token fetch failures.
- Verifies teardown sequencing between container stream flushes and task deletion.
- Inlines dependency construction directly in each test without shared mutable fixtures.

## Why

The container runner coordinates workload execution, log redirection, and shutdown handling. Previously, verifying this orchestration required spinning up complete Confidential Space VM images in GCE integration tests. Decoupling the runner from live container runtime daemons and physical hardware allows validating these lifecycle and failure paths deterministically in fast unit tests.

### DIFF
--- a/launcher/container_runner_test.go
+++ b/launcher/container_runner_test.go
@@ -15,12 +15,14 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"testing"
 	"time"
 
 	clogging "cloud.google.com/go/logging"
 	"github.com/cenkalti/backoff/v4"
 	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/defaults"
 	"github.com/containerd/containerd/namespaces"
@@ -54,6 +56,8 @@ type fakeAttestationAgent struct {
 	sigsCache        []string
 	sigsFetcherFunc  func(context.Context) []string
 
+	closeCalled bool
+
 	// attMu sits on top of attempts field and protects attempts.
 	attMu    sync.Mutex
 	attempts int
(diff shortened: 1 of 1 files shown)


launcher/container_runner_test.go
```diff
@@ -875,16 +883,86 @@ func (c *fakeContainer) ID() string {
 	return "tee-container"
 }
 
-func (c *fakeContainer) Delete(context.Context, ...containerd.DeleteOpts) error {
+func (c *fakeContainer) Delete(_ context.Context, _ ...containerd.DeleteOpts) error {
```
susienme: I think you can omit both `_` as well like the previous line.

launcher/container_runner_test.go
```diff
@@ -850,21 +855,24 @@ func TestMeasureCELEvents(t *testing.T) {
 	}
 }
 
-// This ensures fakeContainer implements containerd.Container interface.
-var _ containerd.Container = &fakeContainer{}
-
-// This ensures fakeImage implements containerd.Image interface.
-var _ containerd.Image = &fakeImage{}
```
susienme: Even for test, I would keep this compile-time check. Consider keeping these.

launcher/container_runner_test.go
```diff
@@ -1153,3 +1237,414 @@ func TestNewRunner(t *testing.T) {
...
+		container:   cont,
+		attestAgent: &fakeAttestationAgent{},
+		launchSpec: spec.LaunchSpec{
+			DisableGcaRefresh: true,
+			LogRedirect:       spec.Nowhere,
+		},
+		logger: logging.SimpleLogger(),
+	}
+
+	if err := runner.Run(context.Background()); err != nil {
```
susienme: t.Context() instead of context.Background()

PR #977, opened 2026-09-04
launcher: eliminate experiment retry test delay

brongan:
Testing experiment retrieval failure previously took over thirteen seconds due to exponential backoff delays across multiple retry attempts. This wall-clock wait added unnecessary latency to local test runs and CI pipelines while only verifying retry counts.

Running the retry test in a synctest bubble advances synthetic time instantaneously when goroutines block. This preserves the real exponential backoff schedule and retry assertions while executing the test in under a millisecond.

### DIFF
--- a/launcher/spec/launch_spec_test.go
+++ b/launcher/spec/launch_spec_test.go
@@ -5,6 +5,7 @@ import (
 	"path"
 	"regexp"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


yawangwang (approved):
Looks like synctest is a new package introduced in Go v1.25 specifically designed for mocking time. Before this package, we implemented a custom time-mocking solution in G3 for our CC unit tests. Could you help replace our legacy implementation with this new native package?

brongan:
> Looks like synctest is a new package introduced in Go v1.25 specifically designed for mocking time. Before this package, we implemented a custom time-mocking solution in G3 for our CC unit tests. Could you help replace our legacy implementation with this new native package?

Great catch. Our custom time-mocking pattern in G3 has been ripe for modernization, and `synctest` is the exact right native fix for it. Definitely worth filing a ticket for the team backlog—this PR should serve as a solid reference implementation for whoever picks it up.

PR #982, opened 2026-09-09
launcher: add support for Nvidia RTX PRO 6000

meetrajvala:
Support NVIDIA RTX PRO 6000 GPU in Confidential Space:
- Enable driver installation flow for RTX PRO 6000
  * Update cos-tools to v0.0.0-20260828164205-7c1f394ced14.
  * Add ConfidentialG4Support experiment flag in launcher.
  * Pass PCI devices directory to GetGPUTypeInfo and validate supported GPUs (H100, B200, RTX PRO 6000).
- Update attestation collection utility for RTX PRO 6000
  * Support SPT attestation report collection for single RTX PRO 6000 GPU.
  * Add unit tests in attestation_test.go.
- Add image integration tests
  * Add CreateSEVCVMWithRTXPro6000, RTXPro6000WorkloadTest, and RTXPro6000VmCleanUp to test_gpu_driver_installation_cloudbuild.yaml.


launcher/internal/gpu/driverinstaller.go
```diff
@@ -57,16 +57,19 @@ func (di *DriverInstaller) InstallGPUDrivers(ctx context.Context) error {
...
 	switch gpuType {
-	case deviceinfo.B200:
-	case deviceinfo.H100:
-	default:
-		return fmt.Errorf("unsupported GPU type %s for Confidential Computing (only supported for H100[a3-highgpu-1g] and B200[a4-highgpu-8g])", gpuType.String())
+	case deviceinfo.NO_GPU:
+		return fmt.Errorf("failed to get the GPU type info: no GPU detected")
+	case deviceinfo.Others:
+		return fmt.Errorf("failed to get the GPU type info: unrecognized GPU")
+	}
```
yawangwang: Duplicate logics with `CheckGPUSupport`

launcher/internal/gpu/attestation.go
yawangwang, line 197: Do we want to add RTX pro 6000 support here?

launcher/internal/experiments/experiments.go
```diff
@@ -17,6 +17,7 @@ type Experiments struct {
 	EnableAttestationEvidence    bool
 	EnableB200DriverInstallation bool
 	EnableH100DriverInstallation bool
+	ConfidentialG4Support        bool
```
yawangwang, line 20: Do we plan to create server-side experiment?

launcher/spec/launch_spec.go
```diff
@@ -377,7 +377,7 @@ func GetLaunchSpec(ctx context.Context, logger logging.Logger, client *metadata.
 		return LaunchSpec{}, fmt.Errorf("failed to validate mounts: %v", errors.Join(errs...))
 	}
 
-	if !spec.Experiments.EnableB200DriverInstallation && !spec.Experiments.EnableH100DriverInstallation && spec.InstallGpuDriver {
+	if !spec.Experiments.EnableB200DriverInstallation && !spec.Experiments.EnableH100DriverInstallation && !spec.Experiments.ConfidentialG4Support && spec.InstallGpuDriver {
```
atulpatildbz, line 380: can we align with similar naming convention? EnableRTXPro6000DriverInstallation?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -299,3 +302,19 @@ func NvidiaSmiOutputFunc(args ...string) NvidiaSmiCmdOutput {
 		return exec.Command(cmdPath, args...).Output()
 	}
 }
+
+// CheckGPUSupport checks if the detected GPU hardware is supported for Confidential Computing.
+func CheckGPUSupport(gpuType deviceinfo.GPUType) error {
```
atulpatildbz: can this be made private?

launcher/go.mod
```diff
@@ -114,3 +115,5 @@ replace (
 	github.com/google/go-tpm-tools/keymanager v0.4.4 => ../keymanager
 	github.com/google/go-tpm-tools/verifier v0.4.4 => ../verifier
 )
+
+replace github.com/opencontainers/runtime-spec => github.com/opencontainers/runtime-spec v1.2.0
```
atulpatildbz, line 122: can you add a comment on why this replace is required?

launcher/spec/launch_spec.go
atulpatildbz, line 380: something to think about separately: we're adding flags per platform, but how different is the driver installation flow between each platform? i believe they're the same, right?

launcher/internal/gpu/driverinstaller.go
```diff
@@ -299,3 +302,19 @@ func NvidiaSmiOutputFunc(args ...string) NvidiaSmiCmdOutput {
...
+	switch gpuType {
+	case deviceinfo.B200:
+	case deviceinfo.H100:
+	case deviceinfo.RTX_PRO_6000:
+	case deviceinfo.NO_GPU:
+		return fmt.Errorf("failed to get the GPU type info: no GPU detected")
+	case deviceinfo.Others:
+		return fmt.Errorf("failed to get the GPU type info: unrecognized GPU")
+	default:
+		return fmt.Errorf("unsupported GPU type %s for Confidential Computing (only supported for H100[a3-highgpu-1g], B200[a4-highgpu-8g], and RTX PRO 6000[g4-standard])", gpuType.String())
```
atulpatildbz: g4-standard → g4-standard-48

launcher/image/test/test_gpu_driver_installation_cloudbuild.yaml
```diff
@@ -65,6 +65,22 @@ steps:
...
+  id: CreateSEVCVMWithRTXPro6000
+  waitFor: ['-']
+  entrypoint: 'bash'
+  env:
+  - 'BUILD_ID=$BUILD_ID'
+  args: ['create_vm.sh','-i', '${_IMAGE_NAME}',
+          '-p', '${_IMAGE_PROJECT}',
+          '-m', 'tee-image-reference=${_WORKLOAD_IMAGE},tee-container-log-redirect=true,tee-install-gpu-driver=true',
+          '-n', '${_VM_NAME_PREFIX}-${BUILD_ID}-rtxpro6000',
+          '-z', '${_ZONE}',
```
atulpatildbz, line 78: this defaults to us-central1-f  correct?
no capacity concern for G4 in it i hope?

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ permissions:
   contents: read
 
 env:
-  GO_VERSION: '1.26.x'
+  GO_VERSION: '1.26.6'
```
atulpatildbz, line 31: IIUC you changed this because cos-tools requires 1.26.6

but 1.26.x already pulls the latest version of go right?

launcher/internal/gpu/driverinstaller.go
meetrajvala: Done. Removed the duplicate switch check in `InstallGPUDrivers()` since `checkGPUSupport()` already handles both NO_GPU and Others.

launcher/internal/gpu/attestation.go
meetrajvala: RTX PRO 6000 is based on the Blackwell architecture, so it is already covered by the existing case `BLACKWELL` check in convertGPUArchToPB.

launcher/internal/gpu/driverinstaller.go
meetrajvala: Done. Renamed CheckGPUSupport to checkGPUSupport.

launcher/internal/gpu/driverinstaller.go
meetrajvala: Done. Updated the supported machine type string in `checkGPUSupport` and its unit test to g4-standard-48.

launcher/spec/launch_spec.go
meetrajvala: This experiment name (ConfidentialG4Support) is read from the experiment binary and matches the name defined in our previously submitted client-side experiment change, so we need to keep this field name to match the key

meetrajvala: Yes, the underlying driver installation flow via cos-gpu-installer is identical across supported GPU families. Having per-platform experiment flags allows us to gate and roll out new GPU family support independently.

launcher/go.mod
meetrajvala: Done. Added a comment explaining that upgrading cos-tools pulls in containerd/v2 v2.3.2 which requires runtime-spec v1.3.0, whereas containerd v1.7.23 (used by launcher) fails to compile with runtime-spec v1.3.0 due to LinuxPids.Limit changing from int64 to *int64.

launcher/image/test/test_gpu_driver_installation_cloudbuild.yaml
meetrajvala: Yes, it defaults to `us-central1-f` and there is enough preemptible quota.

.github/workflows/ci.yml
meetrajvala: By default, actions/setup-go resolves '1.26.x' against the GitHub runner's locally cached Go versions first (unless check-latest: true is passed). The runner image currently has Go 1.26.4 pre-installed, which satisfies '1.26.x' but fails the build since cos-tools requires go >= 1.26.6. Specifying '1.26.6' forces actions/setup-go to fetch the required version.

launcher/spec/launch_spec.go
yawangwang: Actually a new design to remove client-side experiment from image side is under review. The rational of removing the client-side experiment is largely because we want to rely on features managed per-image release. @jkl73 do you think we may want to start the new process from RTX pro 6000?

jkl73: Yes, I think we could, we can ignore the client-side experiment.

Though looks like this flag technically is not blocking anything: I can create a G4 machine with EnableH100DriverInstallation (instead of the g4 flag), and I'm still able to use g4

launcher/internal/experiments/experiments.go
meetrajvala: No, we will only have client side experiment. For server side, we will verify GPU attestation only if request has GPU evidence.

launcher/spec/launch_spec.go
meetrajvala: Removed the client side experiment check.

go.work
```diff
@@ -1,4 +1,4 @@
-go 1.26
+go 1.26.6
```
atulpatildbz, line 1: can you update the README.md? it still says 1.20

launcher/image/test/scripts/gpu/test_gpu_nogpu.sh
```diff
@@ -7,7 +7,7 @@ source util/read_serial.sh
 SERIAL_OUTPUT=$(read_serial $1 $2) 
 print_serial=false
 
-if echo $SERIAL_OUTPUT | grep -q 'failed to get the GPU type info'
+if echo $SERIAL_OUTPUT | grep -q -E 'failed to get the GPU type info|no GPU detected'
```
atulpatildbz: nit: the NO_GPU case returns `fmt.Errorf("failed to get the GPU type info: no GPU detected")`

the original grep already matches it. the |no GPU detected is redundant here.

launcher/internal/gpu/config.go
atulpatildbz, line 13: should InstallerImageRef and InstallerDigest be updated too? this image seems to be built a long time ago. does it have the cos-tools that you need during `/cos-gpu-installer install`?

atulpatildbz (approved):
PR description still mentions `ConfidentialG4Support` flag, please remove it.

Approved. but do check if InstallerImageRef and InstallerDigest need to be updated as well.

launcher/internal/experiments/experiments.go
meetrajvala: We have added new server side and we are removing the client side experiment.

go.work
meetrajvala: Done. Updated README.md to specify Go 1.26.6 or newer.

launcher/image/test/scripts/gpu/test_gpu_nogpu.sh
meetrajvala: Updated to `grep -q 'failed to get the GPU type info: no GPU detected'` to remove the redundant alternation while explicitly matching the NO_GPU case (no GPU detected), since `failed to get the GPU type info` is also a prefix for unrecognized GPU error.

launcher/internal/gpu/config.go
meetrajvala: Ideally, we should derive the cos-gpu-installer reference from the base COS image rather than hardcoding it here. I am planning to address that in a separate follow-up PR.

yawangwang (approved):

launcher/internal/gpu/attestation.go
```diff
@@ -172,12 +172,12 @@ func (a *NvidiaAttester) collectAttestationEvidence(handler gpu.NvmlHandler, non
...
 func determineAttestationType(gpuInfos []*attestationpb.GpuInfo) attestationType {
-	gpuType, _ := getGpuTypeInfo()
-	if gpuType != deviceinfo.H100 && gpuType != deviceinfo.B200 {
+	gpuType, _ := getGpuTypeInfo(PciDevicesDir)
+	if gpuType != deviceinfo.H100 && gpuType != deviceinfo.B200 && gpuType != deviceinfo.RTX_PRO_6000 {
 		return UNSUPPORTED
 	}
-	// H100 can only support single GPU attestation in CS at the moment.
-	if gpuType == deviceinfo.H100 && len(gpuInfos) != 1 {
+	// H100 and RTX PRO 6000 can only support single GPU attestation in CS at the moment.
```
jkl73: should probably also make sure len(gpuInfos) > 0

PR #984, opened 2026-09-11
internal: fix body leak in cert fetcher and bounds issues in PCR helpers

ballesterosa:


internal/pcrs.go
```diff
@@ -123,6 +116,9 @@ func encodePCRSelection(sel tpm2.PCRSelection) []byte {
 	// Encode pcrSelect bitmask
 	pcrBits := make([]byte, 3)
 	for _, pcr := range sel.PCRs {
+		if pcr < 0 || pcr >= 24 {
```
jessieqliu, line 120: Can you add a comment noting where the 24 threshold comes from?

jessieqliu (approved):

internal/pcrs.go
ballesterosa: Done.

PR #986, opened 2026-09-11
Update bcagent and add host RoT

mayafleischer:


agent/agent.go
```diff
@@ -676,6 +682,64 @@ func (t *tdxAttestRoot) ComputeNonce(challenge []byte, extraData []byte) []byte
...
+	return nil, fmt.Errorf("hostServiceRoT does not support direct attestation")
+}
+func (h *hostServiceRoT) ComputeNonce(_, _ []byte) []byte { return nil }
+func (h *hostServiceRoT) Close() error {
+	if h.conn == nil {
+		return nil
+	}
+	return h.conn.Close()
+}
+
```
jkl73: format looks a bit off

agent/agent.go
```diff
@@ -676,6 +682,64 @@ func (t *tdxAttestRoot) ComputeNonce(challenge []byte, extraData []byte) []byte
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	resp, err := h.client.RecordWorkloadEvent(ctx, &hostservicepb.RecordWorkloadEventRequest{
+		CosEvent: tlvBytes,
+	})
+	if err != nil {
+		return fmt.Errorf("hostservice RecordWorkloadEvent failed: %w", err)
+	}
+	if !resp.GetSuccess() {
```
jkl73: looks like this is not needed, but it's fine to keep it now, can remove later if we want to update the proto

jkl73 (approved):

agent/agent.go
mayafleischer: I removed the success check from here, and am going to update the proto now.  I think we will lose track/it will be harder later.

PR #990, opened 2026-09-16
keymanager: fix potential index out of bounds and freed data leakage

ballesterosa:


keymanager/km_common/src/crypto/secret_box.rs
```diff
@@ -10,8 +10,19 @@ pub struct SecretBox(Box<[u8]>);
...
+        if data.capacity() > data.len() {
+            let mut boxed = vec![0u8; data.len()].into_boxed_slice();
+            boxed.copy_from_slice(&data);
+            let cap = data.capacity();
+            data.resize(cap, 0);
+            data.zeroize();
+            Self(boxed)
+        } else {
+            Self(data.into_boxed_slice())
+        }
```
atulpatildbz: we can avoid the if else by copying to a new buffer box, zeroize the data and return the new buffer

```
    let boxed: Box<[u8]> = data.as_slice().into();
    data.zeroize();
    Self(boxed)
```

the secret exists in `[0..len]` only. so zeroizeing just that should be fine
but if you still want to zerioize the `len..cap` part, you could call `data.resize(data.capacity(), 0);` before zerioizeing

atulpatildbz (approved):
Minor comment.

keymanager/km_common/src/crypto/secret_box.rs
ballesterosa: Done.

PR #991, opened 2026-09-16
launcher: fix potential runtime panics, and other bugs

ballesterosa:
- add nil/length checks to avoid runtime panics
- only log envvar name
- add enabled flag to 1 with spec.All


launcher/teeserver/key_attestation_test.go
```diff
@@ -493,3 +493,43 @@ func TestRemoteBindingKeyAttester_GetKeyEndorsement(t *testing.T) {
 		})
 	}
 }
+
+func TestGetKeyEndorsementNilKeyHandle(t *testing.T) {
+	ctx := context.Background()
```
susienme: Delete this line, and use `t.Context()` inside `t.Run()` below.

launcher/spec/launch_policy_test.go
```diff
@@ -947,3 +948,25 @@ func TestGetMonitoringPolicyErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestVerifyEnvVarErrorDoesNotLeakValue(t *testing.T) {
```
susienme: I don't think this test is necessary.

launcher/spec/launch_policy.go
```diff
@@ -265,7 +265,7 @@ func verifyMonitoringConfig(policy MonitoringType, spec MonitoringType) error {
 func (p LaunchPolicy) Verify(ls LaunchSpec) error {
 	for _, e := range ls.Envs {
 		if !contains(p.AllowedEnvOverride, e.Name) {
-			return fmt.Errorf("env var %s is not allowed to be overridden on this image; allowed envs to be overridden: %v", e, p.AllowedEnvOverride)
+			return fmt.Errorf("env var %s is not allowed to be overridden on this image; allowed envs to be overridden: %v", e.Name, p.AllowedEnvOverride)
```
susienme, line 268: This will fix this occurrence, but we have track down all occurrences and keep doing it in the future. Instead, how about implementing the `fmt. Stringer()` interface?
e.g.,

func (e EnvVar) String() string {
  return e.Name
}

Then, we don't need to worry about this bug in the future unless someone explicitly uses `%#v` or prints `e.Value`.

launcher/teeserver/key_attestation_test.go
```diff
@@ -493,3 +493,43 @@ func TestRemoteBindingKeyAttester_GetKeyEndorsement(t *testing.T) {
...
+		}
+	})
+
+	t.Run("bcBindingKeyAttester", func(t *testing.T) {
+		attester := &bcBindingKeyAttester{}
+		_, err := attester.GetKeyEndorsement(ctx, req, agent.AttestAgentOpts{})
+		if err == nil {
+			t.Fatal("expected error for nil KeyHandle")
+		}
+	})
```
susienme: 1. Please use go/go-style/decisions#table-driven-tests.
2. Prefer `t.Error()` go/go-style/decisions#keep-going
3. Got before want, go/go-style/decisions#got-before-want. I think in this case,  "GetKeyEndorsement(nil KeyHandle) = nil error, want error" or similar.

ballesterosa: Done.

launcher/spec/launch_policy.go
ballesterosa: Implemented to protect against future use, and kept the change for .Name in the test for clarity.

launcher/spec/launch_policy_test.go
ballesterosa: Removed.

launcher/teeserver/key_attestation_test.go
ballesterosa: Done.

susienme (approved):

PR #993, opened 2026-09-16
agent: fix potential race condition

ballesterosa:
Fixes `tpmAttestRoot.Extend` calling `tpm2.PCRExtend` on `a.tpm` without holding `a.tpmMu`, causing concurrent command corruption when racing with attestation or quote operations.


agent/agent_test.go
```diff
@@ -1176,3 +1176,39 @@ func TestHostAttestation_NotBcMode(t *testing.T) {
 		t.Error("expected error when BcMode is disabled, got nil")
 	}
 }
+
+func TestTPMAttestRoot_ExtendLocksMutex(t *testing.T) {
+	tpm := test.GetTPM(t)
+	defer client.CheckedClose(t, tpm)
```
jessieqliu: Prefer `t.Cleanup` in tests.

jessieqliu (approved):
LGTM - can you add some info in the PR description about the race condition being fixed for posterity?

agent/agent_test.go
ballesterosa: Done.

